PR #11, opened 2015-07-02
updated script name and link

adrelanos:
The other links on https://github.com/QubesOS/qubes-doc/blob/master/Dom0SecureUpdates.md are still broken. I haven't figured out their new location.

Nonetheless, this pull request should be mergeable as is.


fowlslegs:
You're pulling links from a repo that hasn't been updated in 3 years (abeluck's). Assuming we want the release 2 version (I don't think it would make sense to link the HEAD), I've made my own pull request fixing this (see #12).

These links should be pretty stable and easy to update come R3.

adrelanos:
Thanks.

PR #13, opened 2015-07-03
Removed typesetting artifacts

fowlslegs:
Probably from the old wiki, these shell command typesetting statements are no longer needed (as they have been replaced by simple indentation) and are registered as plaintext by markdown.

### DIFF
--- a/SoftwareUpdateDom0.md
+++ b/SoftwareUpdateDom0.md
@@ -35,15 +35,11 @@ Of course, command line tools are still available for accomplishing various upda
 
 1.  To check and install updates for dom0 software:
 
-    {% highlight trac-wiki %}
     $ sudo qubes-dom0-update
-    {% endhighlight %}
 
 1.  To install additional packages in dom0 (usually not recommended):
 
-    {% highlight trac-wiki %}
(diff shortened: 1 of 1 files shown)


fowlslegs:
So I'm not actually 100% the trac typesetting macros here are not needed/ used anymore. If you go to Github's markdown rendering, they're obviously not necessary, but it looks like on the qubes-os.org site they are still in use. Idk if we're in the middle of transitioning or something. I can make a little script (or probably there's a coreutil would work) and get rid of the rest of them, or otherwise just (obviously) reject this pull request if we shouldn't get rid of them.

Zrubi:
We should ask @Nukama about this.

Nukama:
These aren't trac typesettings artifacts, they are used for jekyll kramdown highlighting.
Indentation should not be the primary way for typesetting (IMHO).

    markdown indentation

```
fenced code
```

``` Fenced code ``` is a easier/efficient way of highlighting (easier than {% highlight %} or indentation). And I may transition them to this redcarpet supported format with a script, if there is no objection against the redcarpet markdown processor.

PR #47, opened 2015-10-16
Windows and Fedora template customizations

ptitdoc:


adrelanos:
I suggest to make `I recommend` less personal and passive `it is recommended`. Not super important, could be done with a pull request on top of this one.

axon-qubes:
Thank you, @ptitdoc!

PR #57, opened 2015-11-20
resize2fs not required for qvm-grow-private

adrelanos:
`sudo resize2fs /dev/xvdb` doesn't work. Please try.

Also it's not required. So let's remove it.

If you can confirm the same behavior, please merge.

### DIFF
--- a/configuration/resize-disk-image.md
+++ b/configuration/resize-disk-image.md
@@ -26,12 +26,6 @@ qvm-grow-private <vm-name> <size>
 
 Note: Size is the target size (i.e. 4096MB or 16GB, ...), not the size to add to the existing disk.
 
-Note2: If once the VM is started, the disk is has not been increased, you can issue in the VM's terminal:
-
-~~~
-sudo resize2fs /dev/xvdb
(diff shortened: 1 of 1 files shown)


marmarek:
Yup, there was a bug with `resize2fs` call, but it is fixed for a while now.

adrelanos:
Upgrades from testing or unstable required?

So resize2fs is still needed? Should not be removed? This this PR should be rejected?

marmarek:
The `resize2fs` is called automatically, should not be required manually. Fixes already in stable for a long time: https://github.com/QubesOS/qubes-issues/issues/772
So no, this PR should be merged.

adrelanos:
Okay. While we're at it... The other three mention of resize2fs on https://github.com/adrelanos/qubes-doc/blob/patch-11/configuration/resize-disk-image.md are still required?

marmarek:
For private.img shrinking - yes. For root.img resizing - in R3.0 yes. For R3.1 - will be done automatically (QubesOS/qubes-issues#1268)

PR #60, opened 2015-11-30
Add Terminal shortcut to Fedora 23 known issues.

mattmccutchen:


axon-qubes:
I think we should instead direct the user to run `qvm-sync-appmenus` as the final step of the Template upgrade process (or make it run automatically correctly). @marmarek, what do you think?

mattmccutchen:
The longer we delay on adding this issue to the list, the more users will potentially waste time on it.  Given that we don't think the solution is as simple as "run qvm-sync-appmenus", are you (the decision maker - marmarek?) willing to just link to the issue report?

marmarek:
I think the instruction with `sed` is almost enough. It works only in case of all templates being Fedora 23, but user may have also some Debian etc. So maybe the way to go would be replacing `gnome-terminal.desktop` line with `org.gnome.Terminal.desktop\ngnome-terminal.desktop`? And ensuring that in practice only entry (that existing one) is created...

mattmccutchen:
What would trigger the replacement?  Would two entries appear in the Applications tab in Qubes Manager?  Could org.gnome.Terminal.desktop get removed from a Fedora 21 VM (either by the user removing a broken entry or automatically when an unrelated change is made), and then be missing when the VM is upgraded to Fedora 23?

The better solution might be to have qvm-sync-appmenus update the *whitelisted-appmenus.list files for a VM to match the desktop file present in the VM.

The discussion of the permanent solution really belongs on QubesOS/qubes-issues#1428.  If it's going to be several more days before the solution is released, let's go ahead and document the issue for the meantime.

marmarek:
> What would trigger the replacement?

It would be a separate point in upgrade instruction.

> Would two entries appear in the Applications tab in Qubes Manager?

I think no - it only displays entries which have the underlying files (and when you save, only those entries are stored in `whitelisted-appmenus.list`).

> Could org.gnome.Terminal.desktop get removed from a Fedora 21 VM (either by the user removing a broken entry or automatically when an unrelated change is made), and then be missing when the VM is upgraded to Fedora 23?

I think this is exactly what currently is happening. Ough, only on R3.1 - QubesOS/qubes-issues#1411

> The better solution might be to have qvm-sync-appmenus update the *whitelisted-appmenus.list files for a VM to match the desktop file present in the VM.

Theoretically yes, but there is no generic way to know that `gnome-terminal.desktop` is renamed to `org.gnome.Terminal.desktop`. Or maybe there is?

> The discussion of the permanent solution really belongs on QubesOS/qubes-issues#1428. If it's going to be several more days before the solution is released, let's go ahead and document the issue for the meantime.

You're right. (I'll copy my response there)

PR #62, opened 2015-12-02
Ported most of Whonix documentation to Docs

bnvk:
In this pull I moved over the basic Whonix + Qubes documentaiton and did some minor reorganization of things that I think are improvements:

- Moving Whonix, TorVM, and VPN pages into a section called `Privacy Guides`
- Added Troubleshooting guide about updating Debian which I ported from Whonix's site

@adrelanos should probably review and make any changes to the Whonix specfic things!


adrelanos:
We should keep the dashes inside Whonix names, i.e. Whonix***-***Gateway.

On updating-whonix.md "please click on expand on the right side" won't work.

On customizing-whonix.md the Tor bridges link won't work.

On updating-debian-and-whonix.md I don't know if `{{Anchor|...` will work.

On updating-debian-and-whonix.md, "[Whonix modular flexible .d style configuration folders]" won't work.

/attachment/wiki/Whonix/Create_Qubes-Whonix-Workstation_AppVM.png is probably still using `whonix` rather than `anon-whonix? (https://github.com/QubesOS/qubes-issues/issues/1433)

enabling apparmor: The best place to do that is right after installing the Whonix TemplateVMs. Other TemplateBasedVMs will only inherit the apparmor kernel settings if they were added to the TemplateVMs before the creation of these TemplateBasedVMs. Users who follow the install instructions, create the TemplateBasedVMs and read customizing-whonix.md later will miss those kernel settings. They could still use the command line and enable the apparmor kernel settings for those TemplateVMs also, but all of this is difficult to explain.

Under "Privacy Guides" doesn't the part "How to Install a Transparent Tor ProxyVM (TorVM)" indicate, that Whonix-Gateway cannot be used as Transparent Tor ProxyVM? Whonix-Gateway does work as Transparent Tor ProxyVM. If one wants to use TorVM, that's fine. But doing so for getting a Transparent Tor ProxyVM is a bogus reason.

bnvk:
@adrelanos thanks for the catches! I've fixed those now!

> We should keep the dashes inside Whonix names, i.e. Whonix-Gateway.

Hrm. Dunno, I see differentiating from the names in the Templates & VMs which have dashes, and the normal spoken language (not have dashes) as easier to understand.

adrelanos:
We've been consistently using the notation with dashes for years
everywhere. It's a proper noun. This isn't the right place to make such
as change.

axon-qubes:
If the hyphen is part of the name, it certainly shouldn't be removed. Consider, by analogy, people who have hyphenated surnames, e.g., "Jane Godfrey-Smith." It would be wrong to remove the hyphen in that case, since it's actually part of her name. Likewise, it would be wrong to remove the hyphen from Whonix-Gateway if the hyphen is part of that thing's name.

bnvk:
> We've been consistently using the notation with dashes for years everywhere. It's a proper noun. 

Does this matter to a Qubes user who simply wants to make their internet browsing more private and secure? I doubt it. Does it matter to devoted followers of Whonix who've read that documentation for years? Sure. 

What's the ultimate goal of Qubes by itself and w/ Whonix integrated. I think we need clarity on that and to make decisions from that POV in terms of usability.

That said, I'm kind of busy trying to get the rest of the site done, so maybe someone else can decide / change this to what is the terms. I've offered my two cents and effort.

adrelanos:
Brennan Novak:
>> We've been consistently using the notation with dashes for years
>> everywhere. It's a proper noun.
> 
> Does this matter to a Qubes user who simply wants to make their
> internet browsing more private and secure?

Yes, because everywhere inside Whonix (whonixcheck, Konsole...) it's
using the notation with dashes. And once the user looks up more
information, the user lands on whonix.org, where the notation with
dashes is being used.

> What's the ultimate goal of Qubes by itself and w/ Whonix integrated.

Integration. Not balkanization.

> I think we need clarity on that and to make decisions from that POV
> in terms of usability.

But also generating a professional impression. Inconsistent terminology
is bad. And maintainability. Questions and confusion about this eats time.

The proper noun ensured to this day, that when people people not try to
mix up Whonix-Gateway/Workstation with other gateway's or workstation'.
And when people are referencing Whonix-Gateway/Workstation in writing,
they're already including "Whonix". In a support request "some problem
with my gateway" to have to ask back "Which gateway, Whonix or
otherwise?" is bad.

On spelling rules... If we removed the dash and not use it as proper
noun, not use it as product name anymore, wouldn't we have to write
lower case 'gateway', i.e. 'Whonix gateway' rather than 'Whonix
Gateway'? Isn't the upper case version easier to read?

The notation could be changed. But then for whole Whonix, not just
Qubes-Whonix documentation. This isn't the time and place for such a change.

adrelanos:
@bnvk will you be working on the remaining issues raised in https://github.com/QubesOS/qubes-doc/pull/62#issuecomment-161467387?

bnvk:
@adrelanos no, I will not. Please change the wording to how you see fit.

adrelanos:
Done. (Except one item, I created an issue for: https://github.com/QubesOS/qubes-issues/issues/1503)

PR #78, opened 2015-12-10
Formalize shutdown step

Thomas-Fleming:
Add additional step for shutting down new Template VM and include command line input

### DIFF
--- a/managing-os/fedora-template-upgrade-21.md
+++ b/managing-os/fedora-template-upgrade-21.md
@@ -54,13 +54,9 @@ template based on the standard Fedora 21 template.
         [user@fedora-23 ~]$ sudo yum clean all
         [user@fedora-23 ~]$ sudo yum --releasever=23 --enablerepo=qubes*current-testing distro-sync
 
- 3. Shutdown the new TemplateVM via dom0 command line or Qubes VM Manager; may need to be killed.)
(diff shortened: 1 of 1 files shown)


managing-os/fedora-template-upgrade-21.md
```diff
@@ -54,11 +54,17 @@ template based on the standard Fedora 21 template.
...
 
-    (Shut down TemplateVM via Qubes VM Manager; may need to be killed.)
-
+ 3. Shutdown the new TemplateVM via dom0 command line or Qubes VM Manager; may need to be killed.)
+
+        [user@dom0 ~]$ qvm-shutdown fedora-23
+    
+    or
+    
+        [user@dom0 ~]$ qvm-shutdown --force fedora-23
```
marmarek: Did you encountered any case when `--force` is needed? I think it would be better to use `--wait`, which would wait for VM shutdown and after 60sec timeout will kill it.

axon-qubes: I went through this process on 10 TemplateVMs and never had to use `--force`. Agree that `--wait` would be safer.

Thomas-Fleming: Actually I just used "qvm-shutdown fedora-23". The reason I added the --force was because the original documentation said "may need to be killed". The point of putting the command in the documentation was that all the steps used the terminal, except this bit which said to use the Qubes VM Manager.

PR #87, opened 2016-01-03
removed extraneous characters

adrelanos:


axon-qubes:
These characters are not extraneous. They're Markdown code fences.

marmarek:
But apparently it is something wrong here:
https://www.qubes-os.org/doc/software-update-dom0/

Maybe shouldn't use both indentation and code fences?

adrelanos:
The above commands, such as `$ sudo qubes-dom0-update` do not use these code fences. Is this a wanted or accidental inconsistency?

https://www.qubes-os.org/doc/software-update-dom0/ uses `~~~` verbatim. That's probably not the indented result.

axon-qubes:
Markdown creates a code block if a line is indented by at least four spaces, *unless* it is in a list, in which case *either* eight spaces *or* code fences are required. Someone (not me) went through and used fences (`~~~`) to create code blocks here, but it seems he accidentally used *both* fences *and* indentations on those two lines under "Kernel Upgrades."

My personal preference is to use indentations to create code blocks, but of course there's nothing wrong with using fences.

> Maybe shouldn't use both indentation and code fences?

Perhaps not, but I wouldn't want to discourage contributions (or refuse them) on this basis, since both methods are valid Markdown, and our existing documentation uses both methods in many different places.

> The above commands, such as `$ sudo qubes-dom0-update` do not use these code fences. Is this a wanted or accidental inconsistency?

But they do. If you remove the fences around those lines (as your commit does), they will no longer display as code blocks. You can verify this locally yourself with jekyll.

Anyway, fixed [here](https://github.com/QubesOS/qubes-doc/commit/6f2bd69e980c588edeff6d5a7f5983f1efa51608).

PR #92, opened 2016-01-19
Update verifying-signatures.md

dan-mi-sun:
Adding missing aspect to command for verifying iso image

### DIFF
--- a/installing/verifying-signatures.md
+++ b/installing/verifying-signatures.md
@@ -115,7 +115,7 @@ The developer signing keys are set to be valid for 1 year only, while the Qubes 
 
 You can now verify the ISO image (`Qubes-R2-x86_64-DVD.iso`) matches its signature (`Qubes-R2-x86_64-DVD.iso.asc`):
 
-    $ gpg -v --verify Qubes-R2-x86_64-DVD.iso.asc
+    $ gpg -v --verify Qubes-R2-x86_64-DVD.iso.asc Qubes-R2-x86_64-DVD.iso
     gpg: armor header: Version: GnuPG v1
     gpg: assuming signed data in `Qubes-R2-x86_64-DVD.iso'
     gpg: Signature made Tue Sep 23 08:38:40 2014 UTC using RSA key ID 0A40E458


woju:
On Mon, Jan 18, 2016 at 10:55:14PM -0800, Dan Hassan wrote:
> Adding missing aspect to command for verifying iso image
> You can view, comment on, or merge this pull request online at:

If the filename of the checked file is same as signature save for .gpg
or .asc, there is no need to put it again on the command line.

-- 
regards,                        _.-._
Wojtek Porczyk               .-^'   '^-.
Invisible Things Lab         |'-.-^-.-'|
                             |  |   |  |
 I do not fear computers,    |  '-.-'  |
 I fear lack of them.        '-._ :  ,-'
    -- Isaac Asimov             `^-^-_>

dan-mi-sun:
Although this is true, being explicit means more cases will be caught?

I guess as long as people have got the gist, then they can always RTFM for GPG :)

axon-qubes:
> Although this is true, being explicit means more cases will be caught?

I'm not sure what you mean. More cases of what?

> I guess as long as people have got the gist, then they can always RTFM for GPG :)

Again, I'm not sure what you mean. What does RTFMing for GPG have to do with the fact that GPG assumes the second argument?

dan-mi-sun:
More of an edge case. Was helping a friend who had downloaded the .asc and the .iso to two different locations.

Leaving the explicit command and relying on the software assumption left them not sure what they had left out...

In this case it would simply have required them to read the GPG documentation :)

axon-qubes:
So, this depends highly on the user's level of understanding, right? A user who fails to understand why the file isn't verifying when the file and the sig are in different directories could also easily fail to understand what this means:

    $ gpg -v --verify Qubes-R2-x86_64-DVD.iso.asc Qubes-R2-x86_64-DVD.iso 

In particular, they may easily fail to understand that the first argument specifies the sig, while the second argument specifies the file to be verified.

dan-mi-sun:
True.

Would be happy to send across a PR with some more explicit explanations.

dan-mi-sun:
Though understandably it can be a rabbit hole if trying to account for all possible skill levels.

axon-qubes:
Yes, exactly. Our goal here is to help (prospective) Qubes users properly verify the Qubes ISOs they download, not to teach them how GPG works. Of course, teaching people how GPG works and how to properly use it is a worthy goal; it's just not what this particular page is meant for.

marmarek:
As [HW42 said](https://groups.google.com/d/msgid/qubes-devel/56A403D6.10706%40ipsumj.de), it may be good idea to include this change anyway:
From man gpg:

> Note: If the option --batch is not used, gpg may assume that a single
> argument is a file with a detached signature and it will try to find a
> matching data file by stripping certain suffixes.  Using this historical
> feature to verify a detached  signature is strongly discouraged; always
> specify the data file too.

axon-qubes:
Ah, good call. I missed those replies from the other thread. Merging.

PR #110, opened 2016-02-24
Update software-update-vm.md to enable rpmfusion

adubois:
Provide a lower level instruction to enable the rpmfusion repo as package source does not seem to be available any more... Also not certain, I therefore did not delete this graphical method.

### DIFF
--- a/common-tasks/software-update-vm.md
+++ b/common-tasks/software-update-vm.md
@@ -140,14 +140,19 @@ Also note that Disposable VMs do not have persistent user filesystem, and so the
 RPMFusion for a Fedora TemplateVM
 ---------------------------------
 
-If you would like to enable the [RPM Fusion](http://rpmfusion.org/) repository...
+If you would like to enable the [RPM Fusion](http://rpmfusion.org/) repository, it is recommended you clone the default fedora template and from its command line you can:
+* List repositories with `pkcon repo-list`
+* Enable a repository with `pkcon repo-enable rpmfusion-free`
+* Disable a repository with `pkcon repo-disable rpmfusion-free`
 
-dom0 Start Menu -> Template:Fedora 21 -> Package Sources -> Enable RPMFusion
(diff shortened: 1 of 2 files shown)


axon-qubes:
The existing instructions are puzzling. Was "Package Sources" some program in F21? (I don't have it in my F23 template.)

The new instructions look better to me. @marmarek?

marmarek:
> Was "Package Sources" some program in F21?

Yes. See QubesOS/qubes-issues#1716

marmarek:
As for the instruction it's ok. There is also PackageKit commands to do that (to not deal with manually editing files):
pkcon repo-list
pkcon repo-enable
pkcon repo-disable

Not sure if worth mentioning? Maybe yes?

marmarek:
(this is what "Software" - gnome GUI for packages - does use)

adubois:
I tried first with a GUI based tool (Software, Package-Updater) but was not able to set-up RPMFusion.
I've tried to look at what Software and package-kit could offer but so far I found these 2 pages that seem to suggest that it is not managed by high level package managers and that dnf / yum must be used... I never use high level tools for that so can't really trust my findings but to me the command line is the only option I found.
https://ask.fedoraproject.org/en/question/38388/can-software-list-packages-from-rpm-fusion/
http://ankursinha.in/blog/2015/05/28/appstream-data-for-rpmfusion-now-available.html
I am going to give a last shot at what is stated on the rpmfusion configuration page and will report...
http://rpmfusion.org/Configuration

adubois:
The instructions on RPMFusion (non command line) expect Firefox to be able to download the rpm which it can not. The instructions using pkcon are the most straight forward and I believe can be merged.

axon-qubes:
What's the status of this PR?

marmarek:
I've already merged #114 which uses more straightforward approach. Also doesn't mention need to verify keys - those are already provided with the template.

PR #114, opened 2016-03-16
update wrong RPMFusion repo enabling commands

Galland:
Tested on fedora-23

### DIFF
--- a/common-tasks/software-update-vm.md
+++ b/common-tasks/software-update-vm.md
@@ -140,10 +140,9 @@ Also note that Disposable VMs do not have persistent user filesystem, and so the
 RPMFusion for a Fedora TemplateVM
 ---------------------------------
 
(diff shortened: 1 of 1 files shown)


Galland:
In the current state of RPMFusion, people may want (temporarily) to also enable the updates-testing repo, since most apps are not yet in stable:
sudo dnf config-manager --set-enabled rpmfusion-free-updates-testing rpmfusion-nonfree-updates-testing

marmarek:
Rpmfusion repo definitions are already installed. The repository is simply disabled. See https://github.com/QubesOS/qubes-doc/pull/110#issuecomment-189216306
This means the `dnf install` step is unnecessary (BTW `--nogpgcheck` isn't quite a good idea, especially with plain HTTP...).

Galland:
Thanks for the info! I changed commit to match it

PR #119, opened 2016-03-28
Add missing available options.

northox:
Sync with the cli.


axon-qubes:
@marmarek, are these pages automatically generated, or does it make sense to make changes here?

northox:
Oh. Here? https://github.com/QubesOS/qubes-core-admin/blob/master/doc/qvm-tools/qvm-backup-restore.rst

marmarek:
Yes, there. Then AFAIR converted rst->md using `pandoc`.

PR #120, opened 2016-03-31
Update dispvm-customization.md

privacyguy:
added additional directions for creating a customized dispvm from a customized template


axon-qubes:
@marmarek Should the `--default-script` option be preserved here?

marmarek:
Since R2 it is optional (no second argument is the same as `--default-script`). So, it's fine to drop it here. Maybe even drop it on the other line for consistency.

PR #122, opened 2016-04-15
remove librem 15 section

mfc:
we don't list laptops that aren't certified. 

anyone can sell a laptop with Qubes pre-installed on it. If we want to make a list of that then let's do that, maybe elsewhere.

### DIFF
--- a/hardware/certified-laptops.md
+++ b/hardware/certified-laptops.md
@@ -46,13 +46,6 @@ compatibility with Qubes:
 
        sudo qubes-dom0-update --enablerepo=qubes-dom0-unstable kernel
 
-Purism Librem 15
-----------------
-
-While it is also currently possible to order a Librem 15 with Qubes
-pre-installed, the Librem 15 is not yet officially certified. However, testing
-is underway, and we expect to be able to certify it soon.
-
 
 [System Requirements]: /doc/system-requirements/
 [Hardware Compatibility List]: /hcl/


axon-qubes:
Here was my reasoning behind adding that section, FWIW:

The Librem 13 is already certified, and the only immediately obvious difference between the Librem 13 and the Librem 15 is that one is smaller and one is larger. So, someone who sees that the Librem 13 is certified, then goes to the Purism website and sees that the Librem 15 also exists will naturally wonder, "Well, what about the Librem 15? Is that one also certified, or what?"

I think that's a reasonable question to ask. Even *I* didn't know the answer and had to ask (#1902). Prior to receiving the answer, I figured that they were probably both certified, but that we just hadn't had a chance to update our Certified Laptops page yet. It turned out that that's not the case. I thought other users might reasonably follow the same thought process I did, so I figured it would be helpful for us to anticipate their need for this information rather than just remaining completely silent about it and withholding our knowledge on the subject.

Now, it's true, any manufacturer can sell a laptop with the option to pre-install Qubes. If we make a section for *this* laptop which isn't certified, won't we have to start making sections for *all* the ones that aren't certified? I don't think so. I think the Librem 15 is different from those other hypothetical laptops for several reasons:

 * It's a different size/version of the only laptop which is currently certified, so there is a reasonable presumption that it could or might be certified, whereas there is no such reasonable presumption when it comes to other random laptops.
 * Qubes is already in some sort of official partnership with the manufacturer, so this is not just some random, unaffiliated manufacturer.
 * It is already in the process of being certified. (Qubes devs currently waiting and expecting to receive a unit to test; already have expectations and plans about what needs testing.) This is, AFAIK, not true of any other laptops. (If it is, then I think those should be added, as well. After all, it might affect a user's decision whether to purchase a laptop now or wait for one that's about to be certified soon.)

mfc:
not sure this needs to be that complicated. laptops get listed there when they are certified. if they are not certified, they are not on the list. the navigation of the certified laptops page on first blush makes it look like librem 15 is certified, since the page is supposed to be a list of certified laptops.

if you want to list laptops that are in the process of getting certified, or can be bought with Qubes pre-installed, then create separate lists/pages for those. that would probably be useful for users.

"different size/version" means different laptop with different hardware which may or may not work, hence certification. some librem 15 users have mentioned issues with the screen and it can only be troubleshooted when we [have a librem 15 to troubleshoot](https://www.qubes-os.org/hardware-certification/).

marmarek:
IMO Librem 15 isn't worth separate section from the reasons Michael
mentioned. But maybe worth a note inside "Librem 13" section.

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

PR #125, opened 2016-04-17
Update fedora-minimal.md

lambono:
Added useful info in case user wants to use Fedora-minimal template for a gpg vm in Split GPG scenario.

### DIFF
--- a/managing-os/templates/fedora-minimal.md
+++ b/managing-os/templates/fedora-minimal.md
@@ -48,7 +48,7 @@ To access the journald log, use the `journalctl` command.
 
 ### as a NetVM
 
-If you want to use this template to for standard NetVMs you should install some more packeges:
+If you want to use this template for standard NetVMs you should install some more packages:
 
 ~~~
 [user@F21-Minimal ~]$ sudo dnf install NetworkManager NetworkManager-wifi network-manager-applet  wireless-tools dbus-x11 dejavu-sans-fonts tinyproxy
@@ -79,3 +79,8 @@ The needed packages depend on the VPN technology. The `dnf search "NetworkManage
 #### TOR
 
 [UserDoc/TorVM](/wiki/UserDoc/TorVM)
+
+### Split GPG
+
+If you choose to use this Fedora minimal template for a machine holding your gpg keys (so called work-gpg machine in [Split GPG] (https://www.qubes-os.org/doc/split-gpg/) guide) and also you choose to use passphrase on your gpg private key, then you should install `pinentry-gtk` package. Package `pinentry-gtk` is responsible for pop-up dialog window where you enter password to use your gpg key.
+


axon-qubes:
Thanks! Just a couple notes:

* I think if we're going to mention Split GPG at all, we need to mention that you have to install the actual Split GPG package to make it work.
* (Minor) Please use relative links (`/doc/split-gpg/`).

@marmarek, can you confirm whether the `pinentry-gtk` package is strictly necessary for this?

marmarek:
On Sun, Apr 17, 2016 at 05:45:40PM -0700, Axon wrote:
> @marmarek, can you confirm whether the `pinentry-gtk` package is strictly necessary for this?

I don't think so. It may be useful in case of password protected keys,
but I don't think its necessary in case of Split GPG.

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

marmarek:
On the other hand, it may be good idea to recommend minimal template usage in Split GPG docs.

axon-qubes:
Any particular reason why? Or just the general reason of having a reduced attack surface? It seems like that reason could apply to any VM with a particular function. In other words, as a general rule, "Use as minimal of a template as possible to keep the attack surface minimal." But of course this has to be weighed against the costs of disk space and having to update and maintain multiple templates.

In the case of Split GPG, we already have a very tightly controlled interface between the frontend and backend domains, and GPG has to be running there no matter what (and GPG seems to provide the main attack surface), so it's unclear to me how big the advantage is to using a minimal template here. (In particular, whether the advantage is so significant that we should explicitly recommend it instead of using the clean, non-minimal fedora-23 default template, which already comes with split-gpg installed.)

marmarek:
You are right, there is no much security gain from using minimal
template for gpg backend. Besides general rule that more specialised VMs
are better.
But I think there is noticeable gain in term of resources (memory).

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

axon-qubes:
For me, running it on `fedora-23-minimal` instead of `fedora-23` only saves ~70MB of memory, but I have to give up ~2200MB of disk space for the dedicated template. I suppose it's a personal choice at that point.

marmarek:
Agree.

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

lambono:
I'm using fedora minimal template as a template for all of my machines. This way I save SSD disk space (for me the minimal template with additional packages (libreoffice, ...)  takes around 2,5GB of disk space instead of around 7G) and I have less frequent updates (saving data on mobile connection). As I understood reason for the existence of fedora minimal template is that people can start from  minimal install (like debian minimal installation option available when installing) and configure it further according their needs. I also do not see it as a way to lessen attack surface much.
So maybe it would be good for others to have somewhere this information, instead of wasting time to make it work.
Also I would add, if one would like to use "Send to vm" context menu in fedora minimal, it has to install additional package "pycairo".

andrewdavidwong:
> So maybe it would be good for others to have somewhere this information, instead of wasting time to make it work.

Yes. The appropriate place is actually this one: https://www.qubes-os.org/doc/fedora-minimal-template-customization/

Would you mind submitting a pull request adding your tips to that page?

-----

Just to clarify why this shouldn't be added to the *main* `fedora-minimal` page:

The problem is that this information is useful only to a very small percentage of users who *intentionally choose* to take the DIY route of starting with the minimal template and adding functionality which already exists in the standard template. Users who choose to take this DIY route should expect that they will have to learn how to make things work. (If they don't want to do that, they can just use the standard template, where things already work.)

If we were to include every small tip and instructions for installing every possibly-useful package, the documentation would become much harder to use. A large portion of it would not be useful to most readers. They would have to slog through the minutiae just to find the generally useful information.

lambono:
Firstly, thank you for providing me the correct URL.  Indeed my proposals fits much better into customization section (whose existence I was unaware of ;). I'll submit a pull request there.

PR #126, opened 2016-04-18
Arch linux template building

tezeb:
update instructions for Qubes 3.1

Fix for: qubes-core-vchan-xen - missing in marmarek repo and required to build Qubes for 3.1.

Also *qubes-attachments* updated.

### DIFF
--- a/managing-os/templates/archlinux.md
+++ b/managing-os/templates/archlinux.md
@@ -119,11 +119,11 @@ Main maintainer of this template is [Olivier Médoc](mailto:o_medoc@yahoo.fr).
    
     *In the future this should not be needed once a change is made to the 'setup' script.*
 
-    *   Edit the '**qubes-os-master.conf**' which is found in **/home/user/qubes-builder/example-configs**  Use the text editor of your choice.
+    *   Edit the '**qubes-os-r3.1.conf**' which is found in **/home/user/qubes-builder/example-configs**  Use the text editor of your choice.
     
         *   **$ cd /home/user/qubes-builder/example-config/**
     
-        *   **$ nano -W qubes-os-master.conf** or **$ gedit qubes-os-master.conf** or etc….
+        *   **$ nano -W qubes-os-r3.1.conf** or **$ gedit qubes-os-r3.1.conf** or etc….
 <br>
 <br>
 ![arch-template-06](/attachment/wiki/ArchlinuxTemplate/arch-template-06.png)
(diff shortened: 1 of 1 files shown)


axon-qubes:
@marmarek

managing-os/templates/archlinux.md
```diff
@@ -204,20 +210,13 @@ Main maintainer of this template is [Olivier Médoc](mailto:o_medoc@yahoo.fr).
...
 
-        *   Screen '**Get Resources**' wants to download additional packages needed for the choosen plugin/s.
+        *   Screen '**Get sources**' wants to download additional packages needed for the choosen plugin/s.
         
-            *   Select '**OK**' Prss '**Enter**'
-            
-        *   Upon completion you will get choose '**OK**' to proceed to the next screen
-        
-            *   Press '**Enter**'
+            *   Select '**No**' Press '**Enter**'
```
marmarek: Why "No"?

marmarek:
I don't think PR description is accurate, but the content looks ok (besides minor question, but not a blocker).

axon-qubes:
Ok, I'll give @tezeb a chance to reply before merging.

managing-os/templates/archlinux.md
tezeb: Sorry I should have been more verbose about it.

This(+ setting _GIT_URL_builder_archlinux_ later) is actually the main reason for the whole change.
Choosing "Yes" here tries to fetch "qubes-builder-archlinux.git" from main QubesOS repository. As it does not exist, it asks for username/password for github(probably to check whether private repo with such name exists). Firstly this is at least confusing, considering that user tries to download from public repo. And the whole downloading step fails afterwards as it cannot access repo.

Additionally as far as I understand this step is done anyway with _make get-sources_, which is done after manually fixing _builder.conf_ and works like a charm.

marmarek: Ok.

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

axon-qubes: @marmarek: Ok to merge?

marmarek: > @marmarek: Ok to merge?

Mhm, :+1:

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

tezeb:
@axon-qubes I'am not sure if you can do this, but there are also changes to qubes-attachment that are referenced in this commit.

axon-qubes:
@tezeb: Ah, thanks for the reminder. That PR has now also been merged.

PR #127, opened 2016-04-19
Update full-screen-mode.md

gutsle:
explicitly mention that alt+tab protects against fake window attacks

### DIFF
--- a/common-tasks/full-screen-mode.md
+++ b/common-tasks/full-screen-mode.md
@@ -25,7 +25,7 @@ Secure use of full screen mode
 ------------------------------
 
 However, it is possible to deal with full screen mode in a secure way assuming there are mechanisms that can be used at any time to show the full desktop, and which cannot be intercepted by the VM. An example of such q mechanism is the KDE's "Present Windows" and "Desktop Grid" effects, which are similar to Mac's "Expose" effect, and which can be used to immediately detect potential "GUI forgery", as they cannot be intercepted by any of the VM (as the GUID never passes down the key combinations that got consumed by KDE Window Manager), and so the VM cannot emulate those. Those effects are enabled by default in KDE once Compositing gets enabled in KDE (System Settings -\> Desktop -\> Enable Desktop Effects), which is recommended anyway. By default they are triggered by Ctrl-F8 and Ctrl-F9 key combinations, but can also be reassigned to other shortcuts.
-Another option is to use Alt+Tab for switching windows. This shortcut is also handled by dom0 and protects from fake windows. However one needs to pay careful attention, two windows switching when pressing Alt+Tab once means one of the windows is fake.
+Another option is to use Alt+Tab for switching windows. This shortcut is also handled by dom0.
 
 Enabling full screen mode for select VMs
 ----------------------------------------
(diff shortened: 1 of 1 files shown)


common-tasks/full-screen-mode.md
```diff
@@ -25,6 +25,7 @@ Secure use of full screen mode
 ------------------------------
 
 However, it is possible to deal with full screen mode in a secure way assuming there are mechanisms that can be used at any time to show the full desktop, and which cannot be intercepted by the VM. An example of such q mechanism is the KDE's "Present Windows" and "Desktop Grid" effects, which are similar to Mac's "Expose" effect, and which can be used to immediately detect potential "GUI forgery", as they cannot be intercepted by any of the VM (as the GUID never passes down the key combinations that got consumed by KDE Window Manager), and so the VM cannot emulate those. Those effects are enabled by default in KDE once Compositing gets enabled in KDE (System Settings -\> Desktop -\> Enable Desktop Effects), which is recommended anyway. By default they are triggered by Ctrl-F8 and Ctrl-F9 key combinations, but can also be reassigned to other shortcuts.
+Another option is to use Alt+Tab for switching windows. This shortcut is also handled by dom0 and protects from fake windows. However one needs to pay careful attention, two windows switching when pressing Alt+Tab once means one of the windows is fake.
```
axon-qubes: IMHO, we should soften the language here a bit. There are probably a lot of edge cases in which it's technically true that "two windows switch" even though one of them isn't exactly "fake." I would just add the word "could" ("*could* mean one of the windows is fake").

@marmarek, what do you think?

marmarek: Something is unclear in this sentence. Which two windows? The only two
visible? In such a case Alt-Tab switching between them means it's ok.
But when you see two windows but Alt-Tab doesn't switch between them (or
switch to some other window), something is wrong.

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

gutsle:
Hm... indeed on second read it's not very clear. What I meant was if you press Alt+Tab you must be able to get every single window to the top. If a window always comes up with another window something is wrong.
But maybe that's not actually true. Axon, what would be valid cases when two windows switch at once?

axon-qubes:
> Axon, what would be valid cases when two windows switch at once?

Silly example: Your window happens to be a Qubes screenshot showing two differently-colored windows, so it looks like two windows are moving at once. (Neither one is *fake*; it's just a picture of two windows. By analogy, it would be wrong to say that a photograph of you is "a fake gutsle." "Fakeness" in this sense implies something intentionally designed to appear genuine.)

Serious example: In GNOME, you can decide whether alt+tab groups windows or treats them individually. By default, it groups them.

The point is that we cannot know for certain that there will never be a case in which (a) two windows appear to move together (or actually do move together), yet (b) neither one is a fake window (especially since this depends on details of the DE which could change).

gutsle:
Okay. I mostly agree except that
1. The current default in (at least XFCE-based) qubes is to switch only one window.
2. If that's not the case Alt+Tab is no longer viable for making sure a window is real and XFCE doesn't have the window presentation feature.

Of course 2. is not really an argument, just motivation. So...hm.

Maybe we should just mention that Alt+Tab is also handled by dom0 and leave it to the user to figure out how to do it safely. That just made me a little uneasy but I suppose it's fine.

axon-qubes:
Yeah, it's not that big of a deal. That's why I was merely advocating for the addition of the word "could". :)

gutsle:
Alright, I removed the last parts.

PR #134, opened 2016-05-16
Start for Kali guide

Vfreeze31:
I have started something for the Kali guide.
I am sorry as I am not able to test at the moment as my system is not setup the way I would like at the moment.


andrewdavidwong:
Thanks, @Vfreeze31, this looks like a good start!

PR #138, opened 2016-05-19
how to avoid creating an ubuntu 14.4 LTS template

john-david-r-smith:
maybe this method should be added as a general point.
e.g.: "Simulating qubes features for vms without qubes support"

a similar thing could be done for windows hvms, too.
you could use teamviewer or vnc and get audio this way (though i don't know whether vnc can stream audio).

### DIFF
--- a/managing-os/templates/ubuntu.md
+++ b/managing-os/templates/ubuntu.md
@@ -31,6 +31,12 @@ want to build.
 
 Known issues
 ------------
+Building an [Ubuntu 14.4 LTS](https://groups.google.com/forum/#!topic/qubes-users/w0uZNr8nno8) template can be difficult.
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
IMHO, the `Ubuntu 14.4 LTS` link is a bit misleading. Would you mind cleaning that up a bit? (You could just use "see" like you do for the later ML link.)

john-david-r-smith:
like this?

andrewdavidwong:
The link is better, but please don't use HTML for line breaks when you can make them with Markdown (two spaces at the end of the line).

john-david-r-smith:
done. (i never stop learning)

andrewdavidwong:
Thanks!

PR #142, opened 2016-05-22
Remove ref to qubes-windows-tools in current repo

ryanchapman:
Hi,

qubes-windows-tools currently only exists in the dom0 current-testing repo.  According to the dom0 file /etc/yum.repos.d/qubes-dom0.repo, it seems this is by design.  If you would prefer to add the package to the current repo instead of this change, that is fine by me.  I just wanted to note this in the docs for the next person.

Best regards,
Ryan

`curl -s yum.qubes-os.org/r3.1/current/dom0/fc20/rpm | grep qubes-windows-tools`
returns nothing

However, current-testing does have it in there:
```
curl -s yum.qubes-os.org/r3.1/current-testing/dom0/fc20/rpm | grep qubes-windows-tools
<li><a href="qubes-windows-tools-3.0.3-1.x86_64.rpm"> qubes-windows-tools-3.0.3-1.x86_64.rpm</a></li>
<li><a href="qubes-windows-tools-3.0.4-1.x86_64.rpm"> qubes-windows-tools-3.0.4-1.x86_64.rpm</a></li>
```

### DIFF
--- a/managing-os/windows-appvms.md
+++ b/managing-os/windows-appvms.md
@@ -38,13 +38,7 @@ Installing Qubes guest tools in Windows 7 VMs
 First, make sure that `qubes-windows-tools` is installed in your system:
 
 ~~~
-sudo qubes-dom0-update qubes-windows-tools
-~~~
-
-You can also install the package from testing repositories, where we usually publish new versions first:
-
-~~~
-qubes-dom0-update --enablerepo=qubes*testing qubes-windows-tools
+sudo qubes-dom0-update --enablerepo=qubes*testing qubes-windows-tools
 ~~~
 
 This package brings the ISO with Qubes Windows Tools that is passed to the VM when `--install-windows-tools` is specified for the `qvm-start` command. Please note that none of this software ever runs in Dom0 or any other part of the system except for the Windows AppVM in which it is to be installed.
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
I think this is just temporary, since @omeg is currently working on Windows Tools support for Windows 8.x/10. Is that right, @omeg?

omeg:
That's correct, windows tools will be in the stable repository eventually. They're not stable enough yet though.

andrewdavidwong:
Ok. In that case, I think it's fair to say that this is what we're doing (over an indeterminate period of time):

> If you would prefer to add the package to the current repo instead of this change, that is fine by me.

ryanchapman:
Does it make sense to change the documentation (in this pull request) so that other users know that the package isn't in the stable repo?  It took me about 15 minutes to figure out the documentation was incorrect and that the command `sudo qubes-dom0-update qubes-windows-tools` is not expected to work.

andrewdavidwong:
Sure, but I think both commands should be left in (and explained). Otherwise, the documentation will go out of date again once Windows Tools move to the stable repo.

ryanchapman:
Not sure if you'll like the verbiage, but I changed this PR to just add a note to try the testing repo if the stable repo does not work.  If you'd like it worded differently, please let me know.

Thanks
Ryan

andrewdavidwong:
That's great! Thanks!

andrewdavidwong:
Merged: https://github.com/QubesOS/qubes-doc/commit/3239b9c4e5719cd15e9a9003da23f114b55933b7

PR #147, opened 2016-05-25
how to setup an openvpn connection using iptables

john-david-r-smith:
everything should be correct (maybe sb else should check it to make sure, since it is security critical)

### DIFF
--- a/configuration/vpn.md
+++ b/configuration/vpn.md
@@ -29,6 +29,7 @@ While the NetworkManager service is not started here (for a good reason), you ca
 
 ### ProxyVM
 
+
 **WARNING:** *You need to use Qubes 3.1-rc2 (or later)! In the previous releases the NetworkManager service was not working in ProxyVMs as expected.* ([#1052](https://github.com/QubesOS/qubes-issues/issues/1052))
 
 One of the best thing in Qubes is that you can use a special type of VM called a ProxyVM (or FirewallVM). The special thing is that your AppVMs see this as a NetVM, and your NetVMs see it as an AppVM. Because of this, you can place a ProxyVM between your AppVMs and your NetVM. This is how the default FirewallVM functions.
@@ -39,7 +40,9 @@ Using a ProxyVM to set up a VPN client gives you the ability to:
 -   Separate your VPN credentials from Your AppVM data.
 -   Easily control which of your AppVMs are connected to your VPN by simply setting it as a NetVM of the desired AppVM.
 
-**To setup a ProxyVM as a VPN gateway you should:**
+#### Setup a ProxyVM as a VPN gateway
+
+**Using NetworkManager**
 
 1.  Create a new VM and check the ProxyVM radio button.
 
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
@marmarek

configuration/vpn.md
```diff
@@ -56,3 +59,104 @@ Using a ProxyVM to set up a VPN client gives you the ability to:
...
+
+    It should have one line starting with `dev` and one starting with `proto`.
+    The first describes the connection type (`tun` or `tap`) and the second the used protocol (`tcp` or `udp`).
+    Depending on your connection type, openvpn will create a new network device (probably `tap0` or `tun0`).
+
+    It also contains a line `remote X.X.X.X 1194`, where `X.X.X.X` is the ip of your openvpn server.
+
+    If it does not contain a line `redirect-gateway def1`, add it.  
+    This will route all traffic through your vpn's network device, after a connection was created.
+    If the connection breaks down all traffic will be routed through the original network device (we will top this with iptables).
```
marmarek: s/top/stop/

configuration/vpn.md
```diff
@@ -56,3 +59,104 @@ Using a ProxyVM to set up a VPN client gives you the ability to:
...
+    
+    `iptables -I OUTPUT -o lo -j ACCEPT`  
+    This allows connections from the system to the system.
+
+    `iptables -I FORWARD -o eth0 -j DROP`  
+    `iptables -I FORWARD -i eth0 -j DROP`  
+    This blocks forwarding of connections through your plain network device (in case the vpn tunnel breaks).
+    
+    `iptables -I FORWARD -o $DEV -j ACCEPT`
+    This allows forwarding of connections through the vpn. (So other AppVMs can use it)
```
marmarek: This effectively disable all the firewall set for connected VMs.

configuration/vpn.md
```diff
@@ -56,3 +59,104 @@ Using a ProxyVM to set up a VPN client gives you the ability to:
...
+        systemctl start user_shutdown_hook;
+
+    to `/rw/config/rc.local`.  
+    This creates and starts a service executing the script `/rw/config/rc.local.shutdown` on shutdown.
+    The service will be deleted after each shutdown, since the `root.img` gets reset at each reboot.
+
+    Now create `/rw/config/rc.local.shutdown`, make it executable `chmod +x  /rw/config/rc.local.shutdown` and put the kill command in it:
+
+        #!/bin/bash
+        killall openvpn
```
marmarek: If going to use systemd anyway, better start openvpn itself from systemd. There is even already provided service file (`/usr/lib/systemd/system/openvpn@.service`), which expects config file in `/etc/openvpn/`. So all what is needed should be copying (or better: symlinking) needed files there, then start it: `systemctl --no-block start openvpn@config-name.service`

configuration/vpn.md
john-david-r-smith: i guess you mean line 115.
i did not notice, since no appvm behind my vpn currently have rules.
how would you suggest to fix this? (if i omit the line my appvms can't get a connection)

marmarek: Omitting should be fine. If not, take a look what rule does block the traffic (I recommend looking at couters in `iptables -nvxL`). Maybe only DNS is blocked?

configuration/vpn.md
john-david-r-smith: ok done.
this does fix the problem with shutting down (but i guess it was caused, because i was not using the `--daemon` flag)

configuration/vpn.md
john-david-r-smith: must have been something like this.
i removed it and everything works.

marmarek:
Now it's ok.

PR #160, opened 2016-06-05
Automatic; No manual coding of IP addresses.

ttasket:
This requires the user only to add a few lines to their ovpn config file, and copy a few scripts (verbatim). They do not have to figure out which IP addresses are appropriate and hard-code them--unless their VPN service is bereft of domain names. Even in that case, they can do it easily within the ovpn config file. This is much less error-prone and should work with a greater variety of services (large commercial services tend to change their IPs so using domain names and DHCP is preferable in that case).

Also converted firewall section (3) to one code block for much less cutting/pasting. Comments are still there as shell comments.

The only required template changes are adding openvpn itself and possibly disabling the default systemd service for it. Everything else should be there in /rw/config.

This doesn't include extra firewall protections against inadvertent net access from within the VPN VM. I'm thinking of proposing those additions in a separate edit.

### DIFF
--- a/configuration/vpn.md
+++ b/configuration/vpn.md
@@ -121,7 +121,7 @@ Using a ProxyVM to set up a VPN client gives you the ability to:
     ```
     #!/bin/bash
     set -e
-    export PATH=$PATH:/usr/sbin:/sbin
+    export PATH="$PATH:/usr/sbin:/sbin"
 
 case "$1" in
 
--- a/configuration/vpn.md
+++ b/configuration/vpn.md
@@ -42,7 +42,7 @@ Using a ProxyVM to set up a VPN client gives you the ability to:
 
 #### Setup a ProxyVM as a VPN gateway
 
-**Using NetworkManager**
+#### Using NetworkManager
 
 1.  Create a new VM and check the ProxyVM radio button.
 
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
@marmarek

configuration/vpn.md
```diff
@@ -60,73 +60,128 @@ Using a ProxyVM to set up a VPN client gives you the ability to:
...
+	#    First, block all outgoing traffic
+	iptables -P OUTPUT DROP
+    iptables -F OUTPUT
+
+	#    Add the `qvpn` group to system, if it doesn't already exist
+    if ! grep -q "^qvpn:" /etc/group ; then
+        groupadd -rf qvpn
+        sync
+    fi
+    sleep 2s
```
marmarek: Why?

configuration/vpn.md
```diff
@@ -60,73 +60,128 @@ Using a ProxyVM to set up a VPN client gives you the ability to:
...
+
+	#    Allow traffic from the `qvpn` group to the uplink interface (eth0);
+	#    Our openvpn will run as group `qvpn`.
+    iptables -A OUTPUT -p all -o eth0 -m owner --gid-owner qvpn \
+    -m state --state NEW,ESTABLISHED -j ACCEPT
```
marmarek: Any particular reason for including state when both NEW and ESTABLISHED are included? If you want to exclude INVALID or sth like this, better write to it explicitly with `! --state INVALID`. Otherwise the rule can be simplified to:
`iptables -A OUTPUT -o eth0 -m owner --gid-owner qvpn -j ACCEPT`

configuration/vpn.md
```diff
@@ -60,73 +60,128 @@ Using a ProxyVM to set up a VPN client gives you the ability to:
...
+4.  Create the DNS-handling script.
+    Use `sudo nano /rw/config/openvpn/qubes-vpn-handler.sh` to edit and add:
+    ```
+    #!/bin/bash
+    set -e
+
+# Pop-up notification variables
+SPID=$(pgrep -U user -f dconf-service)
+dbus=$(grep -z DBUS_SESSION_BUS_ADDRESS /proc/$SPID/environ|cut -d= -f2-)
+export DBUS_SESSION_BUS_ADDRESS=$dbus
```
marmarek: You can get user environment by sourcing `/tmp/qubes-session-env`. Or add `-` to `su` invocation below - to load login scripts, which will source that file.

configuration/vpn.md
```diff
@@ -60,73 +60,128 @@ Using a ProxyVM to set up a VPN client gives you the ability to:
...
+
+case "$1" in
+
+up)
+	# To override DHCP DNS, assign static DNS addresses with 'setenv vpn_dns' in openvpn config;
+	# Format is 'X.X.X.X  Y.Y.Y.Y [...]' with quotes.
+	if [[ -z $vpn_dns ]] ; then
```
marmarek: Quote variable.

configuration/vpn.md
```diff
@@ -60,73 +60,128 @@ Using a ProxyVM to set up a VPN client gives you the ability to:
...
+		# Parses DHCP options from openvpn to set DNS address translation:
+		for optionname in ${!foreign_option_*} ; do
+			option="${!optionname}"
+			unset fops; fops=($option)
+			if [ ${fops[1]} == "DNS" ] ; then vpn_dns="$vpn_dns ${fops[2]}" ; fi
+		done
+	fi
+
+	iptables -t nat -F PR-QBS
+	if [[ -n $vpn_dns ]] ; then
```
marmarek: Quote variable.

configuration/vpn.md
```diff
@@ -60,73 +60,128 @@ Using a ProxyVM to set up a VPN client gives you the ability to:
...
+```
+
+    Now save the script and make it executable:  
+    `sudo chmod +x /rw/config/openvpn/qubes-vpn-handler.sh`
+    
+5.  Setup the VPN's autostart:  
+    Use `sudo nano /rw/config/rc.local` to edit and add:  
+    ```
+    #!/bin/bash
+    groupadd -rf qvpn ; sleep 2s
```
marmarek: Same here - is `sleep 2s` really necessary?

configuration/vpn.md
ttasket: I suppose, if INVALID is the only other state...?

configuration/vpn.md
ttasket: That's most of the doc up there... Why the revision?

configuration/vpn.md
```diff
@@ -60,73 +60,128 @@ Using a ProxyVM to set up a VPN client gives you the ability to:
...
+    sleep 2s
+
+	#    Allow traffic from the `qvpn` group to the uplink interface (eth0);
+	#    Our openvpn will run as group `qvpn`.
+    iptables -A OUTPUT -p all -o eth0 -m owner --gid-owner qvpn \
+    -m state --state NEW,ESTABLISHED -j ACCEPT
+
+	#	Allow queries to DNS server:
+    iptables -A OUTPUT -p udp -o eth0 --dport 53 -m state --state NEW,ESTABLISHED -j ACCEPT
+    iptables -A OUTPUT -p tcp -o eth0 --dport 53 -m state --state NEW,ESTABLISHED -j ACCEPT
```
ttasket: Appears to be common practice. I think specifying what is allowed is generally tighter (or at least more reassuring).

configuration/vpn.md
ttasket: Excellent! I'll try that right away.

configuration/vpn.md
marmarek: It's about `sleep 2s`.

configuration/vpn.md
ttasket: This one was done more out of response to the firewall script (above) -- making sure the new group was recognized.

configuration/vpn.md
marmarek: According to manual page, there is also RELATED and UNTRACKED. I don't see any good reason for excluding them. Anyway, besides some performance impact (probably negligible), it shouldn't be a problem, so may be left as it is.

configuration/vpn.md
marmarek: Who would perform those DNS queries? If openvpn, it should be already handled by the previous rule.

configuration/vpn.md
ttasket: Also, can you tell me if `rc.local` and `qubes-firewall-user-script` are run in a certain order, or independently?

marmarek: Independently.

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

marmarek: Especially first call of `qubes-firewall-user-script` probably will be before `rc.local`, but subsequent calls after.

configuration/vpn.md
ttasket: OK, iptables --gid-owner doesn't see the group right away.

marmarek: It would be good idea to find out why (maybe require some cache flush or so). But for now `sleep 2` is ok.

configuration/vpn.md
ttasket: I wondered if DNS queries acquired the user/owner of the requesting process. Now I know. :)

As for 'who else' would perform queries... I don't know, but sentiment on the lists and issues thread are angling for maximum local restriction (its why you see other proposals making user hard-code addresses into firewall).

I'm glad to remove these if DNS still works.

configuration/vpn.md
ttasket: To clarify... if iptables doesn't see the group yet, the command will fail.

ttasket: If you mean `sync` to flush cache, I already tried that without the sleep. Didn't work.

configuration/vpn.md
ttasket: It works! Going to do the edits now.

PR #161, opened 2016-06-05
fix wiki rendering of 4th point's bash code

Scinawa:
added spacing and tabs


andrewdavidwong:
If you test this commit locally with Jekyll, you can see that it doesn't actually fix the formatting issue, so I'm going to close this with a different commit. Thank for bringing this to our attention, though!

PR #166, opened 2016-06-13
Add second method for MAC randomisation

unman:


andrewdavidwong:
@marmarek

privacy/anonymizing-your-mac-address.md
```diff
@@ -170,3 +172,48 @@ To disable MAC Randomizing if you find that a network connecting to does not lik
...
+
+This method does not require you to specify interface names, will apply to all interfaces, even those added after boot, and will retain the random MAC address through sleep/wake cycles.
+
+Install **macchanger** package in the TemplateVM backing sys-net.
+
+In that template create a udev rule using an editor of your choice.
+
+In /etc/udev/rules.d create the file 00-mac-spoof.rules with content:
+```
+SUBSYSTEM=="net", ACTION=="add", ATTR{type}=="1", RUN+="/usr/bin/macchanger -e  $name"
```
marmarek: This rule should exclude `vif*` interfaces and probably also `eth0` in VMs. IOW should randomize MAC only on real network hardware, not inter-VM links.

unman: @marmarek 
Can I ask why? It's not hard to do but I'd like to understand why it should be restricted as you suggest.

marmarek: To not mess with other configuration. For example `vif*` interfaces are excluded from NetworkManager using mac address (the only supported option): https://github.com/QubesOS/qubes-core-agent-linux/blob/master/network/network-manager-prepare-conf-dir#L11-L16
Otherwise NetworkManager would try to run DHCP client on downlinks to the VMs...
As for `eth0` in VMs (uplink to netvm), there is already setting for that (available via `qvm-prefs`), so users may expect to have this setting respected. In practice this is much more useful for HVM (and Windows there, to not complain about different MAC than it was installed on - in case of TemplateVM and TemplateBasedVMs). But still the setting is there.

unman: @marmarek Thanks for that - I don't generally use Network Manager, so I haven't seen any effects like this. That said, I have it on one machine with NM and that just works fine - that is, it doesn't appear as if NM is running a DHCP client on the downlink. Let me dust that off and have a look.

unman: @marmarek So I've checked the position with NM and there don't seem to be any adverse effects. I've also looked at some of those basic Qubes boxes I set up and they have been running with Network Manager and this method for months now with no adverse effects.
NM can access the original address and so the issue you identified doesnt arise. (If you're wondering about this I see it's exactly the same as the Tails use of macchanger, so the NM issue isn't perceived to be a problem. See the refs linked to above.)

As to the setting for eth0 in VMs, I had looked at that, and hadn't realised that you could set the MAC - I have never been able to do so. When I try with `qvm-prefs -s MAC blah` I get an error "ERROR: wrong property name: 'MAC' "

unman: Ah ok, it needs `qvm-prefs -s mac blah`.

marmarek: > NM can access the original address and so the issue you identified doesnt arise. 

The question is *if* NM will use original address for `unmanaged-devices` setting. Make sure it will not mess with `vif*`. Is it still effective when you restart NM after that udev rule change the macchanger?

> (If you're wondering about this I see it's exactly the same as the Tails use of macchanger, so the NM issue isn't perceived to be a problem. See the refs linked to above.)

Tails doesn't have statically configured inter-VM interfaces...

> As to the setting for eth0 in VMs, I had looked at that, and hadn't realised that you could set the MAC - I have never been able to do so. When I try with qvm-prefs -s MAC blah I get an error "ERROR: wrong property name: 'MAC' "

`qvm-prefs -s ... mac`. Interesting why it is listed in uppercase.

unman: @marmarek Is there a way to tell internally if a device is real hardware? One of the nice things about this approach is that it applies if you have set up VM and then attach an new adapter to it, whereas with other method you have to provide the details in advance.

marmarek: > @marmarek Is there a way to tell internally if a device is real hardware?

I've tried to find some way but found only some workarounds. Generally take a look what udev shows about those devices:
```
$ udevadm info -q all /sys/class/net/eth0
P: /devices/vif-0/net/eth0
E: DEVPATH=/devices/vif-0/net/eth0
E: ID_MM_CANDIDATE=1
E: ID_NET_NAME_MAC=enx00163e5e6c17
E: ID_OUI_FROM_DATABASE=Xensource, Inc.
E: ID_PATH=xen-vif-0
E: ID_PATH_TAG=xen-vif-0
E: IFINDEX=2
E: INTERFACE=eth0
E: SUBSYSTEM=net
E: SYSTEMD_ALIAS=/sys/subsystem/net/devices/eth0
E: TAGS=:systemd:
E: USEC_INITIALIZED=2699433
```
```
$ udevadm info -q all /sys/class/net/vif2.0 
P: /devices/vif-2-0/net/vif2.0
E: DEVPATH=/devices/vif-2-0/net/vif2.0
E: ID_MM_CANDIDATE=1
E: ID_NET_DRIVER=vif
E: ID_NET_LINK_FILE=/usr/lib/systemd/network/99-default.link
E: ID_NET_NAME_MAC=enxfeffffffffff
E: IFINDEX=5
E: INTERFACE=vif2.0
E: SUBSYSTEM=net
E: SYSTEMD_ALIAS=/sys/subsystem/net/devices/vif2.0
E: TAGS=:systemd:
E: USEC_INITIALIZED=441230323
```

Maybe `ID_PATH!="xen-vif-*", ID_NET_DRIVER!="vif"` would be enough?

andrewdavidwong:
Just to check: Is this PR waiting on a response from @unman? Or is @unman waiting for us to merge it? (Just trying to avoid a misunderstanding where each party is waiting for the other.)

unman:
@andrewdavidwong , yes, it's waiting for me to find a half decent way of doing this that works across the board.

andrewdavidwong:
Understood. Thanks.

andrewdavidwong:
Since this has been on hold for almost 6 months, let's close this PR for now. We can reopen it later if there's any progress.

PR #171, opened 2016-07-17
Update usb.md

yilmi:
USB input devices are handled as input devices in a specific qubes-rpc configuration file.

If a user wants to enable USB proxy for other USB devices type, he must update /etc/qubes-rpc/policy/qubes.USB


marmarek:
This is automatically done by `qvm-usb` tool, no manual edit is needed.

yilmi:
I updated to 3.2 rc1 today and had some issues with USB proxy. After playing with a few settings I had it working, I though /etc/qubes-rpc/policy/qubes.USB was the cause.

Thanks for your answer

PR #172, opened 2016-07-19
Add a specific note about where to put iptables in /rw/config/

clayt0nk:
This would have probably saved me a fair bit of flailing around.

### DIFF
--- a/security/qubes-firewall.md
+++ b/security/qubes-firewall.md
@@ -86,7 +86,7 @@ sudo iptables -I FORWARD 2 -s <IP address of A> -d <IP address of B> -j ACCEPT
   issues from VM A. Note however, that this doesn't allow you to reach A from
   B -- for this you would need another rule, with A and B addresses swapped.
 * If everything works as expected, then the above iptables rule(s) should be
-  written into firewall VM's `qubes_firewall_user_script` script which is run
+  written into firewall VM's `qubes-firewall-user-script` script which is run
   on every firewall update. This is necessary, because Qubes orders every
   firewall VM to update all the rules whenever new VM is started in the system.
   If we didn't enter our rules into this "hook" script, then shortly our custom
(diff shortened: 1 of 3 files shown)


andrewdavidwong:
Might be worth explaining that all commands in `/rw/config/rc.local` will b executed as root whenever the VM is started.

@marmarek

clayt0nk:
It would in fact also be nice to explain what is special about qubes_firewall_user_script, to distinguish it from rc.local. I am guessing it gets re-run every time the Qubes firewall dynamically modifies it's firewall? Which process is why iptables settings in rc.local do not persist?

security/qubes-firewall.md
```diff
@@ -315,3 +315,11 @@ fi
...
 This time testing should allow connectivity to the service as long as the
 service is up :-)
+
+Where to put firewall rules
+---------------------------
+
+Implicit in the above example, but worth calling attention to: for all 
+VMs EXCEPT proxy VMs, iptables commands should be added to the 
+'/rw/config/rc.local' script. For proxy VMs, iptables commands should 
+be added to '/rw/config/qubes_firewall_user_script'.
```
marmarek: `/rw/config/qubes-firewall-user-script`
Also worth adding a link to https://www.qubes-os.org/doc/config-files/

clayt0nk:
Good suggestions all around, guys. I copied-n-pasted that script name, so there are multiple locations. I grepped the entire repo and hopefully got all of them. Hope I got that internal link right?

On a tangential note, from contributing.md:

	"Hard wrap Markdown lines at 80 characters."

I am resisting the urge to hard wrap every para. I touch to keep the diff clean, would you prefer I continue to do that or not?

andrewdavidwong:
> I am resisting the urge to hard wrap every para.

I know that urge well. :)

> I touch to keep the diff clean, would you prefer I continue to do that or not?

It probably won't matter much in this case, since diff tools should be able to distinguish content changes from rewrapping. However, it would certainly be better to rewrap unrelated lines in a different commit, and generally I find it cleaner to do that sort of thing in a separate PR. For such a small file, however, that would probably be overkill. (If you decided to rewrap a lot of text across multiple files, though, a separate PR would be great.)

clayt0nk:
Cool, I just made a hard-wrap-only commit of config-files.md, the other files I touched were actually in pretty good shape in this department.

PR #176, opened 2016-08-08
Add Dom0 prompt for root in Debian/Whonix VMs

crat0z:


andrewdavidwong:
@marmarek

crat0z:
In my naivety, and rush to get my first pull request done, I seem to have not tested thoroughly enough. This works by not allowing "su -" to be used to gain root privileges, so if you'd like to get root privilege you'd have to issue "sudo su". This seems to work in Debian and Whonix WS based VMs, but in the Whonix GW it will issue 3 dom0 prompts on first reboot of the machine. The dom0 prompts happen when (I believe) the user "user" attempts to execute sudo or su, or if any other user/group tries to use "su". I tried to analyze the outputs of "journalctl | grep pam", "journalctl | grep sudo" and "ps aux" to try and connect the dots as to which processes were requesting root privilege in that fashion, but I haven't been able to find a clear answer.

The work around is obviously to not apply the dom0 prompt for the Whonix Gateway, as (potentially) there could be processes on startup that require root privileges but aren't getting it in a proper way. For now that is what I've done, and I can write a note for it, but I'd like to eventually get dom0 prompt for root privilege in Whonix GW to work just as it does in my other VMs.

marmarek:
I wonder what else will stop working, looks like `sudo` is used a lot in Whonix:
```
user@host:~$ grep -r sudo /usr/lib/whonix*|wc -l
65
```
Some of those are false positives (like help messages), but still some are the real usage.

adrelanos:
Marek Marczykowski-Górecki:
> I wonder what else will stop working, looks like `sudo` is used a lot in Whonix:
> ```
> user@host:~$ grep -r sudo /usr/lib/whonix*|wc -l
> 65
> ```
> Some of those are false positives (like help messages), but still some are the real usage.

Probably not much would break because Non-Qubes-Whonix uses the more
Debian standard way of requiring a root password from the user by default.

crat0z:
Upon further investigation, ,I tried to open a terminal and then I pressed no to each prompt just to see what would happen. I looked at the journal again for sudo and PAM output, and it seems there were 4 instances of the user trying to run /usr/bin/virt-what with root privileges, and then one instance of "service whonixcheck status". The PAM authentication errors seem to match up, considering I closed 5 prompts on startup. Either way, I hope this helps.

If this can't be resolved immediately, that's OK. I was just curious as to whether or not I could get it the prompt to work like it does in Fedora so well. If there's a workaround I'd personally be interested.

adrelanos:
Latest version of whonixcheck (in stable, Whonix 13) does not use
virt-what anymore.

crat0z:
After banging my head against the wall for a while, I found the issue and a workaround. This is present in the current unstable release, i.e. the version described [here](https://www.whonix.org/blog/qubes-whonix-13-0-0-1-2-testers-wanted). In particular, the two files that (I believe) caused this issue are:
```
/etc/profile.d/20_power_savings_disable_in_vms.sh
/etc/X11/Xsession.d/20power_savings_disable_in_vms
```
Judging by the timestamps on these files, it seems they haven't been touched since 2013 or so. Anyway, in these scripts there are things like
```
result="$(sudo virt-what)"
```
Now of course, we have a sudoers.d file that says any user can run virt-what without password, however for some reason this might have not been applying correctly for some reason? I'm not sure why, but in /etc/sudoers.d/qubes I also added `user ALL=NOPASSWD: /usr/sbin/virt-what` and it seems to work now as it should.

I'm not 100% sure those power-saving scripts are the actual offenders of this, someone who is a better Linux ninja could look into this quickly, but the additional line in sudoers.d/qubes is what I'll use for now. Below will be some logs if you guys are interested. These logs were obtained after denying all requests for privilege, I found it made my search a lot easier.

```
root@host:/usr/lib# journalctl | grep "PAM authentication error"
Aug 09 23:20:49 host sudo[1057]: user : PAM authentication error: System error ; TTY=unknown ; PWD=/home/user ; USER=root ; COMMAND=/usr/sbin/virt-what
Aug 09 23:20:50 host sudo[982]: user : PAM authentication error: System error ; TTY=unknown ; PWD=/home/user ; USER=root ; COMMAND=/usr/sbin/virt-what
Aug 09 23:20:50 host sudo[936]: user : PAM authentication error: System error ; TTY=unknown ; PWD=/home/user ; USER=root ; COMMAND=/usr/sbin/virt-what
Aug 09 23:20:51 host sudo[1619]: user : PAM authentication error: System error ; TTY=unknown ; PWD=/home/user ; USER=root ; COMMAND=/usr/sbin/virt-what
Aug 09 23:20:55 host sudo[1680]: user : PAM authentication error: System error ; TTY=unknown ; PWD=/home/user ; USER=root ; COMMAND=/usr/sbin/service whonixcheck status

root@host:/usr/lib# journalctl | grep "virt-what"
Aug 09 23:09:51 host sudo[988]: root : TTY=unknown ; PWD=/root ; USER=root ; COMMAND=/usr/sbin/virt-what
Aug 09 23:09:54 host sudo[1100]: root : TTY=unknown ; PWD=/root ; USER=root ; COMMAND=/usr/sbin/virt-what
Aug 09 23:18:05 host sudo[1451]: root : TTY=unknown ; PWD=/root ; USER=root ; COMMAND=/usr/sbin/virt-what
Aug 09 23:20:49 host sudo[1057]: user : PAM authentication error: System error ; TTY=unknown ; PWD=/home/user ; USER=root ; COMMAND=/usr/sbin/virt-what
Aug 09 23:20:50 host sudo[982]: user : PAM authentication error: System error ; TTY=unknown ; PWD=/home/user ; USER=root ; COMMAND=/usr/sbin/virt-what
Aug 09 23:20:50 host sudo[936]: user : PAM authentication error: System error ; TTY=unknown ; PWD=/home/user ; USER=root ; COMMAND=/usr/sbin/virt-what
Aug 09 23:20:51 host sudo[1619]: user : PAM authentication error: System error ; TTY=unknown ; PWD=/home/user ; USER=root ; COMMAND=/usr/sbin/virt-what
Aug 09 23:24:04 host sudo[3060]: root : TTY=unknown ; PWD=/root ; USER=root ; COMMAND=/usr/sbin/virt-what
Aug 09 23:30:05 host sudo[3131]: root : TTY=unknown ; PWD=/root ; USER=root ; COMMAND=/usr/sbin/virt-what
Aug 09 23:36:07 host sudo[3206]: root : TTY=unknown ; PWD=/root ; USER=root ; COMMAND=/usr/sbin/virt-what
Aug 09 23:42:05 host sudo[3307]: root : TTY=unknown ; PWD=/root ; USER=root ; COMMAND=/usr/sbin/virt-what


root@host:/usr/lib# journalctl | grep "whonixcheck"
Aug 09 23:11:57 host mount-dirs.sh[350]: ++ binds+=('/var/lib/whonixcheck')
Aug 09 23:11:57 host mount-dirs.sh[350]: + '[' -h /var/lib/whonixcheck ']'
Aug 09 23:11:57 host mount-dirs.sh[350]: + true '/var/lib/whonixcheck is not a symlink'
Aug 09 23:11:57 host mount-dirs.sh[350]: + true 'fso_ro: /var/lib/whonixcheck'
Aug 09 23:11:57 host mount-dirs.sh[350]: + fso_rw=/rw/bind-dirs/var/lib/whonixcheck
Aug 09 23:11:57 host mount-dirs.sh[350]: + umount /var/lib/whonixcheck
Aug 09 23:11:57 host mount-dirs.sh[350]: + '[' -d /var/lib/whonixcheck ']'
Aug 09 23:11:57 host mount-dirs.sh[350]: + cp --verbose --no-clobber --archive --recursive --parents /var/lib/whonixcheck /rw/bind-dirs
Aug 09 23:11:57 host mount-dirs.sh[350]: + mount --bind /rw/bind-dirs/var/lib/whonixcheck /var/lib/whonixcheck
Aug 09 23:09:50 host systemd[1]: Starting whonixcheck...
Aug 09 23:09:50 host systemd[1]: Started whonixcheck.
Aug 09 23:20:52 host sudo[1761]: user : TTY=unknown ; PWD=/ ; USER=user ; COMMAND=/usr/bin/touch /var/run/whonixcheck/whonixcheck_running
Aug 09 23:20:52 host sudo[1764]: user : TTY=unknown ; PWD=/ ; USER=user ; COMMAND=/usr/bin/tee /var/run/whonixcheck/whonixcheck_pid
Aug 09 23:20:55 host sudo[1680]: user : PAM authentication error: System error ; TTY=unknown ; PWD=/home/user ; USER=root ; COMMAND=/usr/sbin/service whonixcheck status
Aug 09 23:20:57 host sudo[2534]: user : TTY=unknown ; PWD=/ ; USER=user ; COMMAND=/usr/bin/touch /var/run/whonixcheck/whonixcheck_recently_run

root@host:/usr/lib/qubes# systemctl status whonixcheck -l
● whonixcheck.service - whonixcheck
   Loaded: loaded (/lib/systemd/system/whonixcheck.service; enabled)
  Drop-In: /lib/systemd/system/whonixcheck.service.d
           └─40_qubes.conf
   Active: active (running) since Tue 2016-08-09 23:09:50 UTC; 1h 0min ago
     Docs: https://www.whonix.org/wiki/whonixcheck
 Main PID: 873 (whonixcheckdaem)
   CGroup: /system.slice/whonixcheck.service
           ├─ 873 /bin/bash /usr/lib/whonixcheckdaemon
           └─2950 sleep 3600

Aug 09 23:20:55 host sudo[2314]: pam_unix(sudo:session): session closed for user root
Aug 09 23:20:55 host sudo[2316]: user : TTY=unknown ; PWD=/ ; USER=root ; COMMAND=/sbin/ifconfig eth1
Aug 09 23:20:55 host sudo[2316]: pam_unix(sudo:session): session opened for user root by (uid=0)
Aug 09 23:20:55 host sudo[2316]: pam_unix(sudo:session): session closed for user root
Aug 09 23:20:55 host sudo[2321]: user : TTY=unknown ; PWD=/ ; USER=debian-tor ; COMMAND=/usr/sbin/tor --verify-config
Aug 09 23:20:55 host sudo[2321]: pam_unix(sudo:session): session opened for user debian-tor by (uid=0)
Aug 09 23:20:55 host sudo[2321]: pam_unix(sudo:session): session closed for user debian-tor
Aug 09 23:20:57 host sudo[2534]: user : TTY=unknown ; PWD=/ ; USER=user ; COMMAND=/usr/bin/touch /var/run/whonixcheck/whonixcheck_recently_run
Aug 09 23:20:57 host sudo[2534]: pam_unix(sudo:session): session opened for user user by (uid=0)
Aug 09 23:20:57 host sudo[2534]: pam_unix(sudo:session): session closed for user user
```

crat0z:
Sorry for double posting, but I forked the power-savings-disable-in-vms package for Whonix, switched every instance of virt-what to systemd-detect-virt and I installed it in a fresh TemplateVM for both the Gateway and Workstation. I was testing it out before hand and the logic seems to work OK with systemd-detect-virt instead of virt-what. If @adrelanos could check it out, I'd like to submit a pull request for it. You can see my commit and changes [here](https://github.com/crat0z/power-savings-disable-in-vms/commit/aebcfddeac1322308fbc4dfd0d4849b58a3e214b). Please note the  `etc/sudoers.d/power-savings-disable-in-vms` change, as it seems meaningless to have that line now. I can remove it if you wish though.

In the TemplateVMs this fixed the prompts at startup, however when I created an AppVM and ProxyVM, as expected, there was one prompt coming from whonixcheck. The faulty file seems to be `usr/lib/qubes-whonix/qubes-whonixsetup` in the `qubes-whonix` package. This script is referenced by `etc/xdg/autostart/qubes-whonixsetup.desktop` on startup I believe, so it's executed as the user and uses sudo to execute `service whonixcheck status` etc. I was digging around in that repository, and it seems there's a systemd unit "whonixcheck.service" that perhaps could be worked together with this script, that way we wouldn't have to use the `etc/xdg/..` script.

What do you guys think? For now I'll use my custom package and the `user ALL=NOPASSWD: /usr/sbin/service whonixcheck *` line in sudoers.d. I would try to stitch together the fix I described above with whonixcheck.service but I don't think I'm knowledgeable enough not to mess it up :D.

adrelanos:
There has not been spend a lot thought on the qubes-whonix package for proper /etc/sudoers.d permissions indeed, since Qubes by default allows them all. However, there are just three uses of sudo.

```
usr/lib/qubes-whonix/qubes-whonixsetup:        kdesudo whonix-setup-wizard quick
usr/lib/qubes-whonix/qubes-whonixsetup:        if sudo service whonixcheck status ; then
usr/lib/qubes-whonix/qubes-whonixsetup:            sudo service whonixcheck restart || true
```

Restarting whonixcheck should be covered by my above attached commit. Please emulate that commit and test.

Starting whonix-setup-wizard should already be covered by https://github.com/Whonix/whonix-setup-wizard/blob/master/etc/sudoers.d/whonix-setup-wizard.

I don't think I want to port power-savings-disable-in-vms to systemd-detect-virt (thereby making it systemd specific) for fixing an /etc/sudoers.d error.

Perhaps rather than `sudo service` the scripts would have to use hardcoded...?
- `sudo service`-> `sudo /usr/sbin/service`
- `sudo virt-what` -> `/usr/sbin/virt-what`

Also...

- `whonix-setup-wizard` -> `/usr/lib/whonix-setup-wizard/whonix-setup-wizard-setup`

Can you please test and send pull requests if applicable?

crat0z:
Yes, that fixed the one issue. The one with the virt-what still persists though, however, besides my suggestion of porting the power-savings-disable-in-vms package to systemd-detect-virt, I think it's just a simple misconfiguration of the /etc/sudoers.d folder that is causing this. It seems that due to my changes in /etc/sudoers.d/qubes, the policy in /etc/sudoers.d/power-savings... isn't being applied properly? I renamed the file from power-savings-disable-in-vms to whonix-power-savings-disable-in-vms and it seems to work without any changes in the power-savings-disable-in-vms package, and with the commit above.

Perhaps with your commit + simply renaming the power-savings-disable-in-vms file to whonix-something (or as long as the first letter comes after q apparently..) it should resolve all of the issues I've been facing with the prompt for Whonix and we can have one consistent guide for both Debian and Whonix based VMs.

adrelanos:
It's a generic, not Whonix specific package. So names with whonix- are
inappropriate. It's better to fix the sudoers bug.

andrewdavidwong:
@crat0z, would you like us to keep this PR open while you work on this?

crat0z:
Yes, my instructions work on Debian based VMs definitely. I think I have a solution for the last issue we have. With the changes @adrelanos made to the qubes-whonix package, the addition of the ``/etc/sudoers.d/qubes-whonix`` file fixes one problem.

The only other issue is with the power-savings-disable-in-vms package and it's sudoers file. I'm not sure if the sudoers.d parsing is really a bug, the [Arch wiki](https://wiki.archlinux.org/index.php?title=Sudo&diff=338053&oldid=334951#Configure_sudo_using_dropin_files_in_.2Fetc.2Fsudoers.d) has documented this phenomena for at least 2 years. Simply put, the changes necessary in the ``/etc/sudoers.d/qubes`` file override the ``/etc/sudoers.d/power-savings-disable-in-vms`` file because the former comes after the latter lexically.

I propose that with the new file ``/etc/sudoers.d/qubes-whonix`` in the qubes-whonix package, we append ``ALL ALL=NOPASSWD: /usr/sbin/virt-what *`` to the qubes-whonix file. Since this doesn't affect the power-savings-disable-in-vms package and is a qubes specific feature, I think it's a good idea. I made the necessary changes and ran it on fresh Whonix TemplateVMs and ProxyVM/AppVMs and it fixed the issue I was having.

adrelanos:
https://github.com/Whonix/qubes-whonix/pull/3 is not good. qubes-whonix package does not use virt-what so that fix does not belong there. Only the power-savings-disable-in-vms package uses virt-what in Whonix.

Can you try with the following commit please?

https://github.com/Whonix/power-savings-disable-in-vms/commit/68d55c3d318105d7dc0961a9b92b67630f0e62e0

(Emulate these changes or `sudo make install` the package from git.)

Otherwise if that still does not work we can modify. [`/etc/sudoers.d/power-savings-disable-in-vms`](https://github.com/Whonix/power-savings-disable-in-vms/blob/master/etc/sudoers.d/power-savings-disable-in-vms)

From:

```
ALL ALL=NOPASSWD: /usr/sbin/virt-what
```

To:

```
ALL ALL=NOPASSWD: /usr/sbin/virt-what *
```

But I don't think the latter would be needed.

adrelanos:
I didn't notice your previous comment https://github.com/QubesOS/qubes-doc/pull/176#issuecomment-242583850.

adrelanos:
I don't think there is a bug in sudo. Whatever the order of sudoers.d snippets being parsed, they should all be properly parsed and regarded. Can you please try the above 1 and if necessary two things?

crat0z:
The modified power-savings-disable-in-vms package made the VMs unusable unfortunately. I tried it twice, one with manual changes and once again with the install method. It gives an error ``Error starting VM: cannot start qubes-guid!``. The guid logs say ``Failed to connect to gui-agent`` and the xen/console/guest-[vm-name].log file seems to boot normally like my other current Whonix VMs. I still have the VMs so if there are any other commands I can pass from dom0 I can do that for diagnostic purposes.

However, I agree with what you said about the sudo problem. I suppose the way the sudoers.d folder is parsed is just one of it's quirks. I still think the fixes I suggested to the ``/etc/sudoers.d/qubes-whonix`` file is appropriate, considering the qubes-whonix package is specifically for configuration options in Qubes. The redundant sudoers.d lines I proposed are a little 'hacky', but I personally can't come up with any other way of handling this issue. Also, I can't imagine there are any Qubes + Whonix users who don't have the power-savings-disable-in-vms package installed anyway.

I'm still open to different ideas and I can continue testing your changes. Thanks for the help in getting this issue handled!

adrelanos:
There was a bug. Now fixed.

https://github.com/Whonix/power-savings-disable-in-vms/commit/02195ed5771ece4f3f476bfb2476f76bc9c4f090

Please try again.

crat0z:
Yes, that fixed the issue. With the qubes-whonix and power-savings-disable-in-vms packages installed from GitHub, everything works beautifully. Thanks for the help.

andrewdavidwong:
> Yes, that fixed the issue. With the qubes-whonix and power-savings-disable-in-vms packages installed from GitHub, everything works beautifully. Thanks for the help.

Does this mean that this PR should be closed, or is some action still desired here?

crat0z:
If the team thinks my submission is up to the standards for documentation, I can confirm the instructions I submitted work for all of the templateVMs I use. The only issue is that currently the changes that were worked out above aren't in the main Whonix systems, but once those changes make their way into everyone's VMs my instructions will work.

andrewdavidwong:
> If the team thinks my submission is up to the standards for documentation, I can confirm the instructions I submitted work for all of the templateVMs I use. The only issue is that currently the changes that were worked out above aren't in the main Whonix systems, but once those changes make their way into everyone's VMs my instructions will work.

@adrelanos, what do you think?

adrelanos:
Not so much technical, but project philosophy related.

I don't know what to make of this. The Qubes Council (I think?) decision has been made to make Qubes work sudo passwordless. It has been opposed on the mailing list some time ago and rejected. And I am fine with that.

So I don't think I should be investing too much time into a geek feature: Qubes asking for sudo passwords. (Time for review, documentation, research, bug fixing, discussion, testing.) Therefore, I am not reviewing these documentation changes, that would require me testing all of this. I had a quick look, seems non-malicious. It's interesting but not something great to look into in detail for said reasons.

(I consider the bug that @crat0z reported, minor, low effort to fix for me and fixed. Will appear in Whonix 14.)

Whether any resources should be spend on this is the old question on how to deal with community contributions for features that are not on the Qubes roadmap. Usability wise in this case it probably does not matter so much. That vm-sudo.md page is probably only viewed by auditors, curious people and advanced users so the extra length does not mess anything up.

So there are two strategies here:

* a) not try to make everyone happy - restore the original vm-sudo.md page as written by Joanna and reject any requests to document using Qubes with sudo to save resources
* b) be friendly by trying to serve the advanced users also

andrewdavidwong:
@adrelanos, thank you for sharing your thoughts. I'll merge @crat0z's contribution and add a suitable disclaimer (similar to the existing one on that page).

andrewdavidwong:
Thanks, @crat0z!

tasket:
@adrelanos As I mentioned on Whonix forum, the Qubes philosophy is stated in the sudoers.d/qubes file and it cites that only *one* Xen priv escalation bug was know to date. But since then, many such bugs have been found. So if it isn't very much trouble to keep a guest OS's original security intact, I think Qubes should make it possible for the user to enable for the sake of mitigation (another early Qubes maxim was "Security is not a boolean" IIRC, which can be applied to this issue).

@crat0z @adrelanos Thanks for looking into this issue. My own path has led me to enable both sudo prompt and AppArmor in my templates. However, I was not able to follow this thread 100%... It seemed to me the most direct course was simply this:

cp -a /etc/sudoers.d/power-savings-disable-in-vms /etc/sudoers.d/zz99

This makes the rule for virt-what come last, and it does prevent sudo prompts from appearing on VM startup.

adrelanos:
tasket:
> @adrelanos As I mentioned on Whonix forum, the Qubes philosophy is stated in the sudoers.d/qubes file and it cites that only *one* Xen priv escalation bug was know to date. But since then, many such bugs have been found. So if it isn't very much trouble to keep a guest OS's original security intact, I think Qubes should make it possible for the user to enable for the sake of mitigation (another early Qubes maxim was "Security is not a boolean" IIRC, which can be applied to this issue).

I don't know if this is the right place at Qubes to discuss this. I
guess @marmarek can tell if this discussion should be reopened in form
of a qubes-dev mailing list thread or new qubes-issues ticket or if this
discussion remains closed.

andrewdavidwong:
@tasket @adrelanos:

I wouldn't say it's *wrong* to discuss this here, but @tasket, if you feel strongly about this, you should probably create an issue for it (or I can do it for you). Then, the Qubes team can decide whether this is something we're willing to implement.

marmarek:
Discussions about architecture should take place on qubes-devel ML.

PR #178, opened 2016-08-15
Fix broken link

rasnauf:
The URL https://www.networkworld.com/news/2007/080207-black-hat-virtual-machine-rootkit-detection.html is a broken link -- presenting an error "Page not found" upon request. However, there exists an archive of this article on archive.org, and I have updated the URL to point to the archive.


andrewdavidwong:
Thanks!

PR #186, opened 2016-09-15
Update vpn.md

tasket:
* Scripts and text mention openvpn only in the context of examples.
* Firewall commands slightly tweaked: Important blocking rules move to top. Removed superfluous check for qvpn OUTPUT rule.
* Clarifications, especially mentioning that NM shouldn't be enabled for iptables/scripts option (this was causing DNS to fail for people who tried both NM and scripts).
* Tells the user when and what they should test (iptables/scripts).
* Change script order to enable testing flow.
* Added Usage and Troubleshooting sections.

https://github.com/QubesOS/qubes-issues/issues/2317

### DIFF
--- a/configuration/vpn.md
+++ b/configuration/vpn.md
@@ -66,7 +66,7 @@ This method is more involved than the one above, but has anti-leak features that
 
     ![Create\_New\_VM.png](/attachment/wiki/VPN/Create_New_VM.png)
     
-    Note: Do not enable NetworkManager, as it can interfere with the scripts' DNS features. If you used NetworkManager or other method in a previous attempt, do not re-use the old ProxyVM... Create a new one according to this step.
+    Note: Do not enable NetworkManager in the ProxyVM, as it can interfere with the scripts' DNS features. If you enabled NetworkManager or used other methods in a previous attempt, do not re-use the old ProxyVM... Create a new one according to this step.
     
     If your choice of template VM doesn't already have the VPN client software, you'll need to install the software in the template before proceeding. Disable any auto-starting service that comes with the software package: for example `sudo systemctl disable openvpn.service`.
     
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
@marmarek

configuration/vpn.md
```diff
@@ -60,67 +56,40 @@ Using a ProxyVM to set up a VPN client gives you the ability to:
...
-    sleep 2s
-
-	#    Allow traffic from the `qvpn` group to the uplink interface (eth0);
-	#    Our openvpn will run as group `qvpn`.
-    iptables -A OUTPUT -p all -o eth0 -m owner --gid-owner qvpn \
-    -m state --state NEW,ESTABLISHED -j ACCEPT
-
-	#	Allow internal system connections:
-    iptables -I OUTPUT -o lo -j ACCEPT
+    Note: Do not enable NetworkManager, as it can interfere with the scripts' DNS features. If you used NetworkManager or other method in a previous attempt, do not re-use the old ProxyVM... Create a new one according to this step.
```
marmarek: Since enabling NetworkManager in ProxyVM is not obvious option, this can be confused with using NetworkManager in NetVM. Which is obviously not what this note is about.

ttasket: Then I should edit and mention 'proxyVM' at the beginning of the note.

ttasket: Also, I submitted another PR to the doc in parallel, to fix the paths in step 3. I can incorporate that change along with the note change, into an edit of 'patch-4' if you prefer.

marmarek: I think you submitted it to wrong repository. Anyway adding it here is also ok.

ttasket: Done.

PR #187, opened 2016-09-15
Add nouveau disable instructions

kewde:
Work in progress - do not merge just yet.

### DIFF
--- a/troubleshooting/nvidia-troubleshooting.md
+++ b/troubleshooting/nvidia-troubleshooting.md
@@ -13,9 +13,16 @@ NVidia Troubleshooting Guide
 
 If you have an NVidia graphics card it will probably not work under Xen out of the box. If your system freezes during boot and you don't see the graphical login manager after you installed Xen, then this problem most likely affects you. The following steps should provide a work around so that you should be able to use your NVidia with X under Xen, however without any fancy "desktop effects".
 
+Boot in failsafe
+---------------------
+
 1.  Boot your system using the "failsafe" boot menu, that should have been automatically added to your `grub.conf` when you installed the Dom0 kernel.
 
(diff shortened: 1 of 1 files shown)


kewde:
@andrewdavidwong 
The numbering of both section is off, it always resets to 1 which is a bit awkward and already present in the existing docs.

kewde:
Okay to merge 👍 

A follow up pull request on this page should deal with explaining more:
What is X.org and X Window System
What is nouveau
Alternative to nouveau: closed source nvidia drivers

kewde:
Also isn't booting in failsafe mode obsolete as it's done automatically?

andrewdavidwong:
@marmarek

troubleshooting/nvidia-troubleshooting.md
```diff
@@ -71,3 +78,70 @@ If you see a terminal window in the top left corner, it means you most likely su
 
 1.  Reboot and let the system boot from the normal boot configuration. You should be able to use X under Xen now.
 
+
+Disabling Nouveau
+---------------------
```
kewde: This is a bit of a paradoxical situation I believe as they wouldn't see an error without a built-in graphics card when nouveau decides to crash.

I'll add it if you believe I should.

marmarek: Ah, you're right.

kewde: @marmarek 
 I've added a section giving a fair mention that this disables nvidia support all together.

The nice thing about this patch it that editing the GRUB boot config at boot is forgiving, so if you disable nouveau all together and weird stuff happens then on reboot it will boot with nouveau enabled again.

PR #198, opened 2016-09-30
Update upgrade-to-r3.2.md

redfish64:
dnf isn't available on debian systems, switched to apt-get

### DIFF
--- a/installing/upgrade/upgrade-to-r3.2.md
+++ b/installing/upgrade/upgrade-to-r3.2.md
@@ -151,7 +151,7 @@ repeated in **all** the user's Template and Standalone VMs.
 
 4.  Add new packages (only needed in default template):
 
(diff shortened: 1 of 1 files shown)


installing/upgrade/upgrade-to-r3.2.md
```diff
@@ -151,7 +151,7 @@ repeated in **all** the user's Template and Standalone VMs.
 
 4.  Add new packages (only needed in default template):
 
-        sudo dnf install qubes-mgmt-salt-vm-connector
+        sudo apt-get install qubes-mgmt-salt-vm-connector
```
marmarek: Actually `apt-get` should be added, not replaced, as it isn't available on Fedora template. It depends on what is default template.

marmarek: Ah, ok. Sorry for the noise.

PR #199, opened 2016-10-11
Make reference to using bind-dirs with cron

unman:
Make it clear that for persistent crontabs bind-dirs should be used. QubesOS/qubes-issues#2326 and QubesOS/qubes-issues#2327 refer

### DIFF
--- a/reference/dom0-tools/qvm-service.md
+++ b/reference/dom0-tools/qvm-service.md
@@ -88,7 +88,7 @@ Enable CUPS service. The user can disable cups in VM which do not need printing 
 crond  
 Default: disabled
 
-Enable CRON service.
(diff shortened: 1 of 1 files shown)


marmarek:
I think it would be better to package bind-dirs configuration for cron, as part of https://github.com/QubesOS/qubes-core-agent-linux/pull/19, instead of relying on user to manually do it.

unman:
Ok, that's packaged now and I've amended doc accordingly 
QubesOS/qubes-issues#2326 and QubesOS/qubes-issues#2327 refer

andrewdavidwong:
Changes merged.

But won't this file get overwritten when it's synced from the other dom0-tools repo, @marmarek?

PR #200, opened 2016-10-13
Add section for Network Manager on Debian 9

tasket:
Added section to incorporate new NM features tracked in issue #938 
https://github.com/QubesOS/qubes-issues/issues/938

### DIFF
--- a/privacy/anonymizing-your-mac-address.md
+++ b/privacy/anonymizing-your-mac-address.md
@@ -31,7 +31,7 @@ Add the settings in /etc/NetworkManager/NetworkManager.conf. The following examp
 wifi.scan-rand-mac-address=yes
 
 [connection]
-wifi.assigned-mac-address=stable
+wifi.cloned-mac-address=stable
 ```
 
 To see the available configuration options, refer to the man page: `man nm-settings`
--- a/privacy/anonymizing-your-mac-address.md
+++ b/privacy/anonymizing-your-mac-address.md
@@ -9,10 +9,41 @@ redirect_from:
 Anonymizing your MAC Address
 ============================
 
-Changing the default [MAC Address](https://en.wikipedia.org/wiki/MAC_address) of your hardware is [crucial in protecting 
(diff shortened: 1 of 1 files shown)


privacy/anonymizing-your-mac-address.md
```diff
@@ -9,10 +9,41 @@ redirect_from:
...
+[device-scan]
+wifi.scan-rand-mac-address=yes
+
+[connection]
+wifi.assigned-mac-address=stable
+```
+
+To see the available configuration options, refer to the man page: `man nm-settings`
+
+Next, create a new NetVM using the new template and assign network devices to it.
```
marmarek: Is new NetVM really necessary? I think changing sys-net template is much easier.

tasket: It seems like the simplest way to get the user to start with fresh connection (AP specific) settings. Carrying over the old settings from 1.2.x prevented the global settings from taking effect, probably because the old version filled in some specific values relating to the mac address.

(There is a small security bonus in getting rid of the old private.img as well, if it was infected.)

Is there a better way?

marmarek: Ok, if there is a reason for this, that's fine.

tasket: There is problem with this edit of the doc... The nm-settings man page advises that `wifi.cloned-mac-address` is deprecated in favor of `wifi.assigned-mac-address`, however the latter does not yet function while the former does. It should be changed to `wifi.cloned-mac-address` for now.

tasket: The two sections "Usage Notes" and "Uninstalling.." should be re-titled to show they belong to macchanger, not NM instructions.

tasket: FYI, I checked on the NM list and wifi.cloned-mac-address is the right property for NetworkManager.conf. Its only deprecated on the DBUS interface, which doesn't affect the conf file.

https://mail.gnome.org/archives/networkmanager-list/2016-October/msg00016.html

PR #201, opened 2016-10-13
Cover new reinstall feature in qubes-dom0-update

tasket:
New support for reinstalling templates in-place was added in Qubes 3.1.

### DIFF
--- a/managing-os/reinstall-template.md
+++ b/managing-os/reinstall-template.md
@@ -11,8 +11,23 @@ How to Reinstall a TemplateVM
 
 If you suspect your [TemplateVM] is broken, misconfigured, or compromised,
 or if you wish to do a clean reinstall in order to upgrade to a new version, you
-can reinstall any TemplateVM from the Qubes repository. In what follows, the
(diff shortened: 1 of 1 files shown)


tasket:
Issue https://github.com/QubesOS/qubes-issues/issues/2061

tasket:
One issue about the current wording:

The Manual section states that a manual reinstall should be done "...if you wish to do a clean reinstallation of a template in order to upgrade to a different version".

My understanding of the new/automatic process is that it provides a clean reinstallation, to the point of removing the old private.img, so it should also be able to do the job for reverting, upgrading, etc.

tasket:
One issue about the current wording:

The Manual section states that a manual reinstall should be done "...if you wish to do a clean reinstallation of a template in order to upgrade to a different version".

My understanding of the new/automatic process is that it provides a clean reinstallation, to the point of removing the old private.img, so it should also be able to do the job for reverting, upgrading, etc.

andrewdavidwong:
@tasket: Updated the wording [here](https://github.com/QubesOS/qubes-doc/commit/65fc5bc0c58887333dd08137e5915fb9354e7237). Thanks!

PR #204, opened 2016-10-20
Remove outdated no-PVUSB claim

jpouellet:
There is an entire section below on how to do single-device passthrough since R3.2.


jpouellet:
I think the sections should probably be re-ordered, as it currently misleads a reader as to what is currently possible vs. not on a first read-through.

andrewdavidwong:
> I think the sections should probably be re-ordered, as it currently misleads a reader as to what is currently possible vs. not on a first read-through.

I'm not sure what you have in mind. Which order do you think the sections should be in?

PR #205, opened 2016-10-20
removed gendered language

mfc:


andrewdavidwong:
I'm open to almost any solution to gendered language *except* sacrificing subject-verb agreement.

As you've probably noticed, I typically default to feminine, rather than masculine, pronouns. The only times I don't are when using feminine pronouns would reinforce sexual stereotypes or, as in the present case, when the referent of the pronoun is being portrayed as adversarial, bad, or evil. For the most part, I aim to follow [the American Philosophical Association's Guidelines for Non-Sexist Use of Language](http://www.apaonlinecsw.org/apa-guidelines-for-non-sexist-use-of-language).

andrewdavidwong:
What do you think of this rewrite that avoids pronouns entirely?

mfc:
That APA resource seems a bit old-school (1986), in the sense that it avoids suggesting they/their as a gender-neutral singular pronoun. I don't know if that's what you're referring to re: subject-verb agreement, but I think it's considered a-ok nowadays.

> What do you think of this rewrite that avoids pronouns entirely?

yep looks good!

andrewdavidwong:
It's certainly OK in casual conversation, but not in formal writing. The docs fall somewhere in between.

PR #211, opened 2016-11-06
Reworked the page structure, added USBVM use case

sprig-florist:
Expanded the description of a standard NetVM with qubes-input-proxy-sender, so that a 3.2 standard installation with NetVM=UsbVM does not leave USB broken when replacing the NetVM template with the fedora-minimal one.

### DIFF
--- a/managing-os/templates/fedora-minimal.md
+++ b/managing-os/templates/fedora-minimal.md
@@ -49,10 +49,10 @@ To access the journald log, use the `journalctl` command.
 
 ### as a NetVM
 
-If you want to use this template to for standard NetVMs with USB forwarding capabilities you should install some more packeges:
+If you want to use this template for standard NetVMs you should install some more packages:
 
 ~~~
-[user@F21-Minimal ~]$ sudo dnf install NetworkManager NetworkManager-wifi network-manager-applet  wireless-tools dbus-x11 dejavu-sans-fonts tinyproxy qubes-input-proxy-sender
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
This is good for people who have "NetVM=USBVM", but it's unnecessary for people who have separate VMs for those two. Instead of encouraging everyone to install the extra package (including people for whom it's unnecessary), why not instead encourage people to install it only if they need it?

sprig-florist:
Better? Also, would it make sense to create a more nicely structured table?
Something like this:

Purpose | Description | Packages
--- | --- | ---
**Firewall** | No extra packages are needed for the template to work as a firewall | 
**NetVM** | These packages are needed for the template to work as a NetVM | NetworkManager NetworkManager-wifi network-manager-applet  wireless-tools dbus-x11 dejavu-sans-fonts tinyproxy
**USBVM** | If you want this template to allow USB input forwarding, you will need to install one extra package. | qubes-input-proxy-sender

andrewdavidwong:
Yes, much better! And that table would be even better!

sprig-florist:
@andrewdavidwong here it is. 

The structure is almost completely new, I have reworked most of the content, see commit comments.
Feedback appreciated, if changes are needed!

andrewdavidwong:
Looks fantastic; thanks for putting in the work on this! I just added a minor cleanup commit on top of yours and merged.

PR #215, opened 2016-11-12
Re-working the template introduction

sprig-florist:
First changes to re-work the introduction of the fedora-minimal template


sprig-florist:
Wrong pull request, incorporating into the main one (#211)

PR #220, opened 2016-11-16
Fix quoting problem

tasket:
Single quotes should be used for the dom0 commands.

Discussed here https://groups.google.com/d/msgid/qubes-users/33faf03d-b963-7136-2b73-6badc75f9efb%40openmailbox.org

### DIFF
--- a/security/vm-sudo.md
+++ b/security/vm-sudo.md
@@ -104,10 +104,10 @@ this for extra security.**
 
 1. Adding Dom0 "VMAuth" service:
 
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
It looks like @unman beat you to it with his [earlier PR to fix this issue](https://github.com/QubesOS/qubes-doc/pull/219) (by backslash escaping the `$anyvm` variable). Both your and his solutions work, but they don't work simultaneously. If one is objectively better for some reason, let me know.

PR #221, opened 2016-11-16
Tweak NetworkManager-based MAC randomization

rustybird:


rustybird:
Hmm, now I wonder if the second commit is misleading, because `ethernet.cloned-mac-address=stable` applies to NetworkManager connection *profiles*, which could each be used to connect to many different Ethernet networks. Would `ethernet.cloned-mac-address=random` be a better example to include?

rustybird:
Force pushed reworded commit that referes to "connection profiles" instead of "connections"

andrewdavidwong:
> Hmm, now I wonder if the second commit is misleading, because `ethernet.cloned-mac-address=stable` applies to NetworkManager connection profiles, which could each be used to connect to many different Ethernet networks. Would `ethernet.cloned-mac-address=random` be a better example to include?

I think this is ok, if for no other reason than that it's consistent with `wifi.cloned-mac-address=stable`, but if @tasket or @marmarek have any objections, we can change it.

tasket:
The doc states that it affects the Wifi address, but the edit expands that to ethernet. May want to change the wording in the paragraph directly above to reflect that.

Also, although the doc refers the reader to the man page, it could also use a quick explanation that 'stable=random that persists per boot session' and 'random=random each time a link goes up'.

andrewdavidwong:
Thanks, @tasket. [Updated.](https://github.com/QubesOS/qubes-doc/commit/4eb23c0e09511dd4c3ca6805874ba499bd052db8)

PR #222, opened 2016-11-18
Update archlinux template building instructions

tezeb:
I've rebuild archlinux so few minor updates to documentation(new screenshots added to qubes-attachment).

Also @marmarek, is it still true that your qubes-builder repo(https://github.com/marmarek/qubes-builder.git) is significantly newer/better/shinier then the official one(https://github.com/QubesOS/qubes-builder.git). Is it the still recommend to use yours for template building?

### DIFF
--- a/managing-os/templates/archlinux.md
+++ b/managing-os/templates/archlinux.md
@@ -23,7 +23,7 @@ Main maintainer of this template is [Olivier Médoc](mailto:o_medoc@yahoo.fr).
 ## Instructions ##
 
 <br>
-**These are the instructions for Qubes 3.1. They will take you step by step through the entire process start to finish**
+**These are the instructions for Qubes 3.2. They will take you step by step through the entire process start to finish**
 
 *Note: Currently there are no binary packages and it must be compiled from source using the instructions below.*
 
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
@marmarek

marmarek:
> Also @marmarek, is it still true that your qubes-builder repo(https://github.com/marmarek/qubes-builder.git) is significantly newer/better/shinier then the official one(https://github.com/QubesOS/qubes-builder.git). Is it the still recommend to use yours for template building?

That was recommended some time ago, as builder-archlinux wasn't created on official QubesOS account. But now it is there too. And the official one should be used. As for the freshness - I try to keep "master" branch in sync.

marmarek:
Also, stable release branches ("release3.1", "release3.2" etc) are available only on QubesOS account.

managing-os/templates/archlinux.md
```diff
@@ -213,11 +216,19 @@ Main maintainer of this template is [Olivier Médoc](mailto:o_medoc@yahoo.fr).
 
         *   Screen '**Get sources**' wants to download additional packages needed for the choosen plugin/s.
         
-            *   Select '**No**' Press '**Enter**'
```
tezeb: Yes. I myself changed it to "No" some time age, when not all sources were downloadable this way and this step was failing. Now it works as expected.

PR #223, opened 2016-11-19
qubes-mgmt-salt-admin-tools aint availble for 3.1

h01ger:
I recently did an upgrade from 3.0 to 3.1 and noticed that qubes-mgmt-salt-admin-tools wasn't available in 3.1 & the upgrade still went smoothly…

### DIFF
--- a/installing/upgrade/upgrade-to-r3.1.md
+++ b/installing/upgrade/upgrade-to-r3.1.md
@@ -90,10 +90,6 @@ complete.
     should be `3.1.4` or higher. If it's not, repeat the previous step with the
     `--clean` option.
 
-4.  Install new packages:
-
-        sudo qubes-dom0-update qubes-mgmt-salt-admin-tools
-
 4.  Reboot dom0.
     
     The system may hang during the reboot. If that happens, do not panic. All


andrewdavidwong:
@marmarek wrote:
> That's interesting, this step **should** be in 3.1->3.2 instruction...

("This step" refers to the step of installing `qubes-mgmt-salt-admin-tools`.)

Is this a typo? Did you mean "should **not**"?

If it is not a typo, and this step should be present in the docs, then why are we trying to remove it?

h01ger:
Marek said the step should be in the 3.1->3.2 instructions, and indeed it is there, though "hidden" with other packages… (rgrep salt-admin qubes-doc.git)

My pull request removes this step from the 3.0->3.1 instructions…

btw, where did marek wrote that?

andrewdavidwong:
> My pull request removes this step from the 3.0->3.1 instructions…

Ah, I see now. I mistakenly thought this was a PR against the 3.1->3.2 instructions.

Still, saying that this step **is** supposed to be present in the 3.1->3.2 instructions does not entail that it is **not** supposed to be present in the 3.0->3.1 instructions.

@marmarek specifically added it in this commit: https://github.com/QubesOS/qubes-doc/commit/e99688b4052ef27e836d67bdd7c8cfc3c1067feb

> btw, where did marek wrote that?

See the "wrote" link in my comment above.

marmarek:
Yes, this step should be in 3.1->3.2, **instead of** 3.0->3.1.

h01ger:
My pull request is correct, as are the 3.1->3.2 instructions.

andrewdavidwong:
Ok, @marmarek. Thank you for confirming.

@h01ger: Yes. Just to be clear, I never claimed that your PR is not correct. I needed conformation that it was correct before merging it, due to the surrounding confusion. I'm sure you will agree that demanding clarity is a better approach than blindly proceeding into confusion.

h01ger:
Thanks, Andrew!

PR #228, opened 2016-11-22
Additional step for Whonix

tasket:
Per this discussion, https://forums.whonix.org/t/fixing-whonix-boot-issue-after-securing-qubes-root-auth/3155/8

Whonix executes sudo commands in non-root startup scripts which causes pop-up auth prompts to appear while Whonix VMs are starting. The problem is partly due to sudo parsing sudoers.d entries in alphabetical order, and some later configs cause earlier ones to get overridden. Adding the right permissions to a lexically 'last' filename resolves the issue.

### DIFF
--- a/security/vm-sudo.md
+++ b/security/vm-sudo.md
@@ -141,11 +141,14 @@ this for extra security.**
 
           auth sufficient pam_permit.so
 
-    - For Whonix, if prompts appear during boot, create /etc/sudoers.d/zz99 and add two lines:
+    - For Whonix, if prompts appear during boot, create /etc/sudoers.d/zz99 and add these lines:
 
     ```
           ALL ALL=NOPASSWD: /usr/sbin/virt-what
-          ALL ALL=NOPASSWD: /usr/sbin/service whonixcheck *
+          ALL ALL=NOPASSWD: /usr/sbin/service whonixcheck restart
+          ALL ALL=NOPASSWD: /usr/sbin/service whonixcheck start
+          ALL ALL=NOPASSWD: /usr/sbin/service whonixcheck stop
+          ALL ALL=NOPASSWD: /usr/sbin/service whonixcheck status
     ```
 
 Dom0 password-less root access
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
@marmarek, @adrelanos: Any thoughts on this?

adrelanos:
Andrew David Wong:
> @adrelanos: Any thoughts on this?

Same as here:

https://github.com/QubesOS/qubes-doc/pull/176#issuecomment-242894132

tasket:
I'll try to write something meaningful about this in qubes-devel tonight. However, I'd ask you to look at such changes more in the spirit of error-correction than "contributions" as its not really code.

marmarek:
If one want to _not_ have passwordless sudo, `/usr/sbin/service whonixcheck *` can be risky. `service` script (and scripts called by it) may not be good at validating arguments. Better explicitly list allowed actions. Also, "funny" things may happen using environment variables (`sudo -E`) - like `SYSTEMD_PAGER`.
What to do about it? I don't know and don't want to spend too much time on it (why: read comment in default sudo configuration on Qubes). Maybe just add some warning.

tasket:
Since Patrick is curator of whonixcheck, I'd defer to his judgment on this. Personally I would prefer to enhance the sudoers pattern to include actions.

The doc already states that risk is an issue, no to rely on it strongly. The alternative is letting attackers having root unopposed, so I think our standard for 'due diligence' is to tighten accuracy to the extent that sudo normally allows.

adrelanos:
start, restart, status should be sufficient.

tasket:
Sudo globbing does not appear to support 'Or' lists, so change file to:
```
/usr/sbin/service whonixcheck start
/usr/sbin/service whonixcheck stop
/usr/sbin/service whonixcheck status
```

I'll try this out...

tasket:
I found a reference that uses 'restart', so I'll add that too.

tasket:
Seems to work OK.

tasket:
@andrewdavidwong : The new lines got all crammed together.

andrewdavidwong:
@tasket: Fixed.

PR #229, opened 2016-11-24
Clarifies R2 vs R3 language

lorentrogers:
I found the current language confusing, so I'm proposing this edit, which should help clarify. I believe that my edits here are accurate, but please correct if not.


andrewdavidwong:
Since R2 is no longer supported, we can remove that portion and focus on functionality for R3.x.

PR #234, opened 2016-12-01
Fixed a couple grammar / syntax issues

lorentrogers:
The "s" in the first section was out of place, so I put it on the VM instead of the work.

Also, I'm really not sure what the standard way to refer to a VM that's used as a template. Is this a "template VM" or a "TemplateVM"? I modified a few that I found, but I realized half way through that there was inconsistency all over the place. ("Standalone VM", "AppVM", etc.) Is there a standard way of writing this?

### DIFF
--- a/common-tasks/software-update-vm.md
+++ b/common-tasks/software-update-vm.md
@@ -11,20 +11,20 @@ redirect_from:
 Installing and updating software in VMs
 =======================================
 
-How Template VM works in Qubes
+How TemplateVMs work in Qubes
 ------------------------------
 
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
> Also, I'm really not sure what the standard way to refer to a VM that's used as a template. Is this a "template VM" or a "TemplateVM"? I modified a few that I found, but I realized half way through that there was inconsistency all over the place. ("Standalone VM", "AppVM", etc.) Is there a standard way of writing this?

Please see the glossary: https://www.qubes-os.org/doc/glossary/

lorentrogers:
Thanks for the info - I didn't notice that before. I'll tidy up the other missing contractions at some point.

andrewdavidwong:
Thanks, @lorentrogers!

PR #236, opened 2016-12-05
StandaloneVM term with parentheses

adrelanos:
Not common and as far I understand not useful to write Standalone(VM) or just Standalone. The canonical way to write that probably is StandaloneVM.

Also removed superfluous `(Virtual Machine)` since the term VM is explained above and linked as `[VM](#vm)`already for better consistency. (Otherwise, also for AppVM we'd have to add `(Virtual Machine)` to the description.)

### DIFF
--- a/reference/glossary.md
+++ b/reference/glossary.md
@@ -85,9 +85,9 @@ TemplateBasedVM
 Any [VM](#vm) which depends on a [TemplateVM](#templatevm) for its root
 filesystem.
 
-Standalone(VM)
+StandaloneVM
 --------------
-Standalone (Virtual Machine). In general terms, a [VM](#vm) is described as
+In general terms, a [VM](#vm) is described as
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
> Not common and as far I understand not useful to write Standalone(VM) or just Standalone.

The parentheses indicate that there are two related terms: "Standalone" and "StandaloneVM." (This is a standard way of writing such terms in a glossary.)

> The canonical way to write that probably is StandaloneVM.

Indeed, and this is explained in the entry.

> Also removed superfluous (Virtual Machine) since the term VM is explained above and linked as [VM](#vm)already for better consistency. (Otherwise, also for AppVM we'd have to add (Virtual Machine) to the description.)

In fact, we already have "Virtual Machine" in the entries for AppVM, NetVM, ProxyVM, etc. This is the standard way of defining terms containing acronyms and maintains consistency with the other entries.

PR #254, opened 2016-12-22
Add Standalone VMs and TRIM/discard

tasket:
Per issue #2479

### DIFF
--- a/managing-os/templates.md
+++ b/managing-os/templates.md
@@ -87,6 +87,17 @@ Important Notes
 
        $ sudo yum remove qubes-template-fedora-24
 
+ * Standalone VMs using Template VMs as a basis can be created easily. These
+   VMs receive a *copy* of the operating system and do not get automatically
+   updated when Template VMs are updated--they must be updated individually.
+   
+ * Template VMs can occupy more space on the dom0 filesystem than necessary
+   because they cannot employ automatic TRIM/discard on the root fs. The
+   `qvm-trim-template` command in dom0 is used to recover this unused space.
+
+   Conversely, the root filesystems in Standalone VMs *can* employ
+   TRIM/discard on the root fs using normal tools and configuration options.
+ 
  * On XFCE based Dom0, a manual action may be required to remove the "Start Menu"
    sub-menu of the removed TemplateVM. For example, to remove a dangling sub-menu
    for a removed "fedora-24" template, open a Dom0 Terminal and type:


managing-os/templates.md
```diff
@@ -87,6 +87,17 @@ Important Notes
 
        $ sudo yum remove qubes-template-fedora-24
 
+ * Standalone VMs using Template VMs as a basis can be created easily. These
+   VMs receive a *copy* of the operating system and do not get automatically
+   updated when Template VMs are updated--they must be updated individually.
+   
+ * Template VMs can occupy more space on the dom0 filesystem than necessary
+   because they cannot employ automatic TRIM/discard on the root fs. The
+   `qvm-trim-template` command in dom0 is used to recover this unused space.
```
jpouellet: Err... are you sure this is true?

IIUC the problem is that the CoW layer doesn't pass trim/discard, and IIUC this layer is not in use for a templatevm's root fs while the template itself has mounting it (r/w).

andrewdavidwong: Unless something changed recently, it's true that you have to use `qvm-trim-template` to trim templates.

andrewdavidwong: To clarify: I don't mean "have to" as in "that is the only possible way." I mean "have to" as in "it won't happen automatically unless you use `qvm-trim-template` or some other method."

jpouellet: Ah, well I suppose it is true regardless because we do not (yet?) enable discard on / at all. My question was incorrectly framed with the assumption that that change had been made. So nevermind :) We'll just need to remember to revisit this if that changes.

jpouellet: Actually maybe not. IIUC "they cannot employ automatic TRIM/discard on the root fs" is still false.

andrewdavidwong: Are you saying it should be "do not" instead of "cannot"?

jpouellet: Well... I seem [unable to reproduce](https://gist.github.com/anonymous/19416b4227f32d426c2bc52172aae156) my [results from a StandaloneVM](https://github.com/QubesOS/qubes-issues/issues/2479#issuecomment-264674619) in a TemplateVM's root filesystem mounted with discard enabled, so I must conclude that the statement is indeed correct.

Sorry for wasting your time.

jpouellet: > Are you saying it should be "do not" instead of "cannot"?

I was claiming that, yes. But I need to investigate deeper. Ultimately my goal is to have discard actually working rather than nitpick about wording in docs.

tasket:
@jpouellet Template VMs and Template-based appVMs both use the COW block layer (though they use them differently). This prevents discard from working.

Standalone VMs use normal block layer on root fs so they can process discard normally for it.

So might as well explain the difference so people know how to free up space in each VM type.

PR #258, opened 2017-01-05
faq: Add how to connect to WiFi

cgwalters:
Now, once one understands the system this does seem obvious, but
I'm not the only person who found this confusing, see e.g.:
https://groups.google.com/forum/#!topic/qubes-users/k8RceuBRwYU

### DIFF
--- a/basics_user/user-faq.md
+++ b/basics_user/user-faq.md
@@ -46,6 +46,7 @@ Qubes Users' FAQ
 
 [Common Problems](#common-problems)
 -----------------------------------
+ * [How do I connect to a wireless network?](#connect-wifi)
  * [My qubes lost Internet access after a TemplateVM update. What should I do?](#my-qubes-lost-internet-access-after-a-templatevm-update-what-should-i-do)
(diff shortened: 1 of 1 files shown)


jpouellet:
Err... what about nm-applet?

Putting this in the FAQ seems to suggest that recommended way to configure wireless is via the command line, and seems rather unfriendly towards non-technical users.

Is this because nm-applet did not show up for you? If so, that sounds like a bug which should be investigated.

cgwalters:
I didn't see it indeed.  I tried using the NM connection editor in the UI, but everything was greyed out.

andrewdavidwong:
@jpouellet is correct. Merging this PR was my mistake. I didn't realize that @cgwalters was actually attempting to document a workaround for a bug. I've now reverted the merge.

@cgwalters: I suspect that this bug is uncommon, since no one else has reported it. (https://github.com/QubesOS/qubes-issues/issues/2323 and the issues it references are somewhat related to your issue but not exactly the same.) If you would like to file an issue for this, you're welcome to do so.

cgwalters:
Ah, that makes sense - and indeed, now that I've rebooted and/or set up a network, I do see nm-applet. Agreed, this looks like https://github.com/QubesOS/qubes-issues/issues/2323 - feel free to revert.

PR #266, opened 2017-01-19
Document usage of precompiled  archlinux template

ptitdoc:


andrewdavidwong:
@marmarek

ptitdoc:
But the way, I did not finished the documentation as there is a slight problem in the binary template:

1/ In /etc/pacman.conf,  the following lines must be inserted just before the # REPOSITORIES line (this should be done by the qubes-core-agent installer script):
### QUBES CONFIG MARKER ###
Include = /etc/pacman.d/*.conf
### QUBES CONFIG END MARKER ###

Because of that, /etc/pacman.d configuration files are ignored (qubes proxy and qubes repositories)

2/ The link in /etc/pacman.d/99-qubes-repository-3.2.conf is not correct, however, I updated my repositories to at least install a fixed qubes-core-agent that will update this file to the right repository.

managing-os/templates/archlinux.md
```diff
@@ -18,367 +18,34 @@ to test new features/updates.
...
-*   **Transfer the install-templates.sh script file into Dom0**
-  *Note: as there is not a typical file transfer method for Dom0, for security reasons, this less than simple transfer function has to be used*
-
-    *   Swtich to Domo and open a terminal window.
 
-    **Note:** Take care when entering these cmd strings.  They are very long and have a number of characters that are easy to mix '**-**' vs '**.**' '**<u>T</u>emplates** (correct) vs **<u>t</u>emplates** (wrong) or **Template_**'(also wrong)  This script will also take care of transfering the actual template.rpm to Dom0 as well.
+Install
+-------
 
+A prebuilt template is available only for Qubes 3.2. Before Qubes 3.2, it should be compiled from source as described in [building-archlinux-template](/doc/building-archlinux-template/).
```
marmarek: I think this page is somehow outdated. Maybe better to replace it with those parts you're removing from here?

managing-os/templates/archlinux.md
```diff
@@ -18,367 +18,34 @@ to test new features/updates.
...
-<br>
-<br>
-
-##### **If everything went correct there should be a Archlinux template listed in your Qubes VM Manager** #####
-
-<br>
-<br>
-<br>
----------------
+    [user@dom0 ~]$ sudo qubes-dom0-update qubes-template-archlinux
```
marmarek: Add `--enablerepo=qubes-templates-community`.

marmarek:
Does your comment above means I should rebuild the template?

managing-os/templates/archlinux.md
ptitdoc: That is what I've done.

ptitdoc:
I need to verify why Include = /etc/pacman.d/*.conf is not present in the template. Maybe it is a bug.

managing-os/templates/archlinux.md
ptitdoc: I replaced the page (https://github.com/QubesOS/qubes-doc/pull/267)

PR #272, opened 2017-01-31
Clarifying syntax for adding and removing USB devices

Wikinaut:
the terms `device` and `vmname` which were in "<" and ">" were not rendered. I removed the characters. because it appears to be impossible to use the lesser and greater in markdown blocks.

If you however find a better method, please change it.

### DIFF
--- a/common-tasks/usb.md
+++ b/common-tasks/usb.md
@@ -90,7 +90,7 @@ follows:
 
  6.  In a dom0 console, detach the stick:
 
-        qvm-block -d &lt;device&gt; &lt;vmname&gt;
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
The problem there is improper indentation (or lack of other code block syntax), which results in the text not being rendered as a code block. The angled brackets should remain, since they indicate text that should be replaced in the command. Therefore, I'll correct this in a separate commit that closes this PR. If you'd like your other minor change (related to the "personal" VM) to be implemented, please submit it as a separate PR.

Wikinaut:
thank you for fixing this!

PR #275, opened 2017-02-03
Updated installation and usage instructions

ptitdoc:


ptitdoc:
The activation instruction could probably be improved if we preload my GPG key into the template (without necessarilly signing it).
This would avoid opening the firewall.

andrewdavidwong:
@marmarek, what do you think?

marmarek:
I don't know how key management in Archlinux works, but what about simply shipping a key in separate file and let the user import it?

ptitdoc:
In Archlinux, there are basically using GPG trust. They generate a keyring, import five developpers master keys, these keys signed other developers keys.
Archlinux also uses a specific package (archlinux-keyring) to do all these tasks automatically and offline.
I can probably create such a package, or as you suggest put the file somewhere. I will thing about that.

ptitdoc:
Such a package can also probably be used to change keys at some point.

PR #277, opened 2017-02-05
extended the salt documentation

john-david-r-smith:


configuration/salt.md
```diff
@@ -53,6 +55,123 @@ function that, when issued, will immediately halt the computer. There is another
...
+
+    qubesctl top.enabled
+
+And to disable one:
+
+    qubesctl top.disable my-new-vm
+
+To actually apply the state:
+
+    qubesctl state.highstate
```
marmarek: IMO it's better to point `qubesctl --all state.highstate` here, as you talk about configuring VMs and without `--all` option only dom0 will be configured.

configuration/salt.md
```diff
@@ -243,13 +354,67 @@ Template name on which given VM is based (if any).
...
+This is important when using a state with `cmd.run`.
+
+### Disk quota exceeded (when installing templates)
+If you install multiple templates you may encounter this error.
+The solution is to shut down the updatevm between each install.
+E.g.:
+
+    install template and shutdown updatevm:
+      cmd.run:
+      - name: sudo qubes-dom0-update -y fedora-24; qvm-shutdown {{salt.cmd.run(qubes-prefs updatevm) }}
```
marmarek: Adding `--clean` option to `qubes-dom0-update` should do too.

Also, we do support `pkg` module in dom0 - it's using `qubes-dom0-update` as its backend, so you can install dom0 packages the salt way.

john-david-r-smith: is there a way to pass the --clean flag via salt? (or is this done automatically?(if not, i think this would be a good choice))
i did not see an option to pass custom flags to pkg.installed.

am i correct in assuming pkg.installed fromrepo is handled correctly for dom0?

marmarek: > am i correct in assuming pkg.installed fromrepo is handled correctly for dom0?

Yes.

> is there a way to pass the --clean flag via salt? (or is this done automatically?(if not, i think this would be a good choice))

It's done when you specify `refresh: True`

jpouellet:
@john-david-r-smith Thank you for this! :)

PR #279, opened 2017-02-09
Add note on services in Debian templates and install problems

unman:
Closes QubesOS/qubes-issues/#2621

### DIFF
--- a/managing-os/templates/debian.md
+++ b/managing-os/templates/debian.md
@@ -12,9 +12,9 @@ redirect_from:
 Debian template(s)
 ===============
 
-If you like to use Debian Linux distribution in your AppVMs, you can install one of available Debian templates.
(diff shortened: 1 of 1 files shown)


adrelanos:
> Unattended Upgrades

Could you please add how to do this? Just one line

sudo apt-get purge unattended-upgrades

> - Install the ssh service.

If you do that - the keys are already generated and then will be shared
among all template based VMs. You'd have to wipe them from /etc/ssh. And
this is specific per daemon.

What about not starting the daemon during apt-get installation?

(You can get the idea on how to do that by looking at
apt-get-noninteractive. [1] [2])

Then you would not have to run systemctl stop/disable. Only mask.

[1]
https://github.com/Whonix/usability-misc/blob/master/usr/bin/apt-get-noninteractive
[2]
https://github.com/Whonix/usability-misc/blob/master/usr/share/usability-misc/policy-rc.d

PR #280, opened 2017-02-09
Add note on installing services in Debian templates and upgrade issues

unman:
Closes QubesOS/qubes-issues#2621

### DIFF
--- a/managing-os/templates/debian.md
+++ b/managing-os/templates/debian.md
@@ -12,9 +12,9 @@ redirect_from:
 Debian template(s)
 ===============
 
-If you like to use Debian Linux distribution in your AppVMs, you can install one of available Debian templates.
+If you would like to use Debian Linux distribution in your qubes, you can install one of the available Debian templates.
 
-Updates for this template are provided by ITL and are signed by this key:
+Updates for these templates are provided by ITL and are signed by this key:
 
     pub   4096R/47FD92FA 2014-07-27
           Key fingerprint = 2D43 E932 54EE EA7C B31B  6A77 5E58 18AB 47FD 92FA
@@ -28,7 +28,7 @@ which is also integrity-protected using signed git tags.
 Install
 -------
 
-It can be installed via the following command:
+Templates can be installed with the following command:
 
 Debian 7 (wheezy) - old stable:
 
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
@unman: Since you make frequent and valuable documentation contributions, please consider PGP-signing your Git commits (or tags).

adrelanos:
> Unattended Upgrades

Could you please add how to do this? Just one line

sudo apt-get purge unattended-upgrades

> - Install the ssh service.

If you do that - the keys are already generated and then will be shared
among all template based VMs. You'd have to wipe them from /etc/ssh. And
this is specific per daemon.

What about not starting the daemon during apt-get installation?

(You can get the idea on how to do that by looking at
apt-get-noninteractive. [1] [2])

Then you would not have to run systemctl stop/disable. Only mask.

[1]
https://github.com/Whonix/usability-misc/blob/master/usr/bin/apt-get-noninteractive
[2]
https://github.com/Whonix/usability-misc/blob/master/usr/share/usability-misc/policy-rc.d

unman:
On Thu, Feb 09, 2017 at 05:08:23PM -0800, Andrew David Wong wrote:
> @unman: Since you make frequent and valuable documentation contributions, please consider PGP-signing your Git commits (or tags).
> 

Sometimes I just make the edits in the browser.
I'll stop doing this.

unman:
On Thu, Feb 09, 2017 at 05:43:04PM -0800, Patrick Schleizer wrote:
> 
> 
> > Unattended Upgrades
> 
> Could you please add how to do this? Just one line
> 
> sudo apt-get purge unattended-upgrades
> 
> > - Install the ssh service.
> 
> If you do that - the keys are already generated and then will be shared
> among all template based VMs. You'd have to wipe them from /etc/ssh. And
> this is specific per daemon.
> 
> What about not starting the daemon during apt-get installation?
> 
> (You can get the idea on how to do that by looking at
> apt-get-noninteractive. [1] [2])
> 
> Then you would not have to run systemctl stop/disable. Only mask.
> 
> [1]
> https://github.com/Whonix/usability-misc/blob/master/usr/bin/apt-get-noninteractive
> [2]
> https://github.com/Whonix/usability-misc/blob/master/usr/share/usability-misc/policy-rc.d
> 
> 
I've toyed with the idea of using policy-rc.d in the Templates - we use
it in the build process but then remove it.

adrelanos:
To improve usability, we could ship the policy-rc.d file with Qubes
(package qubes-core-agent or so). Of course in a path where it does
nothing by default.

And also ship a wrapper similar to apt-get-noninteractive. Not
noninteractive. Still interactive like normal apt. But not starting the
services at package installation if this wrapper is being used. And then
recommend using that wrapper in documentation.

andrewdavidwong:
> Sometimes I just make the edits in the browser. I'll stop doing this.

If that would discourage you from contributing (e.g., by making your workflow significantly more laborious or less pleasant), then there's no need to stop. I would rather see frequent unsigned contributions than fewer signed contributions. :)

adrelanos:
That's great to hear! I'd also sign up for unsigned docs changes. A web
edit is a lot quicker and more comfortable.

unman:
On Sun, Feb 12, 2017 at 07:56:01AM -0800, Patrick Schleizer wrote:
> That's great to hear! I'd also sign up for unsigned docs changes. A web
> edit is a lot quicker and more comfortable.
> 
> 

And I've just determined that I will move to signing all my doc
changes.
:-(

andrewdavidwong:
> That's great to hear! I'd also sign up for unsigned docs changes. A web edit is a lot quicker and more comfortable.

There's actually nothing to sign up for. Unsigned doc changes are already accepted. All you need is a free GitHub account, which you obviously already have. :)

> And I've just determined that I will move to signing all my doc changes. :-(

@adrelanos's decision to refrain from signing his changes is independent of yours. I applaud your determination to begin signing all of your doc changes. :)

PR #282, opened 2017-02-10
Clarify issue arising using secondary disks

unman:


andrewdavidwong:
I don't see how this is clearer. It seems equivalent to me (and perhaps even slightly worse grammatically). Am I missing something?

unman:
On Fri, Feb 10, 2017 at 10:38:51PM -0800, Andrew David Wong wrote:
> I don't see how this is clearer. It seems equivalent to me (and perhaps even slightly worse grammatically). Am I missing something?
> 
For some reason, some people read the existing text as saying that the METHOD
wont work on 3.2 and above, rather than that this ISSUE doesn't arise
there.
Most recently on qubes-users, but I've had this discussion with a few
people.

It seems clear to me as is, but this is an effort to make it still clearer.

andrewdavidwong:
> For some reason, some people read the existing text as saying that the METHOD wont work on 3.2 and above, rather than that this ISSUE doesn't arise there. Most recently on qubes-users, but I've had this discussion with a few people.

Interesting. Can you provide links to those posts?

While I'm all in favor of genuine clarifications, I don't think we should start catering to people's illiteracy (if that's the cause here). In other words, if it's a problem with the text, then the onus is on us to fix it, but if it's a problem with the reader, then the onus is on the reader to read more carefully.

unman:
Off list comment after recent thread - "I did see that document (this page), but dismissed it as I understood it did not apply to R3.2 and later."
I've had similar comments from probably 3 other people, so there must be something about the page, although it seems clear enough to me. I'll leave it to you Andrew.

PR #284, opened 2017-02-12
un-abbreviated title, made generic example

mfc:
content itself is still a mess, with "Disposable VM", "DispVM", and "DVM" all used. At some point we should probably agree on an abbreviation and propagate it.

### DIFF
--- a/customization/dispvm-customization.md
+++ b/customization/dispvm-customization.md
@@ -1,6 +1,6 @@
 ---
 layout: doc
-title: DispVM Customization
+title: Disposable VM Customization
 permalink: /doc/dispvm-customization/
 redirect_from:
 - /en/doc/dispvm-customization/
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
I think you may have missed this: https://github.com/QubesOS/qubes-issues/issues/2486#issuecomment-279191179

The usage of "DVM Template" there is intentional and should **not** be reverted.

Please see the glossary:
https://www.qubes-os.org/doc/glossary/#dispvm
https://www.qubes-os.org/doc/glossary/#dvm
https://www.qubes-os.org/doc/glossary/#dvm-template

mfc:
missed that discussion, thanks. so it sounds like don't merge that one section title change?

andrewdavidwong:
Yes, please add a new commit to undo that part.

mfc:
fixed version: https://github.com/QubesOS/qubes-doc/pull/285

PR #295, opened 2017-03-02
Detailed step-by-step guide for MacBookPro

mpodroid:
Added detailed guide to install and confgure Qubes OS 3.2 on MacBook Pro Retina 15" mid-2015 11,5

### DIFF
--- a/troubleshooting/macbook-troubleshooting.md
+++ b/troubleshooting/macbook-troubleshooting.md
@@ -148,6 +148,9 @@ In this section, I explain how I installed Qubes 3.2 on a MacBook Pro Retina 201
 Good news: the relevant stuff works.
 Bad news: still some minor issue to investigate.
 
+For the time being, my setup is just for testing purposes and help to bypass some blocking issues: do not use it in production or on machine where security is a concern!
+I hope to improve it as soon as possible.
+
 During my nigths trying to get Qubes OS working, I faced tow main and blocking issues:
 *   no boot, due to empty xen.cfg file
 *   system freeze, due to Broadcom BCM43602 wifi card
@@ -178,11 +181,10 @@ Download and prepare a USB with Qubes 3.2
 
 You can install Qubes using BIOS or UEFI:
 *  BIOS/CSM/Legacy: I have not been able to install using legagy, but I did not spent a lot of time on it.
-   * power on the macbook pressing Option button
-   * GRUB menu appears, with option to test media & install, install, troubleshoot and rescue
-   * whatever option I choose, I have a quick flash and back to the menu
+*  UEFI plain: grub menu appears, but any gave me a quick flash and returned the main menu. I can boot it manually fixing the grub.cfg file, adding commands linuexefi and initrdefi, pointing proper files in /efi/boot. After boot, I end up with not root file system.
 *  UEFI, using rEFInd: I have been successful, despite some issues to be fixed manually, after installation completion
-   * download [rEFInd] ZIP file, unzip it and run installer
+   * download [rEFInd] refind-bin-0.10.4.zip: this file is not signed, so decide if you trust it or not. SHA1 sum is 3d69c23b7d338419e5559a93cd6ae3ec66323b1e  
+   * unzip it and run installer, which install rEFIind on the internal SSD
    * if installation fails due to SIP, reboot in recovery mode, open a terminal and issue command
    ~~~
    crsutil disable
@@ -251,7 +253,7 @@ chrooot /mnt/sysimage
 default=4.4.14-11.pvops.qubes.x868_64
 
 [4.4.14-11.pvops.qubes.x868_64]
-options=loglvl=all
+options=loglvl=all dom0_mem=min:1024M dom0_mem=max:4096M
 kernel=vmlinuz-4.4.14-11.pvops.qubes.x86_64
 ramdisk=initramfs-4.4.14-11.pvops.qubes.x86_64.img
 ~~~
@@ -275,8 +277,11 @@ Everything should now be ok, Qubes OS boots using EFI and you will get the last 
 ### 6. Fix pulseaudio, which locks CPU freezing the system often for 20 seconds
 
 My macbook has frequent freezes. Looking at journalctl output I saw that pulseaudio locks CPU for 20 seconds, very often.
-The quickest workaround, is to open a dom0 terminal, as root and edit /etc/pulse/client.conf and adding "autospawn = no"
-Then, as normal user, issue command "pulseaudio --kill"
+
+You can fix this issue, killing audio support with this quick workaround:
+*  open a dom0 terminal, as root and edit /etc/pulse/client.conf
+*  add "autospawn = no"
+*  Then, as normal user, issue command "pulseaudio --kill"
 
 ### 7. Fix system freezes due to Broadcom BCM43602
 
--- a/troubleshooting/macbook-troubleshooting.md
+++ b/troubleshooting/macbook-troubleshooting.md
@@ -139,3 +139,182 @@ Please see [this thread o the qubes-devel mailing list][macbook-air-2012-5-1].
 [bluetooth-replacement]: https://www.ifixit.com/Guide/MacBook+Air+13-Inch+Mid+2011+AirPort-Bluetooth+Card+Replacement/6360
 [macbook-air-2012-5-1]: https://groups.google.com/d/topic/qubes-devel/uLDYGdKk_Dk/discussion
 
+
+
+MacBook Pro Retina, 15 inch, Mid-2015 (MacBookPro 11,5)
+-------------------------------------------------------
+
+In this section, I explain how I installed Qubes 3.2 on a MacBook Pro Retina 2015 (MacBookPro 11,5).
+Good news: the relevant stuff works.
+Bad news: still some minor issue to investigate.
+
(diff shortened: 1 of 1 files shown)


troubleshooting/macbook-troubleshooting.md
```diff
@@ -139,3 +139,182 @@ Please see [this thread o the qubes-devel mailing list][macbook-air-2012-5-1].
...
+
+Download and prepare a USB with Qubes 3.2
+
+You can install Qubes using BIOS or UEFI:
+*  BIOS/CSM/Legacy: I have not been able to install using legagy, but I did not spent a lot of time on it.
+   * power on the macbook pressing Option button
+   * GRUB menu appears, with option to test media & install, install, troubleshoot and rescue
+   * whatever option I choose, I have a quick flash and back to the menu
+*  UEFI, using rEFInd: I have been successful, despite some issues to be fixed manually, after installation completion
+   * download [rEFInd] ZIP file, unzip it and run installer
```
marmarek: Some warning should be here. About using not signed, binary, downloaded from sf.net.
Maybe add a hash of this zip here, for a minimal verification (that at least everyone get the same zip)?

Does it install rEFInd on main disk, or USB installation media? This info should be added here.

troubleshooting/macbook-troubleshooting.md
```diff
@@ -139,3 +139,182 @@ Please see [this thread o the qubes-devel mailing list][macbook-air-2012-5-1].
...
+
+### 2. Boot installer
+
+Download and prepare a USB with Qubes 3.2
+
+You can install Qubes using BIOS or UEFI:
+*  BIOS/CSM/Legacy: I have not been able to install using legagy, but I did not spent a lot of time on it.
+   * power on the macbook pressing Option button
+   * GRUB menu appears, with option to test media & install, install, troubleshoot and rescue
+   * whatever option I choose, I have a quick flash and back to the menu
```
marmarek: Grub (instead of isolinux) suggest that it wasn't legacy mode, but UEFI. Maybe removing `EFI` directory from installation image would force legacy mode, but it will also make image verification fail. I think better write explicitly that legacy mode does not work here.

troubleshooting/macbook-troubleshooting.md
```diff
@@ -139,3 +139,182 @@ Please see [this thread o the qubes-devel mailing list][macbook-air-2012-5-1].
...
+   * reboot and you will see some icons
+   * choose Boot EFI\BOOT\xen.efi from ANACONDA
+   * after a will the graphical installer is up and running, with keyboard and touchpad working
+
+### 3. Installation
+
+*  As a general rule, keep the default values proposed during installation: you will change them later on
+*  Keep English, as language, locale
+*  My macbook has a US keyboard, so I cannot say what happens if you change keyboard layout
+*  DO NOT CHANGE the timezone, because it will trigger the wifi card, leading to a system freeze
```
marmarek: Ough... if that's really the case, could you create an issue in qubes-issues for that? Or maybe you already done that?

troubleshooting/macbook-troubleshooting.md
```diff
@@ -139,3 +139,182 @@ Please see [this thread o the qubes-devel mailing list][macbook-air-2012-5-1].
...
+At rEFInd screen, choose Boot EFI/qubes/xen-4.6.1.efi
+Everything should now be ok, Qubes OS boots using EFI and you will get the last setup screen
+*  select "Qubes OS", do not change anything and click on "Done"
+*  VMs are created, including NetVM
+
+### 6. Fix pulseaudio, which locks CPU freezing the system often for 20 seconds
+
+My macbook has frequent freezes. Looking at journalctl output I saw that pulseaudio locks CPU for 20 seconds, very often.
+The quickest workaround, is to open a dom0 terminal, as root and edit /etc/pulse/client.conf and adding "autospawn = no"
+Then, as normal user, issue command "pulseaudio --kill"
```
marmarek: This will kill audio support. Should be noted here.

troubleshooting/macbook-troubleshooting.md
```diff
@@ -139,3 +139,182 @@ Please see [this thread o the qubes-devel mailing list][macbook-air-2012-5-1].
...
+~~~
+chrooot /mnt/sysimage
+~~~
+* edit /boot/efi/EFI/qubes/xen.cfg file putting the following content
+~~~
+[global]
+default=4.4.14-11.pvops.qubes.x868_64
+
+[4.4.14-11.pvops.qubes.x868_64]
+options=loglvl=all
```
marmarek: Add `dom0_mem=min:1024M dom0_mem=max:4096M` to `options` (it's default in 3.2, was missing in 3.1).

troubleshooting/macbook-troubleshooting.md
```diff
@@ -139,3 +139,182 @@ Please see [this thread o the qubes-devel mailing list][macbook-air-2012-5-1].
...
+
+Ok, setup is complete and we are almost done.
+* Open a dom0 terminal
+~~~
+sudo su -
+xl pci-assignable-list
+echo 04:00.0 > /sys/bus/pci/drivers/pciback/permissive
+qvm-start sys-net
+xl pci-attach sys-net 04:00.0
+~~~
```
marmarek: See https://www.qubes-os.org/doc/assigning-devices/#possible-issues - there is already better way to workaround it - without removing then re-adding device to sys-net.

troubleshooting/macbook-troubleshooting.md
mpodroid: You are right. I just retested on my Macbook Pro early 2011 (8,2): it boots grub2-efi.
Only the second partition (hd0,msdos2) on USB stick is visible, the first is not even listed. Strange.
Apparently something is wrong with grub.cfg: it contains only chainloader command and does not boot. If I add linuxefi and initrdefi commands pointing to proper file in /efI/boot I can boot the installer, but it fails after a few seconds:
"Failed to swith root: Specified switch root path /sysroot does not seem to be an OS tree. os-release file is missing"

troubleshooting/macbook-troubleshooting.md
mpodroid: I already tried all of them them: no way. If NetVM start with broadcom attached the system freezed and I have to do and reset. I did not find any interesting long after the freeze.
The only working workaround that I found is to attach the device after NetVM startup.

troubleshooting/macbook-troubleshooting.md
mpodroid: created https://github.com/QubesOS/qubes-issues/issues/2668

troubleshooting/macbook-troubleshooting.md
marmarek: Using `linuxefi` does not boot Xen... `chainloader` is the only way of booting Xen on EFI (until multiboot2 protocol will be supported in Xen), but it does not work on some hardware.

troubleshooting/macbook-troubleshooting.md
marmarek: Does the device work after such dynamic attach? Generally dynamic attach of PCI devices is something rather fragile... Maybe you've done something wrong while setting up `qubes-pre-netvm.service`? You should be able to check if permissive mode is enabled by reading this file.
Also, just to be sure: try to enable permissive mode (just like you do), but shutdown sys-net, add the device back to it (using qvm-pci or so), and start again.

mpodroid: After dynamic attach, BCM4302 worked without issues. I did not do any specific tests or benchmark, but I updated templateVMs and dom0, without noticing any glitches.
Dynamic attachment is the only workaround I found after these attempts:
* created service qubes-pre-netvm to enable permissive mode: still system freeze
* renamed /var/lib/qubes/servicevms/sys-net to avoid booting it, which freeze the system
* then I discovered that NetVM was triggered by service qubes-netvm which I disabled, to be able to boot Qubes
* manually enabled permissive mode, launched sys-net => system freeze
* manually enabled permissive mode, used qvm-pci to attach device, launched sys-net => system freeze
Let me know if there some relevant logs to collect to troubleshoot the freeze

marmarek: > manually enabled permissive mode, used qvm-pci to attach device, launched sys-net => system freeze

Ok, if you checked this one, then indeed dynamic attach is the only option left. It's still strange that interaction with a single PCI device cause whole system freeze. In theory it should at most cause a single VM (sys-net) freeze... VT-d is specifically to isolate devices from the rest of the system. Maybe VT-d is somehow broken on MacBook? Could you check if `xl dmesg` output contains something like this:
```
(XEN) I/O virtualisation enabled
(XEN)  - Dom0 mode: Relaxed
(XEN) Interrupt remapping enabled
```
?

As for freeze-related logs - I'm afraid nothing will hit the disk when the system is in such state. But you could check `/var/log/xen/console/hypervisor.log` and `/var/log/xen/console/guest-sys-net.log`.

mpodroid: Forgot to mention, that I also tested the NetVM with static assignment with kernel 4.4.38-11 and 4.8.9-12 from unstable repo, both in NetVM and in dom0: no luck.
I also tried to use NetVM based on fc24 template: still no luck. 
I am afraid that something wrong happens in dom0.

PR #299, opened 2017-03-03
Clarify some issues regarding bind-dirs

unman:
QubesOS/qubes-issues#2661

### DIFF
--- a/customization/bind-dirs.md
+++ b/customization/bind-dirs.md
@@ -8,14 +8,19 @@ redirect_from:
 
 # How to make any file in a TemplateBasedVM persistent using bind-dirs #
 
-## What is bind-dirs.sh? ##
+## What is bind-dirs? ##
 
-With [bind-dirs.sh](https://github.com/QubesOS/qubes-core-agent-linux/blob/master/vm-systemd/bind-dirs.sh)
-you can make arbitrary files or folders persistent in TemplateBasedVMs.
+With [bind-dirs](https://github.com/QubesOS/qubes-core-agent-linux/blob/master/vm-systemd/bind-dirs.sh)
+any arbitrary files or folders can be made persistent in TemplateBasedVMs.
 
 ## What is it useful for? ##
 
-For example, it is useful for Whonix, sys-whonix, where [Tor's data dir /var/lib/tor has been made persistent in the TemplateBased ProxyVM sys-whonix](https://github.com/Whonix/qubes-whonix/blob/8438d13d75822e9ea800b9eb6024063f476636ff/usr/lib/qubes-bind-dirs.d/40_qubes-whonix.conf#L5). So sys-whonix does not require to be a StandaloneVM. And therefore can benefit from the Tor anonymity feature 'persistent Tor entry guards' without the overhead of a StandaloneVM.
+In a TemplateBasedVM all of the file system comes from the template except /home and /usr/local.
+This means that changes in the rest of the filesystem are lost when the TemplateBasedVM is shutdown.
+bind-dirs provides a mechanism whereby files usually taken from the template can be persisted across reboots.
+
+For example, in Whonix, [Tor's data dir /var/lib/tor has been made persistent in the TemplateBased ProxyVM sys-whonix](https://github.com/Whonix/qubes-whonix/blob/8438d13d75822e9ea800b9eb6024063f476636ff/usr/lib/qubes-bind-dirs.d/40_qubes-whonix.conf#L5). 
+In this way sys-whonix can benefit from the Tor anonymity feature 'persistent Tor entry guards' but does not have to be a StandaloneVM.
 
 ## Minimum Qubes Version ##
 
@@ -29,26 +34,18 @@ Inside your TemplateBasedVM.
 
        sudo mkdir -p /rw/config/qubes-bind-dirs.d
 
-2. Create a file `/rw/config/qubes-bind-dirs.d/50_user.conf` with root rights inside a VM.
+2. Create a file `/rw/config/qubes-bind-dirs.d/50_user.conf` with root rights inside a TemplateBasedVM.
 
-       sudo nano /rw/config/qubes-bind-dirs.d/50_user.conf
-
-3. Append a folder or file to the `binds` variable. In the following example we are using folder `/var/lib/tor`. You can replace that folder with a folder of your choice.
-
-       binds+=( '/var/lib/tor' )
-       
-In the following example we are using file `/etc/tor/torrc`. You can replace that file with a file of your choice.
-       
-       binds+=( '/etc/tor/torrc' )
-
-Multiple configuration lines, files and folders supported. For example, you could use both.
+3. Edit the file 50_user.conf to append a folder or file name to the `binds` variable. (In the following example we are using folder `/var/lib/tor`. You can replace that name with a folder or file name of your choice.)
 
        binds+=( '/var/lib/tor' )
        binds+=( '/etc/tor/torrc' )
+
+Multiple entries are possible, each on a separate line.
 
 4. Save.
 
-5. Reboot your TemplateBasedVM.
+5. Reboot the TemplateBasedVM.
 
 6. Done.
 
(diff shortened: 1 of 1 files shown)


customization/bind-dirs.md
```diff
@@ -8,14 +8,19 @@ redirect_from:
 
 # How to make any file in a TemplateBasedVM persistent using bind-dirs #
 
-## What is bind-dirs.sh? ##
+## What is bind-dirs? ##
 
-With [bind-dirs.sh](https://github.com/QubesOS/qubes-core-agent-linux/blob/master/vm-systemd/bind-dirs.sh)
-you can make arbitrary files or folders persistent in TemplateBasedVMs.
+With [bind-dirs](https://github.com/QubesOS/qubes-core-agent-linux/blob/master/vm-systemd/bind-dirs.sh)
+any arbitrary files or folderscan be made persistent in TemplateBasedVMs.
```
marmarek: missing space

customization/bind-dirs.md
```diff
@@ -45,16 +52,32 @@ bind-dirs.sh works with Qubes R3.2 and above.
...
 * `/rw/config/qubes-bind-dirs.d` (highest priority, for per VM configuration)
 
+## How does it work? ##
+
+bind-dirs.sh is called on startup of a TemplateBasedVM, and configuration files in the configuration folders above are parsed to build a bash array.
+Files or folders identified in the array are copied to /rw/bind-dirs if they do not already exist there, and are then bind mounted over the original files/folders.
+
+Creation of the file and folders in /rw/bind-dirs should be automatic the first time the TemplateBasedVM is restarted after configuration.
+
+If you want to circumvent this process, you can create the relevant filestructure under /rw/bind-dirs and make any changes at the same time that you perform the configuration, before reboot.
```
marmarek: I think much easier option is to edit configuration and then launch bind-dirs script manually to create appropriate directories structure. It's easier to do and harder to make a typo. But see limitations below.
Do you think it worth mentioning this simpler method, maybe with a reference to a limitation below + a ticket (AFAIK there is none for it yet)?

customization/bind-dirs.md
```diff
@@ -45,16 +52,32 @@ bind-dirs.sh works with Qubes R3.2 and above.
...
+
+
 ## Limitations ##
 
-* Files that exist in the TemplateVM root image cannot be made deleted in the TemplateBasedVMs root image using bind-dirs.sh.
-* Does not work if the file / folder in question does not already exist in the root image. I.e. a file that does not exist in the root image cannot be bind mounted in the TemplateBasedVM.
-* Re-running `sudo /usr/lib/qubes/bind-dirs.sh` without previous `sudo /usr/lib/qubes/bind-dirs.sh umount` does not work.
+* Files that exist in the TemplateVM root image cannot be deleted in the TemplateBasedVMs root image using bind-dirs.sh.
+* The file / folder in question must already exist in the root image. I.e. a file that does not exist in the root image cannot be bind mounted in the TemplateBasedVM.
+* Re-running `sudo /usr/lib/qubes/bind-dirs.sh` without a previous `sudo /usr/lib/qubes/bind-dirs.sh umount` does not work.
```
marmarek: Actually, if you created bind-dirs configuration for the first time (it was empty during VM startup), it works fine. But in other cases, it will indeed result in some weird situation (multiple mounts in the same place etc).

customization/bind-dirs.md
```diff
@@ -45,16 +52,32 @@ bind-dirs.sh works with Qubes R3.2 and above.
...
 * Running 'sudo /usr/lib/qubes/bind-dirs.sh umount' after boot (before shutdown) is probably not sane and nothing can be done about that.
+* Many editors create a temporary file and copy it over the original file. If you have bind mounted an individual file this will break the mount.
```
marmarek: In many cases it's easy to workaround: configure the whole directory for bind-dirs. Of course it isn't feasible if said directory would be the whole `/etc` or such...

adrelanos: Ideally, bind-dirs would be as robust as supporting even `/`. :)

customization/bind-dirs.md
adrelanos: Perhaps bind-dirs should figure out the state etc.. If already mounted, avoid duplicate mount etc.

(And if it is not possible to figure out the state - keep a state file with the state.)

marmarek: There is `mountpoint` tool which can check if a directory or file is a mountpoint. That would be a first step. The second one (less important here): checking if mounted thing is a file/directory from `/rw/bind-dirs`.

adrelanos: Created https://github.com/QubesOS/qubes-issues/issues/2673 for it. (Feel free to edit any of my tickets in case I misquoted.)

customization/bind-dirs.md
adrelanos: Would https://github.com/QubesOS/qubes-issues/issues/2673 suffice as the ticket you'd reference there or did you have something different in mind?

adrelanos:
> If you want to circumvent this process, you can create the relevant filestructure under /rw/bind-dirs

With any examples this is still difficult to figure out.

Anyhow. These changes are okay and I'd appreciate if merged. I am a fan of many small iterations with improvements over time, wiki style. We could still add this on top later.

andrewdavidwong:
@unman or @adrelanos: Please resolve the nontrivial conflicts.

customization/bind-dirs.md
jpouellet: Just here to say that noting that the persisted data lives in /rw/bind-dirs is a useful thing to document, regardless of whether bind-dirs.sh is idempotent.

customization/bind-dirs.md
ypid: > copy it over the original file. If you have bind mounted an individual file this will break the mount.

This is not the actual limitation as I observed it (with `/etc/default/dnsmasq` being a bind mounted file by bind-dirs):

```
root@s-cm:/etc/default# cp dnsmasq dnsmasq-copy
root@s-cm:/etc/default# cp dnsmasq-copy dnsmasq

root@s-cm:/etc/default# strace -e trace=file mv dnsmasq-copy dnsmasq
[...]
rename("dnsmasq-copy", "dnsmasq")       = -1 EBUSY (Device or resource busy)
mv: cannot move ‘dnsmasq-copy’ to ‘dnsmasq’open("/usr/share/locale/en_US.UTF-8/LC_MESSAGES/libc.mo", O_RDONLY) = -1 ENOENT (No such file or directory)
: Device or resource busy
+++ exited with 1 +++

root@s-cm:/etc/default# strace -e trace=file rm dnsmasq
[...]
unlinkat(AT_FDCWD, "dnsmasq", 0)        = -1 EBUSY (Device or resource busy)
rm: cannot remove ‘dnsmasq’open("/usr/share/locale/en_US.UTF-8/LC_MESSAGES/libc.mo", O_RDONLY) = -1 ENOENT (No such file or directory)
: Device or resource busy
+++ exited with 1 +++
```

Moving/recreating is also a thing that config management systems like to do after templating a file (and validating it) which is also effected by this:

```
ansible-playbook -l s-cm ansible/ypid-playbooks/service/dnsmasq-qubes_os.yml
[...]
TASK [debops.dnsmasq : Configure defaults for dnsmasq service] *****************
fatal: [s-cm]: FAILED! => {"changed": true, "failed": true, "msg": "Could not replace file: /home/user/.ansible/tmp/ansible-tmp-1489657252.52-223355023089868/source to /etc/default/dnsmasq:
[Errno 16] Device or resource busy"}
```

Maybe generalize the point to: Deleting/renaming of bind mounted files/directories does not work. Programs might do this internally when updating a file by first creating a new file and them doing a `rename` or `unlinkat`. Examples include editors and configuration management systems. A workaround might be to bind mount the directory instead or do the file operations in the rw location (`/rw/bind-dirs/`).

customization/bind-dirs.md
```diff
@@ -45,16 +52,32 @@ bind-dirs.sh works with Qubes R3.2 and above.
...
+Any changes you make will not survive a reboot. If you think it likely you will want to edit a file, then either include the parent directory in bind-dirs.rather than the file, or perform the file operation on the file in /rw/bind-dirs.
+* Some files are altered when a qube boots - e.g. /etc/hosts. If you try to use bind-dirs on such files you may break your qube in unpredictable ways.
```
ypid: Maybe change "bind-dirs.rather than the file, or perform the file operation on the file in /rw/bind-dirs." to "bind-dirs rather than the file, or perform the file operation on the file in /rw/bind-dirs which will require a unmount/remount for the changes to become visible)."

unman: @ypid I take your point, but since we have highlighted the fact that rerunning the script without umount doesnt work, and umount isn't sane, it's not appropriate at the moment. Many users will be able to use this without knowing how to umount/remount an individual file - that's what this script is for. (Note that the situation may change.)

unman:
@andrewdavidwong  I think you should merge it now. (Patrick is happy as you can see).
we can deal with any further issues or changes as they arise.

andrewdavidwong:
@unman: It looks like there are still nontrivial conflicts that I need you to resolve first.

customization/bind-dirs.md
ypid: Thanks for explaining.

andrewdavidwong:
Thanks, @unman and all!

PR #300, opened 2017-03-04
Add --keyserver command line flag to download developer key

colinnewell:
That prevents the `gpg: keyserver receive failed: No keyserver available` error.


andrewdavidwong:
The `--keyserver` flag is already introduced and explained earlier in the document, so I don't think this is necessary.

(Your additional also makes the example inconsistent. Note the second line.)

PR #304, opened 2017-03-05
remove "TODO" Firewall and Proxy VM section

mfc:
or else link to existing content.


andrewdavidwong:
This is developer documentation. @marmarek, how do you want this handled?

marmarek:
This page is really outdated. I think it was already outdated at Qubes 1.0... Can we drop it instead?
Very similar information is on https://www.qubes-os.org/doc/firewall/, which is much more up to date.
Or at least drop this, and the whole "Location of the network driver domain" section.

andrewdavidwong:
Ok, removed the whole "Location of the network driver domain" section.

PR #305, opened 2017-03-05
Add Fedora 23 EOL and 24 upgrade instructions to 3.2 release notes

ssixty:
Making this info more visible for those who might not read the news posts.


andrewdavidwong:
@marmarek: Do we have any policy about changing release notes retroactively?

marmarek:
We don't have such policy, but I think it's a good idea to include this information.

PR #309, opened 2017-03-09
Formatting fix

Yethal:


andrewdavidwong:
You should use backticks instead.

andrewdavidwong:
Actually, I think that whole string should be excluded. It's redundant given the command that follows immediately afterward.

Yethal:
I'll do better next time.

andrewdavidwong:
@Yethal: You did fine! We appreciate your contributions. :)

PR #310, opened 2017-03-11
Updates to USB and DisposableVm docs

Yethal:
Reorganized and updated USB documentation

Added instructions for DisposableVM Application Menu customization


andrewdavidwong:
Thanks! Two requests:

1. Please move the DispVM menu instructions to `/doc/dispvm-customization/` instead.
2. Please do not use HTML `<br>` tags. Instead, format those sections as Markdown code blocks (indent by 4 spaces (8 in a list), or wrap with three backticks or tildes before and after).

common-tasks/usb.md
```diff
@@ -147,162 +297,6 @@ steps:
...
-4. Save and close the file.
-5. Run the command `grub2-mkconfig -o /boot/grub2/grub.cfg` in dom0.
-6. Reboot.
-
-(Note: Beginning with R3.2, `rd.qubes.hide_all_usb` is set automatically if you
-opt to create a USB qube during installation. This also occurs automatically if
-you choose to [create a USB qube] using the `qubesctl` method, which is the
-first pair of steps in the linked section.)
-
-Supported USB device types
```
andrewdavidwong: I don't think this section should be removed.

common-tasks/usb.md
```diff
@@ -17,6 +17,160 @@ redirect_from:
...
+
+Add a line like this one to the top of the file:
+
+    sys-usb dom0 ask,user=root
+
+(Change `sys-usb` to your desired USB qube.)
+
+You can now use your USB keyboard.
+
+How to use a USB mouse
```
andrewdavidwong: Mouse passthrough is not as risky as keyboard passthrough, but it is still a risk. Perhaps there should be some kind of warning here. @marmarek?

common-tasks/usb.md
Yethal: I removed the following part because it directly contradicts the USB passthrough guide that states that any USB device can be passed through: 

> "-As of Qubes R3.1, it is possible to attach:
> -
> - * USB mice
> - * USB keyboards (see below)
> - * USB block devices (such as USB mass storage devices)
> -   * When attaching one of these, you should get a notification about the
> -     new device, then you should be able to attach it to a qube in Qubes VM
> -     Manager.
> -
> -Other devices, such as USB webcams, will also work, but they will be
> -accessible only from the USB qube itself, as explained above."

andrewdavidwong: Ah, I see. That's fine, then.

andrewdavidwong: (I think it's only the last sentence that contradicts the other part, though. As of R3.1, it *is* possible to attach those types of devices. It's just that R3.2 introduces the ability to pass through even more. But it's probably better just to remove the whole section.)

Yethal: I agree, however R3.1 is about to go EOL and that sentence is no longer true in R3.2.

andrewdavidwong: I think you may have misread my comment. I think we're actually in agreement on all points here.

common-tasks/dispvm.md
```diff
@@ -45,8 +45,7 @@ Use the `qvm-open-in-dvm` command line (from your AppVM), e.g.:
 [user@work-pub ~]$ qvm-open-in-dvm Downloads/apple-sandbox.pdf
 ~~~
 
-The qvm-open-in-dvm will not exit until you close the application in the Disposable VM.
-
+The qvm-open-in-dvm will not exit until you close the application in the Disposable V
```
andrewdavidwong: Typo.

Yethal:
Since this PR is slowly becoming a mess (mostly due to my weak GitHub skills) I've removed all DispVM related changes and moved them to a separate branch.

common-tasks/usb.md
marmarek: Yes, it is on linked input-proxy page, but indeed it may make sense to replicate it here too.

andrewdavidwong: @Yethal: Do you want to copy that warning over?

Yethal: @andrewdavidwong 
Following warning has been added:
In order to use a USB mouse, you must first attach it to a USB qube, then give that qube permission to pass mouse input to dom0. Note that allowing mouse access from a USB qube gives it great power. In short:

    It will be able to change any global and per-domain setting available within GUI
    It will be able to attach untrusted devices to domains that should not have device access
    It will be able to enable networking in network-disconnected domains
    It will be able to initiate file transfer between domains. (For more details, see here.)

If you are sure you wish to proceed, then you must edit the qubes.InputMouse policy file in dom0, which is located here:"
Is this good enough or should I elaborate further?

marmarek: While some warning about USB mouse would be useful too, this one looks like more serious than USB keyboard, which isn't the case. As you noticed, USB mouse cannot unlock screenlocker - because we don't have on-screen keyboard and because USB mouse cannot capture the passphrase, so the risk is much smaller. If anything wrong happens, user will most likely notice it.
So, in this case, I'm in favor of just linking to one place with such warning, instead of replicating it everywhere. So, the question is where should be this place - here, or in https://github.com/QubesOS/qubes-app-linux-input-proxy/blob/master/README.md?

jpouellet:
Is there any way you could do this without the massive add & remove? Maybe by splitting the part where you move a section to the top and modify it into two commits or something?

jpouellet:
Either way, thanks for cleaning this page up @Yethal.

I agree that it is poorly structured and confusing without something like this.

Yethal:
I don't know why GitHub treats it like a giant add&remove considering the fact that most of the stuff was moved around. If it was possible to diff individual sections instead of the entire page it would be much easier to compare changes.

jpouellet:
A local `git diff --minimal ... ...` produces the same thing unfortunately.

I think this new order of contents is an improvement and agree with the idea, but I have not carefully reviewed the giant blocks to be sure no silliness has slipped in ;)

common-tasks/usb.md
Yethal: Changed the warning so that it links to the security notice on the input-proxy page. Updated version available here:
https://github.com/Yethal/qubes-doc/blob/Yethal-USB-rework/common-tasks/usb.md

jpouellet: > Moreover, due to lack of onscreen keyboard the mouse would not be able to input terminal commands (unless it would copy them from a character map which I'm not sure Qubes even has).

It is definitely entirely possible to take over a computer using only the mouse. You can open a terminal and paste into it, and there exist plenty of easily accessible UIs with known text from which to copy from.

Yethal: That would require the user to not lock the screen and leave the PC unattended which is a giant no-no.

jpouellet: Not necessarily. You could still take over the machine with the user is present. Detect some inactivity for N-minutes (meaning user is probably distracted talking to a friend or something), or just not care about being noticed and take it over anyway (maybe that is enough to satisfy the adversary).

andrewdavidwong: I agree that we should not underestimate the creativity of a clever adversary who has full control of the mouse.

andrewdavidwong: > So, in this case, I'm in favor of just linking to one place with such warning, instead of replicating it everywhere. So, the question is where should be this place - here, or in https://github.com/QubesOS/qubes-app-linux-input-proxy/blob/master/README.md?

I think it would be better to have it in the docs (so that it will be included if we ever have offline docs, or if someone clones qubes-doc for their own use).

In the interest of time, I'll merge this PR and update the warnings in follow-up commits.

andrewdavidwong:
Thank you for all of your hard work on this, @Yethal!

PR #317, opened 2017-03-16
Quality over quantity

jpouellet:
IMHO: The bar must be set somewhere... We are not in the business of poorly documenting arbitrary linux defaults.

This information is also shown when you mouse over the disk in the default file manager:
run-media-user


andrewdavidwong:
[Why was that page added to begin with?](https://en.wikipedia.org/wiki/Wikipedia:Chesterton%27s_fence)

Perhaps it should be moved to the User FAQ rather than removed entirely.

jpouellet:
> Why was that page added to begin with?

A fair criticism. I didn't know there was even a name for that, haha.

Introduced with these commits:
```
commit 1cb3fec625658fe2aa873ba022160cc85a9616a0
Author: IX4 SVS <ix4svs@gmail.com>
Date:   Tue Oct 29 11:57:42 2013 +0000

    ExternalDeviceMountPoint changed

commit 7b4cecf6274f75867618435befb6db351fbdeb99
Author: IX4 SVS <ix4svs@gmail.com>
Date:   Tue Oct 29 11:56:18 2013 +0000

    ExternalDeviceMountPoint changed
```
whose time and author coincides with this thread: https://groups.google.com/d/topic/qubes-users/nQOP98Lc0E8/discussion

> Perhaps it should be moved to the User FAQ rather than removed entirely.

Done.

andrewdavidwong:
> I'm not convinced it belongs in the Frequently Asked Questions since doesn't appear to actually be one, and also isn't Qubes-specific, but if you disagree see #321.

Agreed. Thanks for investigating!

andrewdavidwong:
Superseded by #317.

PR #321, opened 2017-03-17
Add note about mount point to FAQ

jpouellet:
See https://github.com/QubesOS/qubes-doc/pull/317

I don't think this actually belongs here, but I defer to your judgement.


jpouellet:
This has also [caused confusion](This note has even [caused confusion](https://groups.google.com/d/msg/qubes-users/iK3mK9QhCuw/AenF6GqvDQAJ).

Perhaps it should be a slightly larger section on "Where are devices mounted?" with this note, and linking people to https://www.qubes-os.org/doc/usb/ (which explains /dev/xvd* as well).

andrewdavidwong:
> Perhaps it should be a slightly larger section on "Where are devices mounted?" with this note, and linking people to https://www.qubes-os.org/doc/usb/ (which explains /dev/xvd* as well).

Agreed. Please make it so.

andrewdavidwong:
And let's make sure to add redirects from all of the following so that existing external (e.g. mailing list) links won't 404:

- /doc/external-device-mount-point/
- /en/doc/external-device-mount-point/
- /doc/ExternalDeviceMountPoint/
- /wiki/ExternalDeviceMountPoint/

jpouellet:
> Agreed. Please make it so.

Done.

> - /doc/external-device-mount-point/
> - /en/doc/external-device-mount-point/
> - /doc/ExternalDeviceMountPoint/
> - /wiki/ExternalDeviceMountPoint/

I think it makes more sense for these to redirect to https://www.qubes-os.org/doc/usb/. It's likely more relevant to what you're looking for than dumping you in the general user FAQ and expecting you to then read it to find your specific issue, which will probably end up in you following the link to the USB docs page anyway.

andrewdavidwong:
> I think it makes more sense for these to redirect to https://www.qubes-os.org/doc/usb/ instead. It's likely more relevant to what you're looking for than dumping you in the general user FAQ and expecting you to then read it to find your specific issue, which will probably end up in you following the link to the USB docs page anyway.

Sure, that's fine. The most important thing is to make sure they're redirected *somewhere* so that we don't break preexisting links.

PR #323, opened 2017-03-18
Update devel-books.md

inktrap:
Fixed link.


jpouellet:
Oops. That's my fault. Thanks!

PR #338, opened 2017-04-02
Rewrite parts of usability-ux.md

stubbybubby:
This proposed file changes probably includes some content changes, but ones I hope you'll immediately agree with because the result is a more accurate article.

Among other things,
* Removed unnecessarily prudent logical connection phrases like "thus," that were ironically included in a document mostly about UI simplification. Just replaced them with "because" and "as a result" where I could or I wasn't re-using them too often.
* Moved some sentences or clarifications to their proper sections, e.g. moved the "Despite this rule, some acronyms..." line to the Avoid Acronyms section.
* Removed some sentences (like "Additionally, security focused software has the responsibility of providing safety to the user and their data") that, while true, without proper context did not seem relevant to the article.
* Removed some phrases that repeated known contextual information like "of usability and experience" in "if you are interested in furthering your understanding of usability and experience" or "which are shown to a user" in "notifications which are shown to a user".
* Introduced some clauses or phrases that clarified information, like adding "of those mentioned" to "All three desktop environments have their own human interface guidelines,"
* Merged the Avoid Repetitive Tasks and Minimize Repetitive Steps sections.
* Rewrote some sentences that were incomplete or didn't make sense into what I assumed the writer intended to say.
* Inserted possessive's' when they were necessary, and turned bigger sentences into smaller ones using possessives when I could ("notification messages presented to user" --> "user's notification messages")
* Grammar edits
* Compartmentalized run-on sentences.

This was a pretty big edit, so you'll probably disagree with some of the changes. Just explain what you want to be re-done or where my edits changes an important meaning/formed an inaccuracy and I will make another commit.

### DIFF
--- a/basics_dev/usability-ux.md
+++ b/basics_dev/usability-ux.md
@@ -7,26 +7,26 @@ permalink: /doc/usability-ux/
 Usability & UX
 ==============
 
-Software that is too complicated to use, is often unused. Thus, usability and user experience of Qubes OS is an utmost priority for us, as we want as many people as possible to benefit from the unique security properties of Qubes OS!
+Software that is too complicated to use, is often unused. Because we want as many people as possible to benefit from its unique security properties, the usability and user experience of Qubes OS is an utmost priority!
 
-We ask anyone developing for Qubes OS to please read through this guide to better understand the user experience we strive to achieve. Also, please review [our style guide](/doc/style-guide/) for other design related information.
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
Thanks. I'll accept the PR as-is, even though there are minor issues that I don't have time to enumerate. The status of this document is somewhat unclear. It's very opinionated, but I don't think it ever went through a vetting process or was approved by the devs. I doubt the devs feel bound by it or even look at it when making design decisions. Anyway, I appreciate your efforts and contributions.

PR #339, opened 2017-04-02
Put recent security bulletins at top

stubbybubby:
Reorganized the page so that the recent security bulletins were at the top instead of the bottom and readers didn't check for recent vulnerabilities by scrolling to the bottom of the page. Also made the security bulletins inside each year go from more recent->older.


andrewdavidwong:
There's a long history of people in the Qubes Project disagreeing and bikeshedding over chronological sort order. In the interest of saving time, we won't change change this.

PR #342, opened 2017-04-07
update example how to upgrade the dom0 kernel in Qubes R3.2 to 4.8.12

h01ger:
Signed-off-by: Holger Levsen <holger@layer-acht.org>

### DIFF
--- a/common-tasks/software-update-dom0.md
+++ b/common-tasks/software-update-dom0.md
@@ -117,10 +117,10 @@ To enable or disable any of these repos permanently, change the corresponding bo
 
 ### Kernel Upgrade ###
 
-Install newer kernel. The following example installs kernel 3.19 and was tested on Qubes R3 RC1.
+Install newer kernel. The following example installs kernel 4.8.12 and was tested on Qubes R3.2.
 
 ~~~
-sudo qubes-dom0-update kernel-3.19*
+sudo qubes-dom0-update --enablerepo=qubes-dom0-unstable kernel
 ~~~
 
 Rebuild grub config.


andrewdavidwong:
Won't this quickly become inaccurate once the kernel received in this manner is no longer 4.8.12?

tasket:
In the section above, it already shows how to enable unstable repo. So I'm not sure what the point is.

h01ger:
On Fri, Apr 07, 2017 at 08:19:10PM -0700, Andrew David Wong wrote:
> Won't this quickly become inaccurate once the kernel received in this manner is no longer 4.8.12?

it currently speaks of R3.0 RC3 and (the example is tied to) kernel 3.19.*…

(while my patch shows an example which is timeless, just the explaination
refers to a specific kernel version.)


-- 
cheers,
	Holger

h01ger:
On Fri, Apr 07, 2017 at 09:35:59PM -0700, tasket wrote:
> In the section above, it already shows how to enable unstable repo. So I'm not sure what the point is.

the example above is useless, as it shows how to temporarily enable the repo 
without doing anything with it. (and then its temporarily, so the next
install command will not use it…)


-- 
cheers,
	Holger

andrewdavidwong:
> (while my patch shows an example which is timeless, just the explaination
refers to a specific kernel version.)

Please make the explanation more generic. Even though it's obvious to us, we're bound to have someone X years from now ask how to get a newer kernel and failing to understand that the example command is timeless (since nothing here explicitly says or explains that it is timeless).

andrewdavidwong:
Also, strictly speaking, this sentence will become false before long:

> The following example installs kernel 4.8.12

h01ger:
On Sat, Apr 08, 2017 at 03:31:02AM -0700, Andrew David Wong wrote:
> > (while my patch shows an example which is timeless, just the explaination
> refers to a specific kernel version.)
> 
> Please make the explanation more generic. Even though it's obvious to us, we're bound to have someone X years from now ask how to get a newer kernel and failing to understand that the example command is timeless (since nothing here explicitly says or explains that it is timeless).

sigh. perfect is the enemy of good. i've been waiting for a newer kernel than
4.8 since some time and I might come back looking at the docs once I see it.

Also kernel-qubes-vm 4.8 is not easily installable atm (see the thread on
qubes-user) so following your perfect philosophy updating the docs should
wait until its clear how to do this properly…

(There are really many places in the docs refering to Qubes 2.0 or 3.0 or
some RC version, so I dont get why you object updating these to be more current.
Surely updating them and making them more timeless would be even better, but
just updating them to be current is also better than the status quo. I think
you should be more encouraging and accepting to small improvements here, as
this will also lead to more and better improvements over time. As it's now,
I rather not spend little time to improve qubes-doc.git if this results in being
asked to spend a lot more time doing so.)


-- 
cheers,
	Holger

andrewdavidwong:
> sigh. perfect is the enemy of good. i've been waiting for a newer kernel than
> 4.8 since some time and I might come back looking at the docs once I see it.
> 
> Also kernel-qubes-vm 4.8 is not easily installable atm (see the thread on
> qubes-user) so following your perfect philosophy updating the docs should
> wait until its clear how to do this properly…
> 
> (There are really many places in the docs refering to Qubes 2.0 or 3.0 or
> some RC version, so I dont get why you object updating these to be more current.
> Surely updating them and making them more timeless would be even better, but
> just updating them to be current is also better than the status quo. I think
> you should be more encouraging and accepting to small improvements here, as
> this will also lead to more and better improvements over time. As it's now,
> I rather not spend little time to improve qubes-doc.git if this results in being
> asked to spend a lot more time doing so.)

That's not the issue. In fact, if you look at our doc guidelines, we explicitly encourage referencing versions like "Qubes 3.2." The problem here is accuracy. You need to either say something like, "As this is being written, the following command will install kernel 4.8.12," or, "The following command will install the most recent kernel available from repo X." The way you have written it, this sentence will almost certainly be false before long.

andrewdavidwong:
> I think
you should be more encouraging and accepting to small improvements here, as
this will also lead to more and better improvements over time. As it's now,
I rather not spend little time to improve qubes-doc.git if this results in being
asked to spend a lot more time doing so.

This goes both ways. I think you should be more receptive to feedback and requests if you're submitting a PR against the docs, especially if the change you're making is more of a lateral move than a major improvement. As the doc maintainer, I'd rather not spend a lot of time arguing and bikeshedding about questionable PRs when I could more quickly fix it myself, but I'm willing to do so because open source collaboration is important, and discussion is required so that we can understand each other.

andrewdavidwong:
Also, I encourage you to take a look at the many, many PRs I've merged without discussion that were "good but not perfect." In fact, that's how things *usually* go. I only ask questions and request changes if I think they're merited. I keep 99.99% of my nitpicks to myself.

PR #343, opened 2017-04-11
Update pentesting.md

Arr0way:
Improved the penetration testing definition :)


andrewdavidwong:
We generally prefer to use more neutral sources (like Wikipedia), rather than linking to specific websites about which we know nothing.

PR #361, opened 2017-04-17
Insignificantly edit martus.md

stubbybubby:
Modified the list heading to be grammatically consistent.

### DIFF
--- a/privacy/martus.md
+++ b/privacy/martus.md
@@ -10,7 +10,7 @@ Martus
 [Martus] is a free, open source, secure information collection and management tool 
 that empowers rights activists to be stronger in their fight against injustice and abuse.
 
-To install Martus in a qube, you must:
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
Is the existing list heading really grammatically inconsistent?

stubbybubby:
Actually this would be the correct phrasing:

stubbybubby:
http://www.whitesmoke.com/punctuation-colon.html

stubbybubby:
You can leave it as is if you don't care about following those rules, but I would still change it to the "To install martus" phrasing if I were you. I just like it more.

andrewdavidwong:
Ok, merged the first version: https://github.com/QubesOS/qubes-doc/commit/36eb4058e4c048603d199edcdf2589e9661e3179

PR #376, opened 2017-04-27
Point to more detailed and recent instructions.

ubestemt:
Of course the Archlinux template building instructions don't explain how to build Qubes from scratch, but they certainly cover the relevant parts a lot clearer than this how-to.


marmarek:
Before promoting Archlinux building instruction everywhere, there is one _very important_ step missing there: verifying qubes-builder integrity.

ubestemt:
Will #378 do?

PR #416, opened 2017-05-08
Add clarity, fix spelling and grammar, and correct style in the salt.…

SamuelHentschel:
As I was reading the Salt Management Stack documentation, I rewrote some of it.  In doing this I believe I have added to the clarity of the documentation.  In addition, I have fixed capitalization and indentation/newlines for consistency (not only to the reader, but in regards to the style guide).

### DIFF
--- a/configuration/salt.md
+++ b/configuration/salt.md
@@ -381,19 +381,18 @@ VM which provides network to the given VM
 The output for each VM is logged in `/var/log/qubes/mgmt-VM_NAME.log`.
 
 If the log does not contain useful information:
-1. Stop `qubesctl` using Ctrl-z
-2. Run `sudo qubesctl --skip-dom0 --target=VM_NAME state.highstate`
-3. When your VM is being started (yellow) press Ctrl-z on qubesctl.
-4. Open terminal in disp-mgmt-VM_NAME.
-5. Look at /etc/qubes-rpc/qubes.SaltLinuxVM - this is what is 
+1. Run `sudo qubesctl --skip-dom0 --target=VM_NAME state.highstate`
+2. When your VM is being started (yellow) press Ctrl-z on qubesctl.
+3. Open terminal in disp-mgmt-VM_NAME.
+4. Look at /etc/qubes-rpc/qubes.SaltLinuxVM - this is what is 
    executed in the management VM.
-6. Get the last two lines:
+5. Get the last two lines:
 
     $ export PATH="/usr/lib/qubes-vm-connector/ssh-wrapper:$PATH"
     $ salt-ssh "$target_vm" $salt_command
 
   Adjust $target_vm (VM_NAME) and $salt_command (state.highstate).
-7. Execute them, fix problems, repeat.
+6. Execute them, fix problems, repeat.
 
 ## Known Pitfalls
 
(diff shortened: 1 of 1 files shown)


configuration/salt.md
```diff
@@ -104,132 +114,137 @@ You also can [order the execution][salt-doc-states-order] of your states:
...
-Because we use some code that is not found in older versions of salt, there is
-a tool called `qubesctl` that should be run instead of `salt-call --local`. It
-accepts all arguments of the vanilla tool.
-
+Because we use some code that is not found in older versions of Salt, there is
+a tool called `qubesctl` that should be run instead of `salt-call --local`.
+It accepts all the same arguments of the vanilla tool.
 
-## Configuring system inside of VMs
+## Configuring Systems from Dom0
```
marmarek: IMO this isn't clear it's about VMs...

configuration/salt.md
```diff
@@ -321,91 +343,88 @@ domains this way.
...
 
-You need to:
-1. run `sudo qubesctl --skip-dom0 --target=VM_NAME state.highstate`
-2. When your vm is being started (yellow) press Ctrl-Z on qubesctl.
-3. Open terminal in disp-mgmt-VM_NAME.
-4. Look at /etc/qubes-rpc/qubes.SaltLinuxVM - this is what is 
-   executed in the management vm.
-5. Get the last two lines:
+If the log does not contain useful information:
+1. Stop `qubesctl` using Ctrl-z
```
marmarek: This is actually covered by step 3.

configuration/salt.md
```diff
@@ -321,91 +343,88 @@ domains this way.
...
+The fedora-24-minimal package is missing the `sudo` package.
 You can install it via:
 
-    qvm-run -p vmname 'dnf install -y sudo'
+    $ qvm-run -p vmname 'dnf install -y sudo'
     
-The `-p` is will cause the execution to wait until the package is installed.
-This is important when using a state with `cmd.run`.
+The `-p` will cause the execution to wait until the package is installed.
+Having `sudo` is important when using a state with `cmd.run`.
```
marmarek: I'm not sure what I meant here initially, but surely not "having `sudo`". If anything, "the `-p`". Also there should be `-u root` in addition to `-p`...

marmarek: I think it may be clarified even further by replacing `vmname` with `fedora-24-minimal`.

configuration/salt.md
SamuelHentschel: This probably is true.  When reading the documentation, I was confused between this section and the heading "Example of vm system configuration" on line 265 (which I changed to "Example of Configuring Systems from Dom0" on line 290).

To me it seems like both of these sections describe configuring VMs from Dom0.  Are they in fact two completely different sections entirely and I just misunderstood?

configuration/salt.md
SamuelHentschel: Ok, so I misunderstood this.  It originally said, "If the log does not contain useful information, you can stop `qubesctl` by 
-pressing `ctrl+z`." and I wanted to make sure that it was just a note rather than a step.  I left it in there just in case.  I'll delete this and move the steps down one.

configuration/salt.md
SamuelHentschel: Ok, so I will replace the "having sudo" to "having the -p flag"? 
I will also add the "-u root" option.

As for the fedora-24-minimal instead of the vmname, I get what you are saying but should it be <fedora-24-minimal-template> instead?  For instance, I have a bunch of templates based off of the fedora-24-minimal base template.  Or are you thinking this would be done at the beginning before they clone off a bunch of templates?

configuration/salt.md
SamuelHentschel: What could I replace that line with that is more descriptive?

SamuelHentschel:
ping mamarek

configuration/salt.md
marmarek: Maybe "Configuring a VM's System from dom0"? And very similar one for example - like "Example of configuring a VM's System from dom0"

SamuelHentschel:
Ok Mamarek, I think I've cleared up all the issues.  Let me know if there is anything else, and I am excited to be included in the Qubes-Docs project!

PR #417, opened 2017-05-09
Add documentation for tweaking Tor Browser

ubestemt:


adrelanos:
Not a full review. Minor things only.

* `Tor`, not `TOR`.

> Also, if you use Whonix as NetVM the traffic will be routed through Tor anyway.

It's true, but I wouldn't recommend it since it breaks Tor Browser's by tab stream isolation since Tor Browser would no longer be directly talking to a Tor `SocksPort`.

> Side effect: Tor Browser is not able to open .onion links anymore.

(Very minor comment: Depending if the context still is `Also, if you use Whonix as NetVM the traffic will be routed through Tor anyway.`, `.onion` connections should still be functional. (When you are behind a Whonix-Gateway using it as DNS server, you can even run `nslookup some.onion`.)

> Using Tor Browser without Tor.

This is great! Using a more hardened (compile options) Firefox (hardening will get better and better with sandboxing in future and whatnot) is great! Would be awesome if data could be persistent for more clearnet like use. Since you're up to it... Would you like to work on the following?

`TODO research and document - How to use Tor Browser for security not anonymity? How to use TBB using clearnet?`
https://forums.whonix.org/t/todo-research-and-document-how-to-use-tor-browser-for-security-not-anonymity-how-to-use-tbb-using-clearnet

-----

Full disclosure:
I am a maintainer of Whonix.

ubestemt:
@adrelanos Tor, not TOR: That is a quote from Super User. The link below leads to it, but I should make it clearer that it's actually a quotation.

Tor Browser without Tor: Thanks for your comments remarks! I was not aware of this. How does TBB's built-in Tor proxy interact with the Whonix gateway to prevent a Tor-over-Tor scenario? I cannot find any documentation in the Whonix wiki or elsewhere (but then I don't really know what to look for either).

Does this mean that there is no disadvantage at all to use a vanilla Tor Browser in a Torified VM?

When you use Tor Browser without Tor `.onion` addresses are disabled in the browser itself. I tried to figure out why once but didn't get anywhere.

Anyway, I will edit the instructions and recommend keeping a separate Tor Browser for clearnet use.

adrelanos:
> How does TBB's built-in Tor proxy interact with the Whonix gateway to prevent a Tor-over-Tor scenario?

In Whonix-Workstation, package https://github.com/Whonix/anon-ws-disable-stacked-tor prevents Tor over Tor.

> Does this mean that there is no disadvantage at all to use a vanilla Tor Browser in a Torified VM?

There is a disadvantage. In a torified Debian VM, vanilla Tor Browser would be using Tor over Tor.

> When you use Tor Browser without Tor .onion addresses are disabled in the browser itself. I tried to figure out why once but didn't get anywhere.

That is pref `network.dns.blockDotOnion`.

ubestemt:
Tor Browser would not work with `anon-ws-disable-stacked-tor` in a normal debian-based VM: https://github.com/Whonix/anon-ws-disable-stacked-tor/issues/1. Do you know/could you think of any relatively simple way of running Tor Browser with tab stream isolation but without Tor over Tor in a non-Whonix VM?

I thought have a Tor Browser without Tor in a DispVM would be a near-perfect solution as you could switch between Tor and clearnet by simply changing the NetVM. But then you would have to do without tab stream isolation. All other implementations I can think seem worse.

The Tor Browser in whonix-ws *has* tab stream isolation, but using that VM without whonix-gw as sys-firewall probably is not a good idea (if it would work at all), even if you have a separate Tor Browser without Tor. Running Tor Browser (with Tor) in a clearnet VM is not an option as it makes you are a lot more vulnerable to exploits in Tor Browser.

I guess we have to wait for Qubes to support multiple DispVM templates. Then you can have a whonix-gw-based DispVMs for Tor Browser (with Tor) and another template with Tor Browser (without Tor) for clearnet browsing.

adrelanos:
Right.

> Tor Browser would not work with anon-ws-disable-stacked-tor in a normal debian-based VM: Whonix/anon-ws-disable-stacked-tor#1

Replied in that ticket.

> Do you know/could you think of any relatively simple way of running Tor Browser with tab stream isolation but without Tor over Tor in a non-Whonix VM?

Not really. Make anon-ws-disable-stacked-tor work.

andrewdavidwong:
@adrelanos, would you be willing to do a full review of this PR (if you haven't done so already)?

adrelanos:
Some feedback here starting from `Could anyone help with the following please?`:
https://forums.whonix.org/t/long-wiki-edits-thread/3477/325

Tor Browser as default browser outside of Whonix:
That's not simple. A package for that purpose has been developed https://github.com/Whonix/tb-default-browser. I don't have time to maintain it outside of Whonix but contributors are very welcome, to maintain it and perhaps upload it to packages.debian.org. Meaning, I don't want to review any manual attempt on doing it since that would be a time sink.

Tor Browser in DispVM:
Not worth working on before Qubes R4 since so much changed there wrt DispVMs.

Tor Browser without Tor:
We have this in Whonix wiki, it was contributed by @ubestemt as well.
https://www.whonix.org/wiki/Tor_Browser_without_Tor

I don't think this is ready, and I am not sure @ubestemt is still around. But I also think I am not a suitable maintainer / reviewer for Tor Browser in Qubes outside of Whonix since this can get a time sunk.

andrewdavidwong:
@adrelanos: Ok, that's very reasonable. Thank you for your time and feedback!

@ubestemt, I apologize for the long delay since we've had any activity on this. In light of @adrelanos's comments above, I don't think we want to support tweaking Tor Browser outside of Whonix. We'd prefer for there to be one canonical way to use Tor Browser in Qubes, namely through Whonix. If you feel strongly about this and would like us to reconsider this PR, please leave a comment. We'll be happy to reopen and discuss it further. Thank you for your understanding.

PR #418, opened 2017-05-09
How to play multimedia-content

praschdorff:
This howto describes how to play multimedia content on Qubes OS.
It covers Spotify, Netflix and Amazon Prime under Qubes OS


phrabe:
I'd like to add Markdown Syntax and format this article correctly, but I don't no how to do so.
If someone can edit this and add header, lists I can continue to clean this up.
I tried to use markdown (https://daringfireball.net/projects/markdown/basics) but the formatting doesn't show up in the preview window

configuration/multimedia
```diff
@@ -0,0 +1,127 @@
...
+I could then start the native spotify client .. easy.
+
+
+Play Netflix
+============
+You can play Netflix by installing a more recent version of firefox or by installing Google Chroome.
+As Google Chrome will play video from Netflx and (!) Amazon, the suggestion is to install Google Chrome.
+If you don't use Amazon Prime, Firefox will work.
+
+Install Firefox
```
jpouellet: Eeek!

This installation procedure does not strongly verify the firefox being installed, and will become old & stale (read: vulnerable) over time due to not being updated automatically in the TemplateVM. I really don't think we should suggest this.

Are you sure this isn't available from some -testing debian repo or something like that?

/cc @unman

configuration/multimedia
```diff
@@ -0,0 +1,127 @@
...
+That's it, login into Netflix and enjoy.
+
+
+How to use Amazon Prime
+=======================
+Playing Video thorugh Amazon Prime via Firefox didn't work, with Google Chrome it works.
+Warning: You need to install Google Chrome, not Chromium.
+
+Install Google Chrome
+---------------------
```
jpouellet: Same thing. Are you sure there's no repo for this?

configuration/multimedia
```diff
@@ -0,0 +1,127 @@
...
+   In Debian Template VM:
+   apt-get install libdvdread4
+   
+3) Install libdvdcss for the decryption of CSS protected-DVDs.
+   Because of license restrictions in various countries this can't be done from the Debian repositories but needs to be installed manually.
+   (not that hard :-)).
+   Download the packaged .deb from: http://www.deb-multimedia.org/dists/stable/main/binary-amd64/package/libdvdcss2
+   In Debian Template VM:
+   wget http://www.deb-multimedia.org/pool/main/libd/libdvdcss/libdvdcss2_1.3.0-dmo1_amd64.deb
+   sudo dpkg -i libdvdcss2_1.3.0-dmo1_amd64.deb
```
jpouellet: Same here, but maybe this one is actually necessary?

jpouellet:
When adding new pages, don't forget to link to them in the index: https://github.com/QubesOS/qubes-doc/blob/master/doc.md

jpouellet:
Oh, I see this is your first contribution. Thanks for the help!

Please don't be discouraged by my comments above, they're just some minor tweaks. Overall it looks good, and I'm sure your notes will save many people lots of time :)

phrabe:
Thank you for the valuable feedback, I don't feel discouraged ;-) the opposite, I thought the document will stay here forever without going into the main repository.
Honestly I thought about adding repositories instead of downloading an install the files, but I was unsure if people like to add repositories to their template VM as this might result in less control what will be installed when the update the template.
As Netflix only seems to work with Chromium, we might skip the installation of Firefox.
AFAIK there is no need for a Google Repository as Google Chrome will take care of this:

(...) Google’s Linux software packages will automatically configure the repository settings necessary to keep your Google Linux applications up-to-date. (...)
Link: https://www.google.com/linuxrepositories/

Regarding the other files I'll try to figure out if those can be installed by adding a repository.

Thanks again for your feedback.

configuration/multimedia
phrabe: Firefox could be installed from a package repository.
Link: https://mozilla.debian.net

To Install Firefox ESR v53.x:
1) Add new repository to the sources:
deb http://mozilla.debian.net/ jessie-backports firefox-esr

2) update repository and install Firefox:
$ apt-get update
$ apt-get install -t jessie-backports firefox-esr

I'll try those steps and if this will work, I'll update the documentation including the correct steps.

jpouellet:
> but I was unsure if people like to add repositories to their template VM as this might result in less control what will be installed when the update the template.

I see. I think this is a valid concern too.

I believe I've largely mitigated it on my own system by having separate "all the crap" templates for things which don't matter (like consuming media) and very minimal templates where I can for things which are security sensitive. I would recommend this practice.

jpouellet:
>> but I was unsure if people like to add repositories to their template VM as this might result in less control what will be installed when the update the template.
> 
> I see. I think this is a valid concern too.

However, the concerns of increased attack surfaces and authenticity and integrity guarantees of installed software are separate concerns IMO.

I think that package repos provide better authenticity and integrity guarantees for initial installs than grabbing download links from (hopefully) the upstream vendor and installing it like that.

The risk you are trading off here is that the people with control of the package repo may upload malicious things in the future (for example, don't upgrade coreutils from mozilla), vs. the risk of not keeping up with updated software. I favor using package repos because:
- If the people fully controlling the repo and signing keys are malicious, nothing prevents them from including an underhanded firefox (or whatever) itself
- Outdated browsers *are* exploitable

If you disagree, feel free to explain why. I try to keep an open mind.

phrabe:
Hello jpouellet,

I have rewritten my howto to setup a debian based template for Amazon/Netflix/Prime etc.
I have taken your feedback into account and instead of working with local package files I've choosen package repositories.

What do you think? I would like to get your feedback.
Also, what do I need to do, so that this content is linked in the main documentation and where should I place it exactly?

'# Become root
su -
 
# Install Qubes-USB-Proxy to be able to play DVD if you're using an external DVD drive
apt-get install qubes-usb-proxy
 
### Installation Spotify
apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys BBEBDCB318AD50EC6865090613B00F1FD2C19886
echo deb http://repository.spotify.com stable non-free | sudo tee /etc/apt/sources.list.d/spotify.list
apt-get update
apt-get install -y spotify-client
 
### Installation VLC
# Add Repository for libdvdcss
# http://www.videolan.org/developers/libdvdcss.html
wget -O - http://download.videolan.org/pub/debian/videolan-apt.asc | sudo apt-key add -
echo "deb http://download.videolan.org/pub/debian/stable/ /" >> /etc/apt/sources.list
echo "deb-src http://download.videolan.org/pub/debian/stable/ /" >> /etc/apt/sources.list
apt-get update
apt-get install -y libdvdcss2
apt-get install -y vlc
 
### Installation Google Chrome
wget -c wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
# Installation will fail because of missing dependencies. Install dependencies:
apt-get install -y fonts-liberation libappindicator1 libdbusmenu-glib4 libdbusmenu-gtk4 libindicator7 libxss1
dpkg -i google-chrome-stable_current_amd64.deb`

jpouellet:
> #Become root
> su -

You also use `... | sudo tee ...` and `... | sudo apt-key add -` patterns below. Keep one or the other - there's no point going from root to root ;)

> #Install Qubes-USB-Proxy to be able to play DVD if you're using an external DVD drive
> apt-get install qubes-usb-proxy

1. not all DVD players are attached over USB
2. regardless, it is potentially a safer idea to attach them as block devices (with qvm-block) instead of exposing usb complexity.

> ###Installation Spotify
> apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys BBEBDCB318AD50EC6865090613B00F1FD2C19886
> echo deb http://repository.spotify.com stable non-free | sudo tee /etc/apt/sources.list.d/spotify.list
> apt-get update
> apt-get install -y spotify-client

Haven't tested or verified the key, but looks about right. If spotify publishes their key on some official web page served over SSL, it would be nice to include a link to that too, in case they need to rotate their key and the fingerprint pinned here becomes obsolete.

> ###Installation VLC
> #Add Repository for libdvdcss
> #http://www.videolan.org/developers/libdvdcss.html
> wget -O - http://download.videolan.org/pub/debian/videolan-apt.asc | sudo apt-key add -

^ This line is a cause for concern. Is the key distributed some other (more secure) way?

> echo "deb http://download.videolan.org/pub/debian/stable/ /" >> /etc/apt/sources.list
> echo "deb-src http://download.videolan.org/pub/debian/stable/ /" >> /etc/apt/sources.list
> apt-get update
> apt-get install -y libdvdcss2
> apt-get install -y vlc

It may also be worth mentioning that this stuff should be done either in a TemplateVM or StandaloneVM, otherwise (in an AppVM) it will be lost on reboot, possibly with a link to a more in-depth explanation of why somewhere already in the docs.

jpouellet:
Instead of trying to provide documentation on how to bootstrap trust in 3rd party repos, it might be a better idea to root the trust in Qubes itself by distributing repo definitions (disabled by default) and pre-verified keys (not trusted for anything by default) for them already, so that the user has less overall work to verify things. We already do this for google chrome, adobe, the rpmfusion repos, etc.

If you're interested in this route, check out what's currently done here: https://github.com/QubesOS/qubes-builder-fedora/blob/master/template_scripts/add_3rd_party_software.sh

phrabe:
Hello, 

thanks for your valuable feedback.

> You also use ... | sudo tee ... and ... | sudo apt-key add - patterns below

true, removed - this was by mistake, before switching to root I was working with sudo, but I run into problems piping echo to another command etc.
switching temporary to root is just simpler-

> it is potentially a safer idea to attach them as block devices

true, I forgott this option and will include qvm-block -a in the docu as this will allow to use internal and external  DVD drives.

Regarding the GPG-keys I don't get why companies/organizations are not publishing the GPG fingerprint when they use self signed certificates. I've asked for a link to a page referencing to the GPG keys for spotify and also VLC in their IRC-channels - but didn't get an answer so far.

> wget -O - http://download.videolan.org/pub/debian/videolan-apt.asc | sudo apt-key add -
> This line is a cause for concern.
> Is the key distributed some other (more secure) way?

Sure, I have switched the key retrivial with wget against:
apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 8F0845FE77B16294429A79346BCA5E4DB84288D9

> It may also be worth mentioning that this stuff should be done either in a TemplateVM or StandaloneVM, 

Sure, but I would like to have the technical description right so far and then add more information.
I just don't want to rewrite the howto several times :-)

Here the total setup routine which will built a debian based multimedia template based on the default debian 8 template:

----8<-----

#Become root
su -

###Installation Spotify
apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys BBEBDCB318AD50EC6865090613B00F1FD2C19886
#TODO where to find the GPG Fingerprint for Spotify??
#Public-Key: http://keyserver.ubuntu.com/pks/lookup?op=vindex&search=0xD2C19886&fingerprint=on
#Fingerprint: BBEB DCB3 18AD 50EC 6865  0906 13B0 0F1F D2C1 9886 
echo deb http://repository.spotify.com stable non-free | tee /etc/apt/sources.list.d/spotify.list
apt-get update
apt-get install -y spotify-client

###Installation VLC
#Add Repository for libdvdcss
#http://www.videolan.org/developers/libdvdcss.html
#TODO where to find the GPG Fingerprint for VLC??
#Public-Key: http://keyserver.ubuntu.com/pks/lookup?op=vindex&search=0xB84288D9&fingerprint=on
#Fingerprint: 8F08 45FE 77B1 6294 429A  7934 6BCA 5E4D B842 88D9  
wget -O - http://download.videolan.org/pub/debian/videolan-apt.asc | apt-key add -
echo "deb http://download.videolan.org/pub/debian/stable/ /" >> /etc/apt/sources.list
echo "deb-src http://download.videolan.org/pub/debian/stable/ /" >> /etc/apt/sources.list
apt-get update
apt-get install -y libdvdcss2
apt-get install -y vlc

###Installation Google Chrome
#Howto verify the debian repository?
#https://www.google.com/linuxrepositories/
apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 8F0845FE77B16294429A79346BCA5E4DB84288D9
#Install dependencies for Google Chrome (if not installation will fail)
apt-get install -y fonts-liberation libappindicator1 libdbusmenu-glib4 libdbusmenu-gtk4 libindicator7 libxss1
dpkg -i google-chrome-stable_current_amd64.deb
rm google-chrome-stable_current_amd64.deb 

----8<-----

If this is ok from the technical point, I can rewrite my old howto.

Happy to hear your feedback.

- PhR

phrabe:
> Instead of trying to provide documentation on how to bootstrap trust in 3rd party repos, it might be a better idea to root the trust in Qubes itself by distributing repo definitions (disabled by default) and pre-verified keys (not trusted for anything by default) for them already, so that the user has less overall work to verify things. We already do this for google chrome, adobe, the rpmfusion repos, etc.
> If you're interested in this route, check out what's currently done here: https://github.com/QubesOS/qubes-builder-fedora/blob/master/template_scripts/add_3rd_party_software.sh

I've looked at the link, but didn't understand what this is about, unfortunately I am still a Qubes OS newbie ;-)

- PhR

jpouellet:
Spotify & VLC look good to me, the chrome example does not.

IMO we should avoid enumerating specific dependencies in the documentation, as these may change over time. I suspect a better solution would be to use `apt-get -f install` to resolve them automatically.

phrabe:
Hello,

I'm getting frustrated as it seems I am not smart enough to work with GitHub.
I have rewritten the document in my local Notepad++ but when I try to copy it into the code-window, all line breaks get messed up.
I also downloaded "GitHub Desktop" and logged in with the same account I am using now, still I can't edit my own document.
Any idea what is wrong?

I have attached the plain text file, if you can update it and make it publish to the Qubes Doc repository this would be nice. I also don't know how to add my multimedia-section to the Qubes Doc Contents.
Do I need to edit another page?

[multimedia.txt](https://github.com/QubesOS/qubes-doc/files/1133956/multimedia.txt)

jpouellet:
> I'm getting frustrated as it seems I am not smart enough to work with GitHub.
> I have rewritten the document in my local Notepad++ but when I try to copy it into the code-window, all line breaks get messed up. 

Are you on windows by any chance? Sounds like you fell victim to the classic windows line endings problem (line endings being carriage-return line-feed instead of just line-feed). This is indeed what happened in the file you attached. This is a very common problem, google around if you're more curious.

If you are on a unix-like machine (such as Qubes), try using the dos2unix command to fix the file.

> I also downloaded "GitHub Desktop" and logged in with the same account I am using now, still I can't edit my own document.
> Any idea what is wrong?

By can't edit, do you mean some permissions error? Or the line ending issue? I do everything from the command line and rarely ever use windows, so I'm really not the best person to help with this unfortunately.

> I have attached the plain text file, if you can update it and make it publish to the Qubes Doc repository this would be nice.

According to my understanding of the way pull requests on github work, I believe the only way is for you to update your own branch (or to give me push access to your clone of the repo - but that's bad practice and I'd recommend not doing so).

> I also don't know how to add my multimedia-section to the Qubes Doc Contents.
> Do I need to edit another page?

Yes, this page: https://github.com/QubesOS/qubes-doc/blob/master/doc.md

praschdorff:
I have rewritten configuration/multimedia.md from scratch and also updated doc.md and hope to see the changes pushed into the master repository.

configuration/multimedia
```diff
@@ -1,136 +1,160 @@
...
 
-This short howto will explain how you can use Spotify, Netflix or Amazon Prime under Qubes OS.
-This steps have been tested by providing an own "multimedia" App VM which is based on the Qubes OS Debian 8 Template VM.
+Multimedia
+==========
 
-Listen to Spotify
-=================
-Install-Howto: https://www.spotify.com/de/download/linux/
+In order to choose Qubes OS as your primary OS it must be able to do all tasks, including playing multimedia content.
```
andrewdavidwong: This is obviously not true.

configuration/multimedia
```diff
@@ -1,136 +1,160 @@
...
+`qvm-clone debian-9 t-multimedia`
 
-1) download firefox from https://www.mozilla.org/en-US/firefox/new/?scene=2
-direct download link:
-https://download-installer.cdn.mozilla.net/pub/firefox/releases/53.0.2/linux-x86_64/en-US/firefox-53.0.2.tar.bz2
+As we need to install some package outside of the regular repositories you need to enable networking for the Template VM.
 
-2) unpack the downloaded file
-tar -xjf firefox-53.0.2.tar.bz2
+`qvm-prefs --set t-multimedia netvm sys-firewall`
```
andrewdavidwong: These steps shouldn't normally be necessary, since the TemplateVM should already have network access (at least to the updates proxy).

configuration/multimedia
```diff
@@ -1,136 +1,160 @@
...
+Install libdvdcss and VLC
+`apt-get install -y libdvdcss2 vlc` 
+
+
+ 
+Installation Google Chrome
+--------------------------
+
+To play Videos with Netflix, Amazon Prime & Co using Chrome is a good option as it has all needed codecs included.
+Hint: Using Chromium will not work for some reasons.
```
andrewdavidwong: Is it because Google Chrome has (had?) flash built-in, whereas Chromium doesn't?

configuration/multimedia
```diff
@@ -1,136 +1,160 @@
...
+
+
+Download Google Chrome package from the Google Debian repository
+
+`wget -c https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb`
+
+FIXME: Howto verify the debian package or better adding a Google Debian package repository?
```
andrewdavidwong: Not sure about Debian, but I think our Fedora TemplateVM already has the appropriate repo. You just have to enable it. (Similarly for RPMFusion, IIRC.)

configuration/multimedia
praschdorff: Ok, there might be different opinions on this topic. I would like to do all tasks that I can do with a computer within Qubes, which includes also multimedia.
Maybe we strip this out:

OLD: In order to choose Qubes OS as your primary OS it must be able to do all tasks, including playing multimedia content.

NEW: You can consolidate most of your media streaming tasks into one "multimedia" App-VM, which can based on an own multimedia template.

configuration/multimedia
praschdorff: This information is old, as the new part of the howto is based on Chrome.
I have attached the multimedia.md file, so that you can review it.
[multimedia.md.txt](https://github.com/QubesOS/qubes-doc/files/1735244/multimedia.md.txt)

configuration/multimedia
praschdorff: I was trying out Chromium before using Chrome, but it seems that Chromium is "missing" some protected DRM. 
See also here: https://ubuntuforums.org/showthread.php?t=2327128

configuration/multimedia
praschdorff: Ok, I found out that the signing keys can be downloaded here: https://www.google.com/linuxrepositories/

> Key Details
> Download: https://dl.google.com/linux/linux_signing_key.pub
> Key ID: Google, Inc. Linux Package Signing Key <linux-packages-keymaster@google.com>
> Fingerprint: 4CCA 1EAF 950C EE4A B839 76DC A040 830F 7FAC 5991
> Google, Inc. (Linux Package Signing Authority) <linux-packages-keymaster@google.com>
> Fingerprint: EB4C 1BFD 4F04 2F6D DDCC EC91 7721 F63B D38B 4796

to verify the downloaded package:

> On an RPM-based system (Fedora, SUSE, Mandriva, RHEL, etc.),
> download the key and then use rpm to install it.
> wget https://dl.google.com/linux/linux_signing_key.pub
> sudo rpm --import linux_signing_key.pub
> You can verify the key installation by running:
> rpm -qi gpg-pubkey-7fac5991-*
> To manually verify an RPM package, you can run the command:
> rpm --checksig -v packagename.rpm

configuration/multimedia
andrewdavidwong: Better, but I think "an own" might be a grammatical mistake.

configuration/multimedia
andrewdavidwong: Please include any new changes in the PR. We don't review separate attached files.

configuration/multimedia
andrewdavidwong: But this shouldn't be necessary if your TemplateVM already has the repo file and key.

marmarek: In Debian template we don't have this repo installed, so this step is needed. Instruction should include adding repository + key and only then installing something from there. Instead of installing deb downloaded from the network.

configuration/multimedia
```diff
@@ -0,0 +1,160 @@
...
+redirect_from:
+- /en/doc/multimedia/
+- /doc/Multimedia/
+- /wiki/Multimedia/
+---
+
+Multimedia
+==========
+
+In order to choose Qubes OS as your primary OS it must be able to do all tasks, including playing multimedia content.
```
marmarek: Same comment as from @andrewdavidwong. "all tasks" is much more than "all tasks you do". And even that isn't strictly necessary, because you may not necessary do all the tasks on the same ("primary") OS - for example you may have another one for gaming.

praschdorff:
The problem is that I can't find a way to install Google Chrome through a
repository.
I would like to add a repository to the Debian template, instead of
downloading the .deb package.


Am 20.02.2018 11:44 nachm. schrieb "Marek Marczykowski-Górecki" <
notifications@github.com>:

*@marmarek* commented on this pull request.
------------------------------

In configuration/multimedia
<https://github.com/QubesOS/qubes-doc/pull/418#discussion_r169485219>:

> +`apt-get install -y libdvdcss2 vlc`
+
+
+
+Installation Google Chrome
+--------------------------
+
+To play Videos with Netflix, Amazon Prime & Co using Chrome is a good
option as it has all needed codecs included.
+Hint: Using Chromium will not work for some reasons.
+
+
+Download Google Chrome package from the Google Debian repository
+
+`wget -c https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb`
<https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb>
+
+FIXME: Howto verify the debian package or better adding a Google
Debian package repository?

In Debian template we don't have this repo installed, so this step is
needed. Instruction should include adding repository + key and only then
installing something from there. Instead of installing deb downloaded from
the network.

—
You are receiving this because you modified the open/close state.
Reply to this email directly, view it on GitHub
<https://github.com/QubesOS/qubes-doc/pull/418#discussion_r169485219>, or mute
the thread

configuration/multimedia
```diff
@@ -0,0 +1,160 @@
...
+Start by cloning the default debian template in dom0.
+Hint:
+t-multimedia is just the template VM where we will install all packages.
+In the last step we will create an AppVM from this template.
+
+`qvm-clone debian-9 t-multimedia`
+
+As we need to install some package outside of the regular repositories you need to enable networking for the Template VM.
+
+`qvm-prefs --set t-multimedia netvm sys-firewall`
```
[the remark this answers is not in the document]
praschdorff: 'cat /etc/apt/apt.conf.d/01qubes-proxy' in the debian template VM shows me:

Acquire::http::Proxy "http://127.0.0.1:8082/";
Acquire::tor::proxy "http://127.0.0.1:8082/";

You wrote:
> The proxy address is constant, but different on 3.2 (http://10.137.255.254:8082/)
> and 4.0 (http://127.0.0.1:8081/)

I guess the last one 8081 is a typo?

I tried to run the following command in a new debian 9 based template VM:

apt-key adv --keyserver-options http-proxy=http://127.0.0.1:8082 --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 0DF731E45CE24F27EEEB1450EFDC8610341D9410

Results in:
Executing: /tmp/apt-key-gpghome.Zvh7HLoIYi/gpg.1.sh --keyserver-options http-proxy=http://127.0.0.1:8082 --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 0DF731E45CE24F27EEEB1450EFDC8610341D9410
gpg: keyserver receive failed: No keyserver available

this doesn't work.

configuration/multimedia
```diff
@@ -0,0 +1,160 @@
+---
```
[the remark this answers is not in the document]
praschdorff: done

marmarek:
The deb package for google chrome contains repository definition:
```
deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main
```
And this key:
```
pub   rsa4096 2016-04-12 [SC]
      EB4C1BFD4F042F6DDDCCEC917721F63BD38B4796
uid           Google Inc. (Linux Packages Signing Authority) <linux-packages-keymaster@google.com>
sub   rsa4096 2016-04-12 [S] [expires: 2019-04-12]
sub   rsa4096 2017-01-24 [S] [expires: 2020-01-24]
```
Which matches what is listed on https://www.google.com/linuxrepositories/

configuration/multimedia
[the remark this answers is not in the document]
marmarek: This is stupid... It does work on debian-8, but on debian-9 it looks like gpg "knows better". When running under strace, I see some messages from dirmngr written to stderr (not sure where it lands, if anywhere):
```
can't connect to '127.0.0.1': no IP address for host
error connecting to 'http://91.189.90.55:11371': Unknown host
```
And this is only after I've provided keyserver by IP. When given by name, dirmngr tries to resolve it through DNS itself, which obviously fails.

Maybe better to instruct people to download the key in other VM (link to http://keyserver.ubuntu.com/pks/lookup?op=get&search=0xEFDC8610341D9410 for example?) then copy to the template using either clipboard or file copy? But that's getting more complicated, without clear benefit for the security. So, for now maybe indeed better option is temporarily enable networking.

praschdorff:
@marmarek Thanks for the hint regarding the google repositories, we can now simplify the chrome installation:

1) From a normal AppVM: Download Googles Public Key via their website https://dl.google.com/linux/linux_signing_key.pub or via a keyserver: http://keys.gnupg.net/pks/lookup?search=0x7721F63BD38B4796&fingerprint=on&op=index

2) Copy the Pubkey to the Clipboard and then to a new file in the multimedia template VM
alternative: qvm-copy-to-vm the Public Key over to the Multimedia Template VM

3) apt-key add <Google Public Signing Key>

4) Add Googles Repository to apt:
sudo su
echo "deb http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list

5) Refresh package list
apt-get update

6) Install Chrome
apt-get install google-chrome-stable

... done.

praschdorff:
Rewritten the howto, so that the Template VM doesn't need full internet access.
Download of the needed repository signing keys by downloading those via another AppVM is also described.
All steps of the howto has been tested, including the launch of the application in a new AppVM based on this "multimedia template".
Please approve changes, so that the invested time pays of ;-)

configuration/multimedia.md
```diff
@@ -30,128 +30,146 @@ In the last step we will create an AppVM from this template.
...
+
 Installation of Spotify
 -----------------------
 
 Import GPG-Key for spotify
+As the template VM can't connect to internet you need to get the public key file from another AppVM and copy it to the template VM. The easiest way is to use the Qubes Clipboard to copy the keys from the AppVM where you get the key to the Template VM.
 
-`apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 0DF731E45CE24F27EEEB1450EFDC8610341D9410`
+In an AppVM which has Internet access:
+- Open http://keyserver.ubuntu.com:11371/pks/lookup?op=get&search=0xEFDC8610341D9410
```
marmarek: While you verify the fingerprint later, better use https anyway.

configuration/multimedia.md
```diff
@@ -30,128 +30,146 @@ In the last step we will create an AppVM from this template.
...
+Add the public key to the repository keyring
+`apt-key add spotify.pubkey`
 
-`apt-get update`
+Verify Fingerprint with
+
+`apt-key finger spotify`
+
+You can (and should) lookup the fingerprint on the keyserver:
+http://keyserver.ubuntu.com:11371/pks/lookup?op=vindex&search=0xEFDC8610341D9410&fingerprint=on
```
marmarek: Verification should be against some independent source. You can simply put the fingerprint here.
Also, it's a bit late for verification - you've already told apt to trust this key. Better call `gpg --with-fingerprint` on the file first, and `apt-key add` only if fingerprint matches.

configuration/multimedia.md
```diff
@@ -30,128 +30,146 @@ In the last step we will create an AppVM from this template.
...
 Installation of VLC
 -------------------
 
 To play DVDs you can install VLC with the needed Codecs
 
-Add Repository for libdvdcss
-(See also: http://www.videolan.org/developers/libdvdcss.html)
+Download the public key which signs the VLC package repositories
+In an AppVM which has Internet access:
+- Open http://keyserver.ubuntu.com/pks/lookup?op=get&search=0x6BCA5E4DB84288D9
```
marmarek: https

configuration/multimedia.md
```diff
@@ -30,128 +30,146 @@ In the last step we will create an AppVM from this template.
...
-`echo "deb-src http://download.videolan.org/pub/debian/stable/ /" >> /etc/apt/sources.list`
+Add the public key to the repository keyring
+`apt-key add vlc.pubkey`
+
+Verify Fingerprint with
+
+`apt-key finger VideoLAN`
+
+You can (and should) lookup the fingerprint on the keyserver:
+http://keyserver.ubuntu.com/pks/lookup?op=vindex&search=0xB84288D9&fingerprint=on
```
marmarek: Same as above.

configuration/multimedia.md
```diff
@@ -30,128 +30,146 @@ In the last step we will create an AppVM from this template.
...
- 
 Installation Google Chrome
 --------------------------
 
 To play Videos with Netflix, Amazon Prime & Co using Chrome is a good option as it has all needed codecs included.
 Hint: Using Chromium will not work for some reasons.
 
+Download the public key which signs the Google package repositories
+In an AppVM which has Internet access:
+- Open http://keyserver.ubuntu.com/pks/lookup?op=get&search=0x7721F63BD38B4796
```
marmarek: And here.

configuration/multimedia.md
```diff
@@ -30,128 +30,146 @@ In the last step we will create an AppVM from this template.
...
-> dependency problems - leaving unconfigured
-> (...)
-> Errors were encountered while processing:
-> google-chrome-stable
+`apt-key finger Google`
 
-Install the missing dependencies for Google Chrome
+You can (and should) lookup the fingerprint on the keyserver:
+http://keyserver.ubuntu.com/pks/lookup?op=vindex&search=0x7721F63BD38B4796&fingerprint=on
+or https://www.google.com/linuxrepositories/
```
marmarek: Here is better, because you provide alternative place for verification.
But still the order is wrong - first verify, then trust.

praschdorff: implemented all suggested changes and improved the handling of the public keys a bit providing info how to store them via xclip to a file.
If there are any further changes necessary feel free to do so, as I want to move on .
Great to see that the quality standard on qubes documentation is very high, even when I sometimes felt that I must give up - thanks for the constructive feedback.

configuration/multimedia.md
```diff
@@ -48,25 +56,36 @@ Import GPG-Key for spotify
...
-Switch to the gnome terminal in the Multimedia Template VM
+Check the signature of the signing key (still in the AppVM where you downloaded the key)
 
-`nano spotify.pubkey`
+`gpg --with-fingerprint spotify.pubkey`
 
-Paste the content from the Qubes Clipboard into nano (Shift+Ctrl+V and then Paste)
-Save the file (Ctrl+O <Enter> Ctrl+X)
+This should look like:
+> [user@my-untrusted ~]$ `gpg --with-fingerprint spotify.pubkey`
```
marmarek: If the VM is untrusted (as the name correctly suggests), you shouldn't trust output of gpg there... IOW the fingerprint verification should be done after moving the file out of this VM. This part was done right in previous iteration (it was just about order of apt-key add and fpr verification, not where to do it)...
There is alternative view on this - possibly malicious key file could target some bug in gpg and exploit it in the template, but:
 - validating fingerprint in untrusted (potentially compromised VM) does not guard against it
 - we need to trust gpg to handle untrusted input correctly anyway, because that's what happening when downloading packages

praschdorff: So you think the `gpg --with-fingerprint` part should be done in the multimedia template VM?
1) Download PGP Signing Key
2) Copy Key to Multimedia Template VM
3) Verify Fingerprint
4) Import Key (if Fingerprint is ok)

?

marmarek: Yes, exactly.

praschdorff:
On 02/21 03:06, Marek Marczykowski-Górecki wrote:

> Yes, exactly.

Moved the verification of the GPG Signing keys to the Template VM.

- P

andrewdavidwong:
@praschdorff: There are a couple of minor typos in the Travis log.

marmarek:
One more typo: `temmplate`

PR #423, opened 2017-05-13
"sign-off" procedure description has moved in the linux kernel tree

zby:
I've updated the link.
I think the line should also be broken into many - so that changes like this are easier to see.


jpouellet:
> I think the line should also be broken into many - so that changes like this are easier to see.

Yes, there is an intention is to move to "one sentence" <=> "one line" policy in the docs overall to ease both reviewing and translating.

PR #425, opened 2017-05-18
Edit + Disagreement

stubbybubby:
Did some prosaic editing:
* Removed unnecessary parentheses
* Shortened long sentences
* Wording changes.
* Removed restatements of earlier sentences
Also, this document made the following error about cooperative covert leaking channels in Qubes OS:

> It is likely that the only way to **fully protect against leaks of type 1** and 2 is to either pause or shut down all other VMs while performing sensitive operations in the target VM(s) (such as key generation).

This is wrong. Closing the other VMs while performing such important activities does nothing to stop leaks of type 1, assuming you turn the other VMs back on at some point. The (presumably compromised) AppVM in question can easily write the information it needs to leak down until the other Qubes come back online. Inserted a new sentence clarifying this.


andrewdavidwong:
Thank you for your contribution. We apologize for the long delay before merging.

The only edit I selectively reverted was the "attacker emailing himself" sentence, which I thought was clearer before. All the other edits were improvements. Good catch with the "type 1 and 2" leaks!

PR #429, opened 2017-05-25
dpi  option for supporting L7 firewall rules

Zrubi:
Added the "dpi" Deep Packet Inspection option for supporting L7 firewall rules


marmarek:
Is "dpi" established term for this? I know this acronym only for "dots per inch"...

Zrubi:
In context of firewalls yes:
https://www.google.hu/search?q=firewall+dpi

PR #435, opened 2017-06-19
Update vpn.md

privacyguy:


privacyguy:
updated with some additional, more specific directions. (These, additional pointers would have truly saved me hours in getting my config going)

privacyguy:
Checking this pull request from June and nearly 5 months later, it still hasn't been commented upon or added to update the guide in qubes-doc.  I spent a number of hours on these revisions -- I would truly like them to be at least looked at.  I was going to update another qubes-doc that needed work, but will hold off to see if anything happens to this one.  (Not worth my labors, if my revisions are simply ignored).

andrewdavidwong:
> Checking this pull request from June and nearly 5 months later, it still hasn't been commented upon or added to update the guide in qubes-doc. I spent a number of hours on these revisions -- I would truly like them to be at least looked at. I was going to update another qubes-doc that needed work, but will hold off to see if anything happens to this one. (Not worth my labors, if my revisions are simply ignored).

@privacyguy: We sincerely apologize for the delay. It is entirely understandable that you would feel that way after putting hours of work into a PR without any response. Unfortunately, our documentation system is almost entirely a volunteer effort right now, so it comes down to when people with the required expertise happen to have free time to review pull requests. Regrettably, we don't have the resources to hire or redirect salaried staff to this area at the moment.

@tasket, @Zrubi, @unman: Would any of you have a spare moment to review this PR?

tasket:
Thanks for the input, although I don't agree with all of this from just reading the changes. It might be more constructive to point out where and why the existing version seems confusing; users can open issues to discuss problems with documentation.

* This looks wrong:

> These directions are specifically written for those wishing to use OpenVPN as their VPN client software

When I wrote it, I took great pains to address strenuous objections to any appearance that the solution was specific to only one VPN client software. As it is, the scripts work well and yet have OpenVPN-specific code in only a couple variable assignments in rc.local. It is fair to say that the doc is a general solution that includes OpenVPN as a working example and if clarification is needed, it would be best to state that instead.

In guessing what was specifically confusing for privacyguy, I think its the fact that OpenVPN is VPN client software referenced as an example. But in assisting people in qubes-users with various VPN issues, I don't recall this coming up as a problem. Still, the doc never states "We are showing OpenVPN as a functional example here" and this could be made clear with only small changes to the 1st or 3rd paragraph of the section. Adding multiple paragraphs seems unnecessary.

* There also seems to be some confusion over `nano`. I'm guessing the nano command lines in each step looked unfamiliar, while the mention of installing nano was only made in passing near the start of the instructions. This may be a case where trying to be very specific (writing-out command lines for the editor) to make things easy can confuse some people instead. A way to address this would be to remove the `nano` commands and simply mention 'edit this file...' in keeping with most other Qubes docs. Although, TMK no one has complained about this before, either.

* Edits from line 79 "First, make sure..." through 98 seem OK, although its adding a discrete step to the instructions in step 2. I think it would be better overall to just add `--auth-user-pass auth.txt` to the rc.local script and possibly add a few lines of code to automate inputting/creating the file (i.e. if trying to make it easier, make it much easier).
    - Instructions for using `gnome-terminal` seem over-specific, but I'm not sure what Qubes custom is in this regard.

* Line 100 typo: auth-user-pass must be prefixed with a double-dash when used as a command option.
* Lines 233+234 look spurious.

----
Since Qubes R4 is due to be released soon and the current scripts cannot work in it, this doc is due for a major revision anyway. I already have a working version that needs a bit more revision and will be simpler to use overall.

Instead of merging, I'd _suggest using the edits here as considered input_ for the R4-compatible document.

---

andrewdavidwong:
It seems like a good idea to hold off for the major R4 revision, but I'd like to give @privacyguy (and any others with a stake in this) a chance to respond before closing this PR. @privacyguy, do you mind if we take your contributions into account when revising this document for R4 rather than merging this PR as-is? I want to emphasize that we greatly appreciate your contributions, and I hope you feel that it's worthwhile to continue to share them with the project in the future.

privacyguy:
@andrewdavidwong  and @tasket , thank you for looking into this and your responses. 
--
To reply to tasket's comments:
My hope in making the revisions was to provide specific, step-by-step directions for those wishing to set up a VPN in Qubes.  While one could use a different VPN client, openVPN is a real-world solution that currently exists.  Therefore, it would be very beneficial to Qubes-OS users if there are specific directions in setting up a VPN w/ a real-world solution that works (i.e. openVPN).  

"Too specific"
IMO specificity should never be an "issue":  If were are trying to make Qubes-OS, and it's use as user-friendly as possible (IMO, an important goal), then we need to be willing to "break out the crayola crayon" / "show the math", so to speak, to those users that are not as advanced.
------
About the Doc:
I understand that this Doc (in it's current form) and pull request will be deprecated w/ the coming release of R4.  However, I believe that this Doc (and all old Qubes-OS docs pertinent to R3.x but not R4) should be kept and placed in a separate Docs catalog, demarcated as Qubes-OS R3.x documentation.  This way users who are still using R3.x (for whatever reason, a very real reason being R4 hardware compatibility) will still be able to refer to the 3 Docs as needed.

Pull Request:
Feel free to do whatever you wish with any of my edits/revisions.  I do not need any recognition for them. I just request that my concerns be taking into account:  namely that the future Doc will be more user friendly and will have step-by-step directions for setting up a VPN in Qubes-OS R4 and making it work with a real-world VPN-client.

tasket:
@privacyguy Thanks again... I will definitely take your input into account. I'm also a believer in providing context and specifics to the user, although this doc was a bit harder to formulate because several complete scripts had to be stuffed in among the text. In the meantime, my downloadable [VPN project](https://github.com/tasket/Qubes-vpn-support) is very easy to setup for R3.2 and will serve as testbed for the revised doc.

andrewdavidwong:
> I believe that this Doc (and all old Qubes-OS docs pertinent to R3.x but not R4) should be kept and placed in a separate Docs catalog, demarcated as Qubes-OS R3.x documentation. This way users who are still using R3.x (for whatever reason, a very real reason being R4 hardware compatibility) will still be able to refer to the 3 Docs as needed.

That's a good point. We will, of course, keep all the existing R3.x documentation around (we always keep all revisions of everything forever via git and backups), though perhaps not in a separate docs catalog or specifically demarcated as R3.x documentation, since this would require a significant reworking of our documentation infrastructure that we would not complete in time.

We've never had entire documentation sets as version-specific documentation. It's always just been a rolling update targeting the latest stable version. But we've also never had a very long overlapping period of rather different supported versions as we expect to have with R3.2 and R4.0. Typically, what we'd do in cases where functionality for a given feature differs significantly between versions is to have two sections within the same document -- one devoted to each version. This is probably how we'll approach adding new R4.0 documentation to the existing docs.

PR #446, opened 2017-07-12
Fix PAM auth for 'su' command

tasket:
The behavior seen when a simple `auth ... pam_exec` rule fails to permit `su` (despite user response to prompt) appears to fit a pattern of PAM quirks that require more elaborate rule-making.

Here, I use a rule pattern employed by the Debian distro, which uses the pam_permit and pam_deny modules to put the rule's success/fail decision on solid footing. The result is that 'su' can now be used when auth dialogs are enabled.

https://github.com/QubesOS/qubes-issues/issues/2693


tasket:
Note: The first commit represents a 'catch up to upstream' PR from Qubes to tasket. Let me know if this presents a problem.

PR #457, opened 2017-08-30
fixed path where .desktop files are kept

Yethal:


marmarek:
Both will work. Do you see any specific preference for the later? Maybe even better choice would be `~/.local/share/applications`.

Yethal:
On my system the dispvm desktop files are kept at /usr/local/share/applications. usr/share/applications contains contents of System Tools menu. ~/.local/share/applications contains .desktop files for all AppVMs but not for DispVMs. Therefore, the instructions provided in the docs are incorrect because the dispvm .desktop files are not present at the provided directory. 
Although moving them to ~/.local would make this customization persistent across backups which by itself is a good idea.

unman:
I would favour ~/.local/share/applications only because it would pass with backups. But that's also a reason NOT to use it, because it assumes that the dispVM or dispVM Template has also been backed up and restored. That seems to me to provide scope for confusion.
Personally, I dont see any reason to make a change.

Yethal:
Change in the docs was made due to my observation that the directory provided in the doc does not match the actual location of the files. So unless the qubes-dispvm-*.desktop files are kept in /usr/local/share/applications on my machine only can we separate those two issues (mismatch between Qubes config and documentation and whether the directory chosen for those files is a correct one)
disp

marmarek:
Indeed on R3.2 those files are in `/usr/local/share/applications`. I'd recommend adding custom one in `~/.local/share/applications`, as @unman suggests. But this change is an improvement anyway.

andrewdavidwong:
> Indeed on R3.2 those files are in `/usr/local/share/applications`. I'd recommend adding custom one in `~/.local/share/applications`, as @unman suggests. But this change is an improvement anyway.

@unman, @Yethal: Please consider opening a separate PR for this.

PR #459, opened 2017-09-11
Additional info for installing Windows 7 in a HVM - addressing two issues that come up fairly often

Jommoner:
Added two sections, one addressing the issue where it seems that 512MB is the 'max' memory size of a HVM, leading to blue screens of death after several steps into the installer, and another addressing the issue whereby the Windows 7 installer 'freezes' on the 'Starting Windows' screen, with the windows logo still pulsing.

### DIFF
--- a/managing-os/hvm.md
+++ b/managing-os/hvm.md
@@ -36,7 +36,34 @@ If you receive an error like this one, then you must first enable VT-x in your B
 libvirt.libvirtError: invalid argument: could not find capabilities for arch=x86_64 
 ~~~
 
-Now we need to install an OS inside this VM.  This can done by attaching an installation ISO to and starting the VM (this can currently only be done from command line, but in the future we will surely add an option to do this also from the manager):
+Installing an operating system in the HVM domain
+------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


Jommoner:
One issue - I searched for 'nano text editor exploits' and there are some, that can result in privilege escalation, generally by the attacker doing things to the file on the file system while you are editing it. This probably couldn't happen in dom0 due to it being isolated, but feel free to change from 'nano' to a more secure editor if there is one!

Jommoner:
I saw the 'Travis' CI found two typos (Occured) and another one - I can't change them now, but would be grateful if they could be changed before the page is updated - if the update is appropriate of course!

Jommoner:
I am not sure how to fix the 'Travis CI build' error, so my commit is stuck in limbo; I will close the request for now; of note, two problems that cause unpredictable errors in Win7 HVMs are probably fixed by the info in this commit, which could help others, so I will try to re-commit in a few weeks!

marmarek:
That travis builds errors were about typos. You can see details at the of its log. For convenience:
```
    +1024MB
    +1536MB
    +512MB
    +Win7
    +customised
    +greyed
    +occuring
    +troubleshoouting

The introduced suspects in context are:

(...)

If these are false-positives, simply commit as is and they will be ignored in the future.
```
At least some of them are not false-positives.

PR #464, opened 2017-09-27
set Qubes policy to allow by default for USB Input

adrelanos:
https://github.com/QubesOS/qubes-mgmt-salt-dom0-virtual-machines/blob/master/qvm/sys-usb.sls ets `/etc/qubes-rpc/policy/qubes.InputMouse` to `sys-usb dom0 allow,user=root`.

A user following `How to use a USB keyboard` / `How to use a USB mouse` probably wants this. Otherwise, if the user doesn't have a PS2 keyboard / mouse, the user locks itself out. I suggest to either change to `allow` by default (already done by sys-usb salt config) or to explain this in documentation.

Please carefully review this since it's not my area of expertise.

### DIFF
--- a/common-tasks/usb.md
+++ b/common-tasks/usb.md
@@ -165,7 +165,7 @@ Edit the `qubes.InputKeyboard` policy file in dom0, which is located here:
 
 Add a line like this one to the top of the file:
 
-    sys-usb dom0 ask,user=root
+    sys-usb dom0 allow,user=root
 
 (Change `sys-usb` to your desired USB qube.)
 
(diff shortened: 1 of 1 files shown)


marmarek:
Giving `qubes.InputKeyboard` access to sys-usb, means it can do almost anything in the system. Setting it to `ask` at least let you know that sys-usb can now have control over the keyboard. If it isn't your primary keyboard, you probably want to know that. But if that's your primary input method, you indeed may want to set it to allow. But this should be conscious decision, understanding the risk. It is already explained on that page.
@andrewdavidwong do you think that explanation is clear enough to suggest (also?) "allow" action?

andrewdavidwong:
The section [Security Warning about USB Input Devices](https://www.qubes-os.org/doc/usb/#security-warning-about-usb-input-devices) clearly explains the security risks of using USB keyboards and USB mice. However, what is not explained anywhere on the USB page is the difference between `ask` and `allow`. That's what needs to be explained in order for the user to understand what @marmarek just said in the comment above.

PR #470, opened 2017-10-19
debian-8-backports electrum has bug, use fedora-25?

thomaskey:
A bug in debian-8-backports electrum package makes network request when sending offline transaction.  When Qubes has no network vm the electrum hangs when making offline transactions.  It was suggested to use fedora-25 in issues.

### DIFF
--- a/security/split-bitcoin.md
+++ b/security/split-bitcoin.md
@@ -17,28 +17,24 @@ A "split" bitcoin wallet is a strategy of protecting your bitcoin by having your
 A "Watching" Wallet and a "Cold" Wallet
 ---------------------------------------
 
-1. Create a Debian 8 backports template using the Qubes VM Manager or running
-   `qvm-clone debian-8 debian-8-backports` in dom0.
+1. Create a fedora-25-electrum template using the Qubes VM Manager or running
+   `qvm-clone fedora-25 fedora-25-electrum` in dom0.
 
-2. Add backports to the sources for the new template by opening a terminal in
-   the new template, run `sudo vi /etc/apt/sources.list` and add
-   `deb http://http.debian.net/debian jessie-backports main`.
+2. Start the new template and install electrum:
+   `qvm-start fedora-25-electrum`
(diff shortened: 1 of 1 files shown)


security/split-bitcoin.md
```diff
@@ -17,30 +17,26 @@ A "split" bitcoin wallet is a strategy of protecting your bitcoin by having your
 A "Watching" Wallet and a "Cold" Wallet
 ---------------------------------------
 
-1. Create a Debian 8 backports template using the Qubes VM Manager or running
-   `qvm-clone debian-8 debian-8-backports` in dom0.
+1. Create a fedora-25-electrum template using the Qubes VM Manager or running
+   `qvm-clone fedora-25 fedora-25-electrum` in dom0.
```
marmarek: Since there is no need to add/enable additional repositories in the fedora-25 template, I'd mark this step as optional. But on the other hand, that would make the instruction more complex (because of possibly different template name). @andrewdavidwong what do you think?

andrewdavidwong: I'd just mark it as optional.

andrewdavidwong: @thomaskey, do you want to mark this step as optional?

thomaskey: Since the offline qube can't install electrum, the previously documented strategy was to install electrum into a template first.  You "could" install electrum in your default template but I like a clean default template with no extra packages.  So I like the above steps as is.

thomaskey: I'm not sure you gain any extra security following the extra step (opposed to cloning default template, install package, then switch off net vm - vs - installing electrum into your default template than cloning).  But I like the above steps as encouraging a cleaner default template.

andrewdavidwong: Ok, that's fine. Users can learn that this step is optional from the primary template documentation, then decide for themselves.

andrewdavidwong:
@thomaskey, thank you for your contribution.

PR #477, opened 2017-11-13
specifiy where to change what

adrelanos:


marmarek:
Oh, I see the documentation does not mention we have packaged those scripts. It's here: https://github.com/QubesOS/qubes-app-yubikey (packages available in 3.2 and 4.0 repos)

adrelanos:
Actually, please hold.

>  +3. Create `/usr/local/bin/yubikey-auth` script your USB VM: 

This is wrong. Correcting myself. In https://groups.google.com/forum/#!topic/qubes-users/r7evJ_LoO8I you said in dom0. And since that script runs `qvm-run -u root --nogui -p sys-usb`, I am very sure I was wrong. Will correct in another PR.

adrelanos:
From https://www.qubes-os.org/doc/yubi-key/ I also doubt the following line would work.

    hash=`head -c -1 | openssl dgst -sha1 -r | cut -f1 -d ' '`

`openssl dgst -sha1 -r` is blocking and nothing was `echo`ed.

----

Note to self:

    sudo qubes-dom0-update qubes-yubikey-dom0

marmarek:
> openssl dgst -sha1 -r is blocking and nothing was echoed.

Actually something was echoed. pam_exec send a password on stdin.

PR #479, opened 2017-11-13
improve yubikey usage instructions

adrelanos:


marmarek:
The content is ok. Actually it is only formatting change. But I haven't checked if this approach for enumerated list work with github pages...

PR #481, opened 2017-11-14
make commands more copy and paste friendly

adrelanos:


andrewdavidwong:
Part of the reason for having the terminal prompts is to remind the reader which VM the commands should be entered in. Are they really *that* copy/paste unfriendly? I mean, can't one simply not copy the terminal prompt part?

adrelanos:
Yes, because otherwise you could just double click into the field and it would select all. No matter how long the line is. If you manually click to select, it's easier to forget something. Saw this quite a few time when supporting users.

andrewdavidwong:
Ok, sounds good.

andrewdavidwong:
Actually, I think that if you're going to remove these, you need to precede each command (not in the code block, but on the line above it) with something like "in dom0" or "in `debian-8`". Otherwise it's too confusing where each command should be entered.

andrewdavidwong:
FWIW, it's extremely common in instructions to include `$` at the beginning of a code block for a terminal command ([example](https://docs.docker.com/engine/installation/linux/docker-ce/debian/#set-up-the-repository)). This also interferes with the "copy/paste the whole code block" method, so I'm not sure we really need to go out of our way to support that method (i.e., since almost no one else does).

adrelanos:
> Actually, I think that if you're going to remove these, you need to precede each command (not in the code block, but on the line above it) with something like "in dom0" or "in debian-8". Otherwise it's too confusing where each command should be entered.

Agreed.

> FWIW, it's extremely common in instructions to include $ at the beginning of a code block for a terminal command (example). This also interferes with the "copy/paste the whole code block" method, so I'm not sure we really need to go out of our way to support that method (i.e., since almost no one else does).

That's a hard decision. I would defer it to a usability designer. The ux designer who worked with us (bnvk) wrote/rewrote https://www.qubes-os.org/doc/whonix/install/ (among other Whonix documentation pages on qubes-os.org) where he preferred the "in dom0" plus copy/paste friendly instructions. That's why I would think that's the way to go.

andrewdavidwong:
I don't have a strong opinion either way, so long as it's unambiguous which VM each command should be entered in.

andrewdavidwong:
This PR is being closed because there has been no activity on it in over a year. However, if anyone would like it to be reopened, simply leave a comment saying so, and we'll be happy to reopen it. Thank you.

PR #491, opened 2017-12-02
Add instructions for newer UEFI systems

voltylicious:
Add instructions on how to create a USB installation medium that works on newer ThinkPad systems in UEFI-only mode. This is necessary because existing troubleshooting steps for ThinkPad and UEFI are no longer sufficient. The content was created based on the thread at https://groups.google.com/forum/#!topic/qubes-users/TEmVIozLJh0


voltylicious:
previous pull request 490 closed because of a typo. remaining spell check failures are false positives.

kushaldas:
Tried the instructions given in this patch, and they worked as suggested. Thank you for this.

PR #496, opened 2017-12-12
Add link to email archive

jackdesert:
It's been challenging for me to find the email archive. I find it one day, then the next day I'm duckduckgoing again and can't find it. 

Let's put it on the website so we can just click the link!

### DIFF
--- a/about/mailing-lists.md
+++ b/about/mailing-lists.md
@@ -134,8 +134,9 @@ Examples of topics or questions suitable for this list include:
 
 ### Read these first
 
-Please try searching both the Qubes website and the archives of the mailing
-lists before sending a question. In addition, please make sure that you have
+Please try searching both the Qubes website and the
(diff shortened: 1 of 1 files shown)


jackdesert:
@andrewdavidwong 
1. Thank you for clarification regarding the google group already being a searchable archive. 

2a. I have pushed another commit to my branch be clear that the mail-archive.com version is *in addition* to the google group.

2b.  One useful feature of the mail-archive.com site is that it is searchable without enabling javascript. 

3. This time I have included the link for both qubes-users and qubes-devel. I tried using the same url scheme for qubes-translation and qubes-project, but they lead to 404 errors as seen in the following links:

https://www.mail-archive.com/qubes-translation@googlegroups.com/
https://www.mail-archive.com/qubes-project@googlegroups.com/

PR #505, opened 2018-01-05
Create Fedora 25 to 26 upgrade instructions

andrewdavidwong:
See: QubesOS/qubes-issues#3429

I basically just took the existing F24->F25 documentation and updated the version numbers. I've tested the summary procedure for the full template (which worked fine), but I haven't tested the minimal template. I'm submitting this for @marmarek's review in case there are any new security issues (compared to F24->F25) that I may not be aware of (so that we can ensure the instructions are accurate and safe before we recommend that all Qubes users perform these actions).

### DIFF
--- a/managing-os/templates/fedora/upgrade-25-to-26.md
+++ b/managing-os/templates/fedora/upgrade-25-to-26.md
@@ -1,0 +1,241 @@
+---
+layout: doc
+title: Upgrading the Fedora 25 Template to Fedora 26
+permalink: /doc/template/fedora/upgrade-25-to-26/
+redirect_from:
+- /doc/fedora-template-upgrade-25/
+- /en/doc/fedora-template-upgrade-25/
+- /doc/FedoraTemplateUpgrade25/
+- /wiki/FedoraTemplateUpgrade25/
+---
+
+Upgrading the Fedora 25 Template
+================================
+
+This instruction is about upgrading Fedora 25 to Fedora 26 template. The same
+instructions can be used to upgrade Fedora 24 to Fedora 26 - simply start with
+cloning fedora-24 instead of fedora-25.
+
+Summary: Upgrading the Standard Fedora 25 Template to Fedora 26
+---------------------------------------------------------------
+
+**Note:** The prompt on each line indicates where each command should be entered
+(`@dom0` or `@fedora-26`).
+
+        [user@dom0 ~]$ qvm-clone fedora-25 fedora-26
+        [user@dom0 ~]$ truncate -s 5GB /var/tmp/template-upgrade-cache.img
+        [user@dom0 ~]$ qvm-run -a fedora-26 gnome-terminal
+        [user@dom0 ~]$ qvm-block -A fedora-26 dom0:/var/tmp/template-upgrade-cache.img
+        [user@fedora-26 ~]$ sudo mkfs.ext4 /dev/xvdi
+        [user@fedora-26 ~]$ sudo mount /dev/xvdi /mnt/removable
+        [user@fedora-26 ~]$ sudo dnf clean all
+        [user@fedora-26 ~]$ sudo dnf --releasever=26 --setopt=cachedir=/mnt/removable distro-sync
+
+    (Shut down TemplateVM by any normal means.)
+
+        [user@dom0 ~]$ rm /var/tmp/template-upgrade-cache.img
+        [user@dom0 ~]$ qvm-trim-template fedora-26
+
+(Optional cleanup: Switch everything over to the new template and delete the old
+one. See instructions below for details.)
+
+Detailed: Upgrading the Standard Fedora 25 Template to Fedora 26
+----------------------------------------------------------------
+
+These instructions will show you how to upgrade the standard Fedora 25
+TemplateVM to Fedora 26. The same general procedure may be used to upgrade any
+template based on the standard Fedora 25 template.
+
+**Note:** The command-line prompt on each line indicates where each command
+should be entered (`@dom0` or `@fedora-26`).
+
+ 1. Ensure the existing template is not running. 
+
+        [user@dom0 ~]$ qvm-shutdown fedora-25
+ 
+ 2. Clone the existing template and start a terminal in the new template.
+
+        [user@dom0 ~]$ qvm-clone fedora-25 fedora-26
+        [user@dom0 ~]$ qvm-run -a fedora-26 gnome-terminal
+
+ 3. Attempt the upgrade process in the new template.
+
+        [user@fedora-26 ~]$ sudo dnf clean all
+        [user@fedora-26 ~]$ sudo dnf --releasever=26 distro-sync
+
+    **Note:** `dnf` might ask you to approve importing a new package signing
+    key. For example, you might see a prompt like this one:
+
+        warning: /var/cache/dnf/fedora-d02ca361e1b58501/packages/python2-babel-2.3.4-1.fc26.noarch.rpm: Header V3 RSA/SHA266 Signature, key ID 8fdb19c98: NOKEY
+        Importing GPG key 0x81B46521:
+         Userid     : "Fedora (26) <fedora-26-primary@fedoraproject.org>"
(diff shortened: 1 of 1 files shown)


managing-os/templates/fedora/upgrade-25-to-26.md
```diff
@@ -0,0 +1,241 @@
...
+    and other options).
+
+        [user@dom0 ~]$ qvm-trim-template fedora-26
+
+ 7. (Recommended) Switch everything that was set to the old template to the new
+    template, e.g.:
+
+     1. Make the new template the default template:
+
+        Qubes Manager --> Global settings --> Default template
```
marmarek: R4.0: Menu -> System Tools -> Qubes Global Settings -> Default template

managing-os/templates/fedora/upgrade-25-to-26.md
```diff
@@ -0,0 +1,241 @@
...
+
+     2. Base AppVMs on the new template. In Qubes Manager, for each VM that is
+        currently based on `fedora-25` that you would like to base on
+        `fedora-26`, enter its VM settings and change the Template selection:
+
+        Qubes Manager --> (Select a VM) --> VM settings --> Template
```
marmarek: Qubes 4.0: Menu -> (select a VM) -> VM settings -> Template

managing-os/templates/fedora/upgrade-25-to-26.md
```diff
@@ -0,0 +1,241 @@
...
+
+     3. Base the [DispVM] template on the new template.
+
+        If you have set the new template as your default template:
+
+            [user@dom0 ~]$ qvm-create-default-dvm --default-template
+
+        Otherwise:
+
+            [user@dom0 ~]$ qvm-create-default-dvm fedora-26
```
marmarek: Qubes 4.0:
```
qvm-create -l red -t fedora-26 fedora-26-dvm
qvm-prefs fedora-26-dvm template_for_dispvms True
qvm-features fedora-26-dvm appmenus-dispvm 1
qubes-prefs default-dispvm fedora-26-dvm
```

managing-os/templates/fedora/upgrade-25-to-26.md
```diff
@@ -0,0 +1,241 @@
...
+
+
+Summary: Upgrading the Minimal Fedora 25 Template to Fedora 26
+--------------------------------------------------------------
+
+**Note:** The prompt on each line indicates where each command should be entered
+(`@dom0` or `@fedora-26`).
+
+        [user@dom0 ~]$ qvm-clone fedora-25-minimal fedora-26-minimal
+        [user@dom0 ~]$ qvm-run -a fedora-26-minimal xterm
```
marmarek: Better use `qvm-run -u root -a fedora-26-minimal xterm` instead. This method is good for both 3.2 and 4.0.
This will avoid su/sudo/others differences (minimal template in R4.0 do not have passwordless su by default). This do not work with gnome-terminal, unfortunately.

managing-os/templates/fedora/upgrade-25-to-26.md
```diff
@@ -0,0 +1,241 @@
...
+    `xterm` for the standard template, if you prefer.)
+ 2. `sudo` is not installed by default. Unless you've installed it, use `su` as
+    demonstrated above. (Of course, you can also use `su` for the standard
+    template, if you prefer.)
+
+
+Compacting the Upgraded Template
+--------------------------------
+
+Neither `fstrim` nor the `discard` mount option works on the TemplateVM's root
```
marmarek: This is no longer true in Qubes 4.0

managing-os/templates/fedora/upgrade-25-to-26.md
```diff
@@ -0,0 +1,241 @@
...
+
+Upgrading StandaloneVMs
+-----------------------
+
+The procedure for upgrading a StandaloneVM from Fedora 25 to Fedora 26 is the
+same as for a TemplateVM, except that `qvm-trim-template` does not work on
+StandaloneVMs. Instead, you should run the following command inside the
+StandaloneVM in order to compact it:
+
+    $ sudo fstrim -v -a
```
marmarek: Well, this will work on Qubes 4.0 for templates too...

andrewdavidwong:
> There are some differences for Qubes 4.0. Do we have already some workflow for handling such differences in documentation? We need to update other docs for it very soon...

Yes, I just added this section to the Documentation Guidelines to address this issue:

https://www.qubes-os.org/doc/doc-guidelines/#version-specific-documentation

If you have something different in mind by "workflow," or think more detail is needed, let me know.

andrewdavidwong:
@marmarek: Ok, how does it look now?

managing-os/templates/fedora/upgrade-25-to-26.md
```diff
@@ -147,15 +173,125 @@ should be entered (`@dom0` or `@fedora-26`).
...
+        currently based on `fedora-25` that you would like to base on
+        `fedora-26`, enter its VM settings and change the Template selection:
+
+        Qubes 4.0: Menu --> (select a VM) --> VM settings --> Template
+
+     3. Base the [DispVM] template on the new template.
+
+        If you have set the new template as your default template:
+
+            [user@dom0 ~]$ qvm-create-default-dvm --default-template
```
marmarek: This command is gone in Qubes 4.0, so there is only one case (the one below).

managing-os/templates/fedora/upgrade-25-to-26.md
```diff
@@ -147,15 +173,125 @@ should be entered (`@dom0` or `@fedora-26`).
...
+ 6. Remove the cache file, if you created one.
+
+        [user@dom0 ~]$ rm /var/tmp/template-upgrade-cache.img
+
+ 7. (Recommended) Switch everything that was set to the old template to the new
+    template, e.g.:
+
+     1. Make the new template the default template:
+
+        R4.0: Menu --> System Tools --> Qubes Global Settings --> Default template
```
marmarek: Lets be consistent: `Qubes 4.0` instead of `R4.0` here. But since this section apply to Qubes 4.0 only, probably the prefix is not needed at all.

andrewdavidwong:
> I wonder if it's possible to not duplicate the whole instruction (for 3.2 and 4.0), while still having it readable. At least the "detailed" part. Since you've duplicated it, I guess it was the only option for having it readable, right?

We could do it either way, but this way seemed slightly better to me. My concern is not just readability, but also consistency and maintainability.

Consistency: If I'm on 3.2 or 4.0, will I always be able to see instructions relevant to me by scanning the headings, or must I look within each section for a step that bifurcates into instructions for each version? (This also applies across different docs.) Also, suppose we have a section that's mostly the same, but then we gradually add more version-specific differences until it seems obvious that there should be two different sections. It might be hard to consistently draw this line (especially across docs), which could also make it harder for users to find the content relevant to their specific versions.

Maintainability: Often, when 4.0 users contribute to the docs, they try to *replace* 3.2 content with 4.0 content (probably because, even though 3.2 is still a supported release, they no longer use it, so they think of it as deprecated). Having a policy of always making a new section for a different version seems like it might make this easier.

marmarek:
Fair enough.

managing-os/templates/fedora/upgrade-25-to-26.md
andrewdavidwong: In that case, I would go with no prefix. But it would be nice to have a name for that menu other than just `Menu`. That doesn't uniquely identify it, since there are a lot of menus.

marmarek: When I hover over "Q" icon, there is "Applications" tooltip. Is it good enough for name?

andrewdavidwong: Yes, `Applications Menu` sounds like the right name.

andrewdavidwong:
Ok, I followed my own policy and more clearly segregated the main instructions for 3.2 and 4.0. Let me know if you see any more room for improvement, @marmarek.

PR #507, opened 2018-01-05
Update fedora-minimal.md

mossy-nw:
* updated for fedora-26-minimal template (no-one should safely be using -25- anymore, correct?)
* removed packages installed by default
* I've never seen the minimal template weigh less than 2GB (was 300mb from a pre-3.2 version of qubes?)
* added Qubes R4.0 specific things (I've always seen sudo installed by default for the R3.2 minimal template)

### DIFF
--- a/managing-os/templates/fedora-minimal.md
+++ b/managing-os/templates/fedora-minimal.md
@@ -12,7 +12,7 @@ redirect_from:
 Fedora - minimal
 ================
 
-The template only weighs about 300 MB and has only the most vital packages installed, including a minimal X and xterm installation.
+The template only weighs about 2.3 GB and has only the most vital packages installed, including a minimal X and xterm installation.
 The minimal template, however, can be easily extended to fit your requirements. The sections below contain the instructions on duplicating the template and provide some examples for commonly desired use cases.
 
 Installation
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
> updated for fedora-26-minimal template (no-one should safely be using -25- anymore, correct?)

Correct. We'll be making an announcement ASAP (see https://github.com/QubesOS/qubes-issues/issues/3429).

> I've never seen the minimal template weigh less than 2GB (was 300mb from a pre-3.2 version of qubes?)

@FabrizioRomanoGenovese observed 600MB in https://github.com/QubesOS/qubes-doc/pull/508#discussion_r160035952

I wonder what accounts for the different observations. Comparing compressed vs. uncompressed?

> added Qubes R4.0 specific things (I've always seen sudo installed by default for the R3.2 minimal template)

My `fedora-26-minimal` on R3.2 does appear to have `sudo` installed by default.

andrewdavidwong:
> @FabrizioRomanoGenovese observed 600MB in #508 (comment)
> 
> I wonder what accounts for the different observations. Comparing compressed vs. uncompressed?

Or perhaps even comparing compressed vs. installed (size on disk)?

andrewdavidwong:
> Or perhaps even comparing compressed vs. installed (size on disk)?

Yep: https://github.com/QubesOS/qubes-doc/pull/508#discussion_r160036362

PR #508, opened 2018-01-06
Update fedora-minimal.md

FabrizioRomanoGenovese:
Information regarding fedora-26-minimal template and customization in Qubes 4.0 added.

### DIFF
--- a/managing-os/templates/fedora-minimal.md
+++ b/managing-os/templates/fedora-minimal.md
@@ -52,7 +52,7 @@ The following table provides an overview of which packages are needed for which 
 As expected, the required packages are to be installed in the running template with the following command. Replace "packages` with a space-delimited list of packages to be installed.
 
 ~~~
-[user@your-new-clone ~]$ sudo dnf install packages
+[user@your-new-clone ~]$ sudo dnf --enablerepo=qubes-vm-*-current-testing install packages
 ~~~
 
 Use case | Description | Required steps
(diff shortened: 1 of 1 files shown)


managing-os/templates/fedora-minimal.md
```diff
@@ -52,7 +52,7 @@ The following table provides an overview of which packages are needed for which
 As expected, the required packages are to be installed in the running template with the following command. Replace "packages` with a space-delimited list of packages to be installed.
 
 ~~~
-[user@your-new-clone ~]$ sudo dnf install packages
+[user@your-new-clone ~]$ sudo dnf --enablerepo=qubes-vm-*-current-testing install packages
```
andrewdavidwong: This suggests that using the current-testing repo is required, but some users on stable systems may not want to install packages that are still in testing.

managing-os/templates/fedora-minimal.md
```diff
@@ -65,11 +65,30 @@ Use case | Description | Required steps
...
-
+ 
 A comprehensive guide to customizing the minimal template is available [here][GUIDE]
 
+Qubes 4.0, fedora-26
+------
+In Qubes4.0 the `fedora-26-minimal` template is available and can be installed issuing the following command:
+
+~~~
+[user@dom0 ~]$ sudo qubes-dom0-update qubes-template-fedora-25-minimal
```
andrewdavidwong: Should be `26` instead of `25`.

managing-os/templates/fedora-minimal.md
```diff
@@ -65,11 +65,30 @@ Use case | Description | Required steps
...
+~~~
+
+This template is slightly bigger (~600MB). All the suggestions made in the previous section apply. In addiction to this, in many situations packages from the qubes-core-agent suite will be needed to make the customized minimal template to work properly. These packages are:
```
andrewdavidwong: Interesting. @mossy-nw claims it's at least 2GB in https://github.com/QubesOS/qubes-doc/pull/507.

managing-os/templates/fedora-minimal.md
```diff
@@ -65,11 +65,30 @@ Use case | Description | Required steps
...
+
+- `qubes-core-agent-passwordless-root`, `polkit`: By default the 'fedora-26-minimal' template doesn't have passwordless root. These two packages fix the situation.
+- `qubes-core-agent-nautilus`: This package provides integration with the Nautilus file manager (without it things like "copy to VM/open in disposable VM" will not be shown in Nautilus).
+- `qubes-core-agent-qrexec`: Qubes qrexec agent.
+- `qubes-core-agent-systemd`: Qubes unit files for SystemD init style.
+- `qubes-core-agent-sysvinit`: Qubes unit files for SysV init style or upstart.
+- `qubes-core-agent-networking`: Networking support. Useful if the template has to be used for a `sys-net` VM.
+- `qubes-core-agent-network-manager`: Integration for NetworkManager. Useful if the template has to be used for a `sys-net` VM.
+- `qubes-core-agent-dom0-updates`: Script required to handle `dom0` updates. Any template which the VM respondible for 'dom0' updates is based on must contain this package.
```
andrewdavidwong: @marmarek: Any opinion on this list of packages?

marmarek: `qubes-core-agent-qrexec`, `qubes-core-agent-systemd` are installed by default, it may be good to mark it here.
`qubes-core-agent-passwordless-root` - it may be good to point also here: https://www.qubes-os.org/doc/vm-sudo/#replacing-password-less-root-access-with-dom0-user-prompt

There is also (not strictly qubes-core-agent) `pulseaudio-qubes`, for audio support. This is also not installed by default in minimal template.

managing-os/templates/fedora-minimal.md
marmarek: It looks like compressed/uncompressed difference. RPM package is about 650MB, which results in about 2GB disk usage after installation.

managing-os/templates/fedora-minimal.md
FabrizioRomanoGenovese: current-testing is required for qubes-input-proxy-sender. It is worth to point this out, I wasted two hours trying to figure out what was wrong with my sys-usb because of this! :D

managing-os/templates/fedora-minimal.md
FabrizioRomanoGenovese: Clearly a typo, my bad.

managing-os/templates/fedora-minimal.md
FabrizioRomanoGenovese: Yup. 600MB is the amount of stuff dnf has to download. it reaches ~1.5GB when installed, if I remember well.

managing-os/templates/fedora-minimal.md
FabrizioRomanoGenovese: My explanation of what these packages do is bad at least: I figured it out by trial and error for some of them, while for other I just copy-pasted what dnf search told me. Probably it is worth to have someone with a better insight to further elaborate on this.

managing-os/templates/fedora-minimal.md
marmarek: It might be the case shortly after releasing first rc, but now all packages are available in stable repository.

managing-os/templates/fedora-minimal.md
marmarek: Besides the points I listed above, IMO your description is ok.

managing-os/templates/fedora-minimal.md
andrewdavidwong: > current-testing is required for qubes-input-proxy-sender. It is worth to point this out, I wasted two hours trying to figure out what was wrong with my sys-usb because of this! :D

Yes, but my point still stands. It wouldn't be right to tell *every* user who wants to customize `fedora-minimal` to download testing packages.

managing-os/templates/fedora-minimal.md
mossy-nw: Thanks for following up on this -- in R3.2 Qubes manager reports ~2680MiB, which agrees with `[dom0]$ du` -- `private.img` alone is 2.0G.  I don't want to needlessly nitpick, I just want the docs to be clear for folks using, say, older hardware with a 120GB SSD.  Finally, I ran `[dom0]$ qvm-trim-template`, just to see, and it's now 1.7G.  Should folks with limited disk space be trimming templates?

andrewdavidwong: Yes.

PR #509, opened 2018-01-06
install Whonix using salt

adrelanos:
https://github.com/QubesOS/qubes-issues/issues/3447


adrelanos:
Would be cool if @marmarek could have a quick look for review.

PR #510, opened 2018-01-10
Add policy parameter for R4.0

tasket:
Issue https://github.com/QubesOS/qubes-issues/issues/3270

### DIFF
--- a/security/vm-sudo.md
+++ b/security/vm-sudo.md
@@ -104,8 +104,8 @@ this for extra security.**
 
 1. Adding Dom0 "VMAuth" service:
 
-        [root@dom0 /]# echo -n "/usr/bin/echo 1" >/etc/qubes-rpc/qubes.VMAuth
-        [root@dom0 /]# echo -n "\$anyvm dom0 ask,default_target=dom0" \
+        [root@dom0 /]# echo "/usr/bin/echo 1" >/etc/qubes-rpc/qubes.VMAuth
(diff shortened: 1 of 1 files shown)


security/vm-sudo.md
```diff
@@ -105,7 +105,8 @@ this for extra security.**
 1. Adding Dom0 "VMAuth" service:
 
         [root@dom0 /]# echo -n "/usr/bin/echo 1" >/etc/qubes-rpc/qubes.VMAuth
-        [root@dom0 /]# echo -n "\$anyvm dom0 ask" >/etc/qubes-rpc/policy/qubes.VMAuth
+        [root@dom0 /]# echo -n "\$anyvm dom0 ask,default_target=dom0" \
+        >/etc/qubes-rpc/policy/qubes.VMAuth
```
marmarek: I know it was here already, but why `echo -n`, in both cases? Looks unnecessary.

tasket: Don't know why. I'll remove it.

andrewdavidwong:
Will this also work on 3.2? If not, you should create a new section, rather than *replacing* accurate 3.2 instructions with 4.0 instructions.

marmarek:
Should be fine for R3.2 too, see https://github.com/QubesOS/qubes-issues/issues/3270#issuecomment-341479193

andrewdavidwong:
Oh, ok! My mistake for missing that.

PR #512, opened 2018-01-13
Custom "kernel" key in xen.cfg on USB install media

yilmi:
Reviewing the HCL a lot of reporters had to install in legacy mode because UEFI mode would fail. For various reports, they used custom kernel parameters by changing the grub entry. This is frequent for recent hardware or some configurations (dual gpu,...).

With this procedure a user can perform UEFI install with custom kernel parameters

### DIFF
--- a/troubleshooting/uefi-troubleshooting.md
+++ b/troubleshooting/uefi-troubleshooting.md
@@ -10,15 +10,19 @@ Troubleshooting UEFI related problems
 Change installer kernel parameters in UEFI
 ---------------------
 
-If you've installed successfully in legacy mode but had to change some kernel parameters for it to work, you should try installing in UEFI with the same kernel parameters.
+If you've installed successfully in legacy mode but had to change some kernel parameters for it to work, you should try installing in UEFI mode with the same parameters.
 
-Change the `xen.cfg` on a USB media
-
-01. Attach the usb disk, find the EFI partition and mount it
-02. Edit your xen.cfg changing the `kernel` key to add your kernel parameters on the boot entry of your choice
+**Change the xen configuration on a USB media**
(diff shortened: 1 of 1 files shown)


troubleshooting/uefi-troubleshooting.md
```diff
@@ -7,6 +7,19 @@ permalink: /doc/uefi-troubleshooting/
...
+
+If you've installed successfully in legacy mode but had to change some kernel parameters for it to work, you should try installing in UEFI with the same kernel parameters.
+
+Change the `xen.cfg` on a USB media
+
+01. Attach the usb disk, find the EFI partition and mount it
+02. Edit your xen.cfg changing the `kernel` key to add your kernel parameters on the boot entry of your choice
+03. Install using your modified boot entry
+
+You can also update an iso image, use `losetup` to isolate the EFI partition and mount it.
```
marmarek: Not sure what you mean here. `losetup -P`? If so, better name it explicitly.

troubleshooting/uefi-troubleshooting.md
```diff
@@ -7,6 +7,19 @@ permalink: /doc/uefi-troubleshooting/
...
 ========================================
 
+Change installer kernel parameters in UEFI
+---------------------
+
+If you've installed successfully in legacy mode but had to change some kernel parameters for it to work, you should try installing in UEFI with the same kernel parameters.
+
+Change the `xen.cfg` on a USB media
+
+01. Attach the usb disk, find the EFI partition and mount it
```
marmarek: I'd add a hint it is the second partition. Or the one with just "EFI" directory, not all other stuff like "Packages", "isolinux" etc. Generally give some info how to "find the EFI partition".

PR #523, opened 2018-01-22
Improve instructions for randomizing MAC addresses

pgerber:
NetworkManager, by default, uses a connection ID and a per-host random
and secret key to generate `stable` MAC addresses. The intention is to
keep a connection's MAC address stable indefinitely but for it to be
different on every host.

The current instruction mention that "`stable` generates a random
address that persists for each boot session". This is indeed true for
AppVMs using stock TemplateVMs. The reason is that the secret key doesn't
exist in the template and thus is only created when the AppVM starts.
This, however, may not be true for other VMs.

In order to ensure that MACs are always only `stable` during one boot
session, `stable-id`, which is used to generate MACs, can be adjusted.
[NetworkManager's documentation] suggests to use `${CONNECTION}/${BOOT}`
to ensure generated MACs are unique to a boot session and connection.

[NetworkManager's documentation]: https://developer.gnome.org/NetworkManager/stable/nm-settings.html

### DIFF
--- a/privacy/anonymizing-your-mac-address.md
+++ b/privacy/anonymizing-your-mac-address.md
@@ -34,10 +34,11 @@ wifi.scan-rand-mac-address=yes
 [connection]
 wifi.cloned-mac-address=stable
 ethernet.cloned-mac-address=stable
+connection.stable-id=${CONNECTION}/${BOOT}
 ~~~
 
-`stable` generates a random address that persists for each boot session.
-`random` generates a random address each time a link goes up.
+* `stable` in combination with `${CONNECTION}/${BOOT}` generates a random address that persists for each boot session.
+* `random` generates a random address each time a link goes up.
 
 To see all the available configuration options, refer to the man page: `man nm-settings`
 
--- a/privacy/anonymizing-your-mac-address.md
+++ b/privacy/anonymizing-your-mac-address.md
@@ -37,7 +37,7 @@ ethernet.cloned-mac-address=stable
 connection.stable-id=${CONNECTION}/${BOOT}
 ~~~
 
-* `stable` in combination with `${CONNECTION}/${BOOT}` generates a random address that persists for each boot session.
+* `stable` in combination with `${CONNECTION}/${BOOT}` generates a random address that persists until reboot.
 * `random` generates a random address each time a link goes up.
 
 To see all the available configuration options, refer to the man page: `man nm-settings`


tasket:
@pgerber Interesting. So a template-based netvm will keep generating a random address on boot, but a standalone netvm will keep the MAC the same per access point. I just tested this with a debian standalone netvm and that is the case.

There is a question of which behavior is more desirable to the user: After all, the NM default is to randomize then preserve across reboots. But yours is probably the easiest way to make the behavior consistent.

To make randomize-then-preserve consistent may require an entry in bind-dirs. I think '/var/lib/NetworkManager' may be the right path. Maybe we could include this info to give the user a choice about having the address preserved.

tasket:
I verified that NM retains the MAC address between reboots if '/var/lib/NetworkManager' is added to bind-dirs (on a template-based netvm).

As instructions for the doc, the commands would look like:
```
sudo mkdir -p /rw/config/qubes-bind-dirs.d /rw/bind-dirs/var/lib
echo "binds+=( '/var/lib/NetworkManager' )" | sudo tee /rw/config/qubes-bind-dirs.d/50_nm.conf
sudo cp -a /var/lib/NetworkManager /rw/bind-dirs/var/lib
```

@marek I wonder if `network-manager-prepare-conf-dir` should handle this instead?

privacy/anonymizing-your-mac-address.md
```diff
@@ -34,10 +34,11 @@ wifi.scan-rand-mac-address=yes
 [connection]
 wifi.cloned-mac-address=stable
 ethernet.cloned-mac-address=stable
+connection.stable-id=${CONNECTION}/${BOOT}
 ~~~
 
-`stable` generates a random address that persists for each boot session.
-`random` generates a random address each time a link goes up.
+* `stable` in combination with `${CONNECTION}/${BOOT}` generates a random address that persists for each boot session.
```
marmarek: I think it isn't clear that address will be different after VM reboot. This is what will happen, right?
Maybe better write "... persists until reboot"?

pgerber: > I think it isn't clear that address will be different after VM reboot. This is what will happen, right?

Yes, that's what happens.

> Maybe better write "... persists until reboot"?

Agreed, that's clearer and easier to understand.

andrewdavidwong:
If I understand the situation correctly:
 * @pgerber has implemented @marmarek's requested changes, so this PR is now acceptable to merge.
 * However, @tasket has some concerns about the way the content might be further improved.
 * However, it may be a complex matter to work out what those further improvements will be.
Rather than delaying the merge of this PR (and allowing the perfect to be the enemy of the good), I'm going to merge it now and suggest that @tasket open an issue to pursue those further improvements.

PR #525, opened 2018-01-22
Update fedora-minimal.md

mossy-nw:
for R4.0, removed incorrect `sudo` package install, added `qubes-usb-proxy` for usb qube source/sink to Qubes R4.0 (not sure if this should replace `qubes-input-proxy-sender` for all versions -- see [qubes-users thread here](https://groups.google.com/forum/#!topic/qubes-users/ltTbP7pFwdA)), and added `sys-firewall` to dom0 update domain for clarity.


andrewdavidwong:
> for R4.0, removed incorrect sudo package install

Just out of curiosity, what was incorrect about this?

mossy-nw:
I don't recall where I got that initial advice (`apt install sudo`) --  the qubes-users list?  Perhaps that approach worked at one time but someone contributed to `fedora-minimal.md` the following:

 > `qubes-core-agent-passwordless-root`, `polkit`: By default the `fedora-26-minimal` template doesn’t have passwordless root. These two packages fix the situation.

So I realized that `sudo` is indeed installed, the problem is only sudo has an unknown password.  In any case the quoted advice is sufficient to get passwordless sudo working, so I deleted that line you quoted.

I think this also means qubes-issue #3157 can be closed?  I'll comment there.  Thanks, @andrewdavidwong !

PR #530, opened 2018-01-25
formatting and rewording enhancements

adrelanos:


adrelanos:
These enhancement suggestions are based on working with two users in person explaining how to do this. Users are always looking out for nice gray boxes where they can copy and paste from. As I understand, each gray box represents a step they have to do in their mind.

Copying a `sudo nano something` command from within the running text makes users miss running that command. Each step to do should be atomic. If/after this gets hopefully merged, I would like to add another revision on top where I give each step its own distinctive number.

Somehow github makes the diff look difficult. However, no actual instructions were changed. Only the wording and formatting was improved.

configuration/vpn.md
```diff
@@ -68,29 +68,168 @@ This method is more involved than the one above, but has anti-leak features that
...
+---
+layout: doc
+title: VPN
+permalink: /doc/vpn/
+redirect_from:
+- /doc/privacy/vpn/
+- /en/doc/vpn/
+- /doc/VPN/
+- /wiki/VPN/
+---
```
marmarek: Something went really wrong here, you've duplicated part of this file...

adrelanos:
Sorry!

Fixed. Please review.

configuration/vpn.md
```diff
@@ -68,29 +68,74 @@ This method is more involved than the one above, but has anti-leak features that
...
-    ```
+Copy your VPN config files to `/rw/config/vpn`. Your VPN config file should be named `openvpn-client.ovpn`) so you can use the scripts below as is without modification. Otherwise you would have to replace the file name. `openvpn-client.ovpn` contents:
+
+Files accompanying the main config such as `*.crt` and `*.pem` should also go to `/rw/config/vpn` folder.
+
+Files referenced in `openvpn-client.ovpn` should not use absolute paths such as `/etc/...`.
+
+The VPN scripts here are intended to work with commonly used `tun` interfaces, whereas `tap` mode is untested.
+
+Also, the config should route all traffic through your VPN's interface after a connection is created; For openvpn the directive for this is `redirect-gateway def1`. For OpenVPN.
```
marmarek: "for openvpn" is duplicated here.

configuration/vpn.md
```diff
@@ -68,29 +68,74 @@ This method is more involved than the one above, but has anti-leak features that
...
+
+       sudo nano /rw/config/vpn/openvpn-client.ovpn
+       
+Make sure it already includes or add.
```
marmarek: IMO better use colon here instead of dot

configuration/vpn.md
```diff
@@ -68,29 +68,74 @@ This method is more involved than the one above, but has anti-leak features that
...
+
+      auth-user-pass pass.txt
+      
+ Save file `/rw/config/vpn/openvpn-client.ovpn`.
+ 
+ Make sure a `/rw/config/vpn/pass.txt` file actually exists.
+ 
+      sudo nano /rw/config/vpn/pass.txt
+ 
+ Add.
```
marmarek: And here.

configuration/vpn.md
```diff
@@ -68,29 +68,74 @@ This method is more involved than the one above, but has anti-leak features that
...
+
+Also, the config should route all traffic through your VPN's interface after a connection is created; For openvpn the directive for this is `redirect-gateway def1`. For OpenVPN.
+
+       sudo nano /rw/config/vpn/openvpn-client.ovpn
+       
+Make sure it already includes or add.
+
+       redirect-gateway def1
+
+The VPN client may not be able to prompt you for credentials when connecting to the server. Create a file in the `/rw/config/vpn` folder with your credentials and using a directive. For example for OpenVPN, add.
```
marmarek: And here.

configuration/vpn.md
```diff
@@ -68,29 +68,74 @@ This method is more involved than the one above, but has anti-leak features that
...
+ Make sure a `/rw/config/vpn/pass.txt` file actually exists.
+ 
+      sudo nano /rw/config/vpn/pass.txt
+ 
+ Add.
+ 
+ ```
+ username
+ password
+ ```
```
marmarek: Use indented blocks, as everywhere else.

adrelanos:
All fixed.

PR #532, opened 2018-01-26
Update software-update-dom0.md

awokd:
Dnf vs. Yum, add EFI update

### DIFF
--- a/common-tasks/software-update-dom0.md
+++ b/common-tasks/software-update-dom0.md
@@ -58,10 +58,14 @@ Of course, command line tools are still available for accomplishing various upda
     sudo qubes-dom0-update package-version
     ~~~
 
-    Yum will say that there is no update, but the package will nonetheless be downloaded to dom0.
+    Dnf/Yum will say that there is no update, but the package will nonetheless be downloaded to dom0.
 
-2.  Downgrade the package:
+2.  Downgrade the package (R4.0+):
 
+    ~~~
+    sudo dnf downgrade package-version
+    ~~~
+    R3.2 and earlier
     ~~~
     sudo yum downgrade package-version
     ~~~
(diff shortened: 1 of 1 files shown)


awokd:
Good catch. Removed the yum lines- that's going to simplify edits on the rest of the docs too.

common-tasks/software-update-dom0.md
```diff
@@ -124,8 +124,16 @@ is needed for the VMs. (Note that the following example enables the unstable rep
...
 ~~~
 
-Rebuild grub config.
+If the update process does not automatically do it (you should see it mentioned in the CLI output
+from the update command), you may need to manually rebuild the EFI or grub config depending on which
+your system uses.
 
+EFI (Replace the file names with the correct versions for your updated kernel)
+~~~
+sudo /usr/bin/dracut -f /boot/efi/EFI/qubes/initramfs-4.4.31-11.pvops.qubes.x86_64.img 4.4.31-11.pvops.qubes.x86_64
```
andrewdavidwong: @marmarek: Can you confirm this command?

marmarek: The command is correct, but does something different - rebuild initramfs on system started with EFI.
There is also equivalent command for non-EFI system:
```
sudo /usr/bin/dracut -f /boot/initramfs-4.4.31-11.pvops.qubes.x86_64.img 4.4.31-11.pvops.qubes.x86_64
```
Both should happen automatically during kernel installation and I'm not aware of any cases when it doesn't.
The grub case applies only to legacy (non-EFI) case and problem does happen from time to time.

awokd: That dracut command has helped me and others (https://mail-archive.com/qubes-users@googlegroups.com/msg17978.html) when running out of space on the EFI partition during a kernel upgrade. Should I remove from here and add it (plus the grub line) to the FAQ?

marmarek: I think it's ok to leave it here.

awokd:
I think this one is ready to go, then.

PR #538, opened 2018-01-28
Dispvm 4.0 updates

awokd:
Sourced primarily from https://github.com/QubesOS/qubes-issues/issues/2253 and https://groups.google.com/forum/?_escaped_fragment_=topic/qubes-devel/UGh8NDdkrXo#!topic/qubes-devel/UGh8NDdkrXo

### DIFF
--- a/common-tasks/dispvm.md
+++ b/common-tasks/dispvm.md
@@ -11,37 +11,54 @@ redirect_from:
 Disposable VMs (DispVMs)
 ========================
 
-A Disposable VM (DispVM) is a lightweight VM that can be created quickly and will disappear when closed. 
-Disposable VMs are usually created in order to host a single application, like a viewer, editor, or web browser. 
-Changes made to a file opened in a Disposable VM are passed back to the originating VM. 
-This means that you can safely work with untrusted files without risk of compromising your other VMs. 
-DispVMs can be created either directly from Dom0 or from within AppVMs. 
-Once a DispVM has been created it will appear in Qubes VM Manager with the name "dispX". 
+A Disposable VM (DispVM) is a lightweight VM that can be created quickly and will disappear when closed.
+Disposable VMs are usually created in order to host a single application, like a viewer, editor, or web browser.
+
+From inside an AppVM, choosing the `Open in Disposable VM` option on a file will launch a DispVM for just that file.
+Changes made to a file opened in a DispVM are passed back to the originating VM.
+This means that you can safely work with untrusted files without risk of compromising your other VMs.
+DispVMs can be launched either directly from Dom0's Start Menu or terminal window, or from within AppVMs.
+While running, DispVMs will appear in Qubes VM Manager with the name `disp####`.
 
 See [this article](https://blog.invisiblethings.org/2010/06/01/disposable-vms.html) for more on why one would want to use a Disposable VM.
+
+
+DVM Templates
+----------
+
+Similarly to how AppVMs are based on their underlying [TemplateVM](https://www.qubes-os.org/doc/glossary/#templatevm), DispVMs are based on their underlying [DVM Template](https://www.qubes-os.org/doc/glossary/#dvm-template).
+
+On a fresh installation of Qubes, the DVM template is called `fedora-XX-dvm` (where `XX` is the Fedora version of the default TemplateVM). 
 
 Disposable VMs and Networking (R4.0 and later)
 -----------------------------
 
+R4.0 introduces the concept of multiple disposable VM templates, whereas R3.2 was limited to only one.
+You can set any AppVM to have the ability to act as a DVM Template with:
 
-R4.0 introduces the concept of multiple disposable VM templates (R3.2 was limited to one).
-This allows for the creation of multiple differently configured disposable VMs that can be accessed from 
-the Applications menu. Even more types of DispVMs can be created on-the-fly on a per AppVM basis.
-As you can see, this is a very flexible and powerful system for managing your Disposable VMs.
+    qvm-prefs <vmname> template_for_dispvms true
 
-NetVM and firewall rules for Disposable VMs can be set as they can for a normal VM. 
-By default a DispVM will inherit the NetVM and firewall settings of the DispVM Template from which it is built. 
-Thus if an AppVM uses sys-net as its NetVM, but the default system DispVM uses sys-whonix,
-any DispVM launched from this AppVM will have sys-whonix as its NetVM.
-The default system wide DispVM template can be changed with `qubes-prefs default_dispvm`.
-You can change this behaviour for individual VMs: in the Application Menu, open Qube Settings
-for the VM in question and go to the "Advanced" tab. 
+The default system wide DVM template can be changed with `qubes-prefs default_dispvm`.
+By combining the two, choosing `Open in Disposable VM` from inside an AppVM will open the document in a DispVM based on the default DVM template you specified.
+
+You can change this behaviour for individual VMs: in the Application Menu, open Qube Settings for the VM in question and go to the "Advanced" tab. 
 Here you can edit the "Default DispVM" setting to specify which DispVM template will be used to launch DispVMs from that VM.
-Disposable VMs will temporarily appear with the name `disp####`. 
+This can also be changed from the command line with:
 
-A Disposable VM launched from the Start Menu inherits the NetVM and firewall settings of the [DVM Template](https://www.qubes-os.org/doc/glossary/#dvm-template) from which it is built. 
-By default the DVM template is called `fedora-XX-dvm` (where `XX` is the Fedora version of the default TemplateVM). 
-Note that changing the "NetVM" setting for the DVM Template *does* affect the NetVM of DispVMs launched from the Start Menu.
+    qvm-prefs <vmname> default_dispvm <dvmtemplatename>
+
+You can even set an AppVM that has also been configured as a DVM template to use itself, so DispVMs launched from within the AppVM/DVM Template would inherit the same settings.
+
+NetVM and firewall rules for DVM templates can be set as they can for a normal VM. 
+By default a DispVM will inherit the NetVM and firewall settings of the DVM Template on which it is based.
+Launching a DispVM from an AppVM will result in it using the DispVM's network/firewall settings (which default to the DVM template on which it is based).
+Thus if an AppVM uses sys-net as its NetVM, but the default system DispVM uses sys-whonix, any DispVM launched from this AppVM will have sys-whonix as its NetVM.
+
+**Note** The opposite is also true. This means if the default system DispVM uses sys-net, launching a DispVM from inside anon-whonix will result in the DispVM using sys-net.
+
+A Disposable VM launched from the Start Menu inherits the NetVM and firewall settings of the DVM Template on which it is based.
+Note that changing the "NetVM" setting for the system default DVM Template *does* affect the NetVM of DispVMs launched from the Start Menu.
+Different DVM Templates with individual NetVM settings can be added to the Start Menu. 
 
 Disposable VMs and Networking (R3.2 and earlier)
 -----------------------------
@@ -52,8 +69,7 @@ Thus if an AppVM uses sys-net as its NetVM, any DispVM launched from this AppVM 
 You can change this behaviour for individual VMs: in Qubes VM Manager open VM Settings for the VM in question and go to the "Advanced" tab. 
 Here you can edit the "NetVM for DispVM" setting to change the NetVM of any DispVM launched from that VM.
 
-A Disposable VM launched from the Start Menu inherits the NetVM of the [DVM Template](https://www.qubes-os.org/doc/glossary/#dvm-template). 
-By default the DVM template is called `fedora-XX-dvm` (where `XX` is the Fedora version of the default TemplateVM). 
+A Disposable VM launched from the Start Menu inherits the NetVM of the [DVM Template](https://www.qubes-os.org/doc/glossary/#dvm-template).
 As an "internal" VM it is hidden in Qubes VM Manager, but can be shown by selecting "Show/Hide internal VMs". 
 Note that changing the "NetVM for DispVM" setting for the DVM Template does *not* affect the NetVM of DispVMs launched from the Start Menu; only changing the DVM Template's own NetVM does.
 
--- a/common-tasks/dispvm.md
+++ b/common-tasks/dispvm.md
@@ -20,7 +20,31 @@ Once a DispVM has been created it will appear in Qubes VM Manager with the name 
 
 See [this article](https://blog.invisiblethings.org/2010/06/01/disposable-vms.html) for more on why one would want to use a Disposable VM.
 
-Disposable VMs and Networking
+Disposable VMs and Networking (R4.0 and later)
+-----------------------------
+
+
+R4.0 introduces the concept of multiple disposable VM templates (R3.2 was limited to one).
+This allows for the creation of multiple differently configured disposable VMs that can be accessed from 
+the Applications menu (e.g. `fedora-XX-dvm`). Even more types of DispVMs can be created on-the-fly on a per AppVM basis.
+As you can see, this is a very flexible and powerful system for managing your Disposable VMs.
+
+NetVM and firewall rules for Disposable VMs can be set as they can for a normal VM. 
+By default a DispVM will inherit the NetVM and firewall settings of the DispVM Template from which it is built. 
+Thus if an AppVM uses sys-net as its NetVM, but the default system DispVM uses sys-whonix,
+any DispVM launched from this AppVM will have sys-whonix as its NetVM.
+The default system wide DispVM template can be changed with `qubes-prefs default_dispvm`.
+You can change this behaviour for individual VMs: in the Application Menu, open Qube Settings
+for the VM in question and go to the "Advanced" tab. 
+Here you can edit the "Default DispVM" setting to specify which DispVM template will be used to launch DispVMs from that VM.
+Disposable VMs will temporarily appear with the name `disp####`. 
+
+A Disposable VM launched from the Start Menu inherits the NetVM and firewall settings of the [DVM Template](https://www.qubes-os.org/doc/glossary/#dvm-template) from which it is built. 
+By default the DVM template is called `fedora-XX-dvm` (where `XX` is the Fedora version of the default TemplateVM). 
+As an "internal" VM it is hidden in Qubes VM Manager, but can be shown by selecting "Show/Hide internal VMs". 
+Note that changing the "NetVM" setting for the DVM Template *does* affect the NetVM of DispVMs launched from the Start Menu.
+
+Disposable VMs and Networking (R3.2 and earlier)
 -----------------------------
 
 NetVM and firewall rules for Disposable VMs can be set as they can for a normal VM. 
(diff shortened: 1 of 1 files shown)


common-tasks/dispvm.md
```diff
@@ -20,7 +20,31 @@ Once a DispVM has been created it will appear in Qubes VM Manager with the name
...
+Thus if an AppVM uses sys-net as its NetVM, but the default system DispVM uses sys-whonix,
+any DispVM launched from this AppVM will have sys-whonix as its NetVM.
+The default system wide DispVM template can be changed with `qubes-prefs default_dispvm`.
+You can change this behaviour for individual VMs: in the Application Menu, open Qube Settings
+for the VM in question and go to the "Advanced" tab. 
+Here you can edit the "Default DispVM" setting to specify which DispVM template will be used to launch DispVMs from that VM.
+Disposable VMs will temporarily appear with the name `disp####`. 
+
+A Disposable VM launched from the Start Menu inherits the NetVM and firewall settings of the [DVM Template](https://www.qubes-os.org/doc/glossary/#dvm-template) from which it is built. 
+By default the DVM template is called `fedora-XX-dvm` (where `XX` is the Fedora version of the default TemplateVM).
```
marmarek: Those two sentences looks duplicated with the first paragraph.

common-tasks/dispvm.md
```diff
@@ -20,7 +20,31 @@ Once a DispVM has been created it will appear in Qubes VM Manager with the name
...
+any DispVM launched from this AppVM will have sys-whonix as its NetVM.
+The default system wide DispVM template can be changed with `qubes-prefs default_dispvm`.
+You can change this behaviour for individual VMs: in the Application Menu, open Qube Settings
+for the VM in question and go to the "Advanced" tab. 
+Here you can edit the "Default DispVM" setting to specify which DispVM template will be used to launch DispVMs from that VM.
+Disposable VMs will temporarily appear with the name `disp####`. 
+
+A Disposable VM launched from the Start Menu inherits the NetVM and firewall settings of the [DVM Template](https://www.qubes-os.org/doc/glossary/#dvm-template) from which it is built. 
+By default the DVM template is called `fedora-XX-dvm` (where `XX` is the Fedora version of the default TemplateVM). 
+As an "internal" VM it is hidden in Qubes VM Manager, but can be shown by selecting "Show/Hide internal VMs".
```
marmarek: fedora-26-dvm is not marked as "internal" by default in 4.0. Also, Qube Manger in 4.0 no longer hide "internal" VMs.

common-tasks/dispvm.md
awokd: In the first paragraph of the R4.0 section, I was trying to introduce the R4.0 concept of multiple disposable templates vs. R3.2's single. The paragraph you cited is to spell out the differences between R4.0 and R3.2. I'll take out the (e.g.) from the first para to lessen duplication, but is there a better way the first para could be worded?

common-tasks/dispvm.md
```diff
@@ -20,7 +20,30 @@ Once a DispVM has been created it will appear in Qubes VM Manager with the name
...
+
+R4.0 introduces the concept of multiple disposable VM templates (R3.2 was limited to one).
+This allows for the creation of multiple differently configured disposable VMs that can be accessed from 
+the Applications menu. Even more types of DispVMs can be created on-the-fly on a per AppVM basis.
+As you can see, this is a very flexible and powerful system for managing your Disposable VMs.
+
+NetVM and firewall rules for Disposable VMs can be set as they can for a normal VM. 
+By default a DispVM will inherit the NetVM and firewall settings of the DispVM Template from which it is built. 
+Thus if an AppVM uses sys-net as its NetVM, but the default system DispVM uses sys-whonix,
+any DispVM launched from this AppVM will have sys-whonix as its NetVM.
```
andrewdavidwong: Aren't these two sentences contradictory?

andrewdavidwong: Also, I think we need to first explain that an AppVM can be a DVM Template.

awokd: Not contradictory- it's saying that under 4.0, launching a DispVM from an AppVM will result in it using the DispVM's network/firewall settings (which default to the DVM template from which it was built). Compare & contrast with the same section a little further down under 3.2 which has different behaviour. It's definitely hard to follow, but I'm not sure how else to phrase it...

awokd: I could add "in other words" and include that same sentence from my comment, if it helps understand what's going on?

common-tasks/dispvm.md
```diff
@@ -20,7 +20,30 @@ Once a DispVM has been created it will appear in Qubes VM Manager with the name
 
 See [this article](https://blog.invisiblethings.org/2010/06/01/disposable-vms.html) for more on why one would want to use a Disposable VM.
 
-Disposable VMs and Networking
+Disposable VMs and Networking (R4.0 and later)
+-----------------------------
+
+
+R4.0 introduces the concept of multiple disposable VM templates (R3.2 was limited to one).
```
andrewdavidwong: This would be a great place to introduce the term "DVM Template" in this document, with a link to its glossary entry.

common-tasks/dispvm.md
```diff
@@ -20,7 +20,30 @@ Once a DispVM has been created it will appear in Qubes VM Manager with the name
...
+This allows for the creation of multiple differently configured disposable VMs that can be accessed from 
+the Applications menu. Even more types of DispVMs can be created on-the-fly on a per AppVM basis.
```
andrewdavidwong: Most readers won't understand what this means unless they already know how the new system works.

common-tasks/dispvm.md
```diff
@@ -20,7 +20,30 @@ Once a DispVM has been created it will appear in Qubes VM Manager with the name
...
+As you can see, this is a very flexible and powerful system for managing your Disposable VMs.
+
+NetVM and firewall rules for Disposable VMs can be set as they can for a normal VM. 
+By default a DispVM will inherit the NetVM and firewall settings of the DispVM Template from which it is built.
```
andrewdavidwong: "DispVM Template" should be changed to "DVM Template" so that we use consistent terminology throughout the documentation.

andrewdavidwong: Also, "from which it is built" might be a bit misleading. I think it's more common to say, "on which it is based."

common-tasks/dispvm.md
andrewdavidwong: We need to clearly explain the difference between the "based on" and "launched from" relationships, and we need provide examples for the other main cases.

In your example, I launch a DispVM from an AppVM using `sys-net`, but the DispVM doesn't use `sys-net`. Instead, it uses `sys-whonix`, because it's based on my default DVM Template, which uses `sys-whonix`. I can imagine a user having several questions about this:
 - Will the same pattern of behavior occur if I launch a DispVM from `anon-whonix` while my default DVM Template uses `sys-net`? That would be incredibly dangerous.
 - Why was my DispVM based on the default DVM Template rather than the AppVM from which I launched it?
 - How do I base the DispVM on the AppVM instead?

Some of these questions are answered below, but in a way that at least partially contradicts what you wrote above (e.g., because it depends on the settings explained below).

common-tasks/dispvm.md
```diff
@@ -20,7 +20,30 @@ Once a DispVM has been created it will appear in Qubes VM Manager with the name
...
+This allows for the creation of multiple differently configured disposable VMs that can be accessed from 
+the Applications menu. Even more types of DispVMs can be created on-the-fly on a per AppVM basis.
+As you can see, this is a very flexible and powerful system for managing your Disposable VMs.
+
+NetVM and firewall rules for Disposable VMs can be set as they can for a normal VM. 
+By default a DispVM will inherit the NetVM and firewall settings of the DispVM Template from which it is built. 
+Thus if an AppVM uses sys-net as its NetVM, but the default system DispVM uses sys-whonix,
+any DispVM launched from this AppVM will have sys-whonix as its NetVM.
+The default system wide DispVM template can be changed with `qubes-prefs default_dispvm`.
+You can change this behaviour for individual VMs: in the Application Menu, open Qube Settings
```
andrewdavidwong: Grammatically, "this behavior" seems to refer to "the default system wide DispVM template," which is problematic for two reasons:

1. "The default system wide DispVM template" is not a behavior.
2. You can't actually change it in the Qube Settings of individual VMs.

common-tasks/dispvm.md
```diff
@@ -20,7 +20,30 @@ Once a DispVM has been created it will appear in Qubes VM Manager with the name
...
+for the VM in question and go to the "Advanced" tab. 
+Here you can edit the "Default DispVM" setting to specify which DispVM template will be used to launch DispVMs from that VM.
```
andrewdavidwong: Out of curiosity, can you set it to itself (i.e., make it so that all DispVMs launched from an AppVM are based on that AppVM)? If so, it might be worth mentioning this case.

common-tasks/dispvm.md
```diff
@@ -20,7 +20,30 @@ Once a DispVM has been created it will appear in Qubes VM Manager with the name
...
+Disposable VMs will temporarily appear with the name `disp####`.
```
andrewdavidwong: What does this mean? It almost sounds as if they momentarily appear with that strange name right after you edit the Qube Settings, but I don't think that's the case.

If all you mean is that DispVMs are themselves disposable, please don't say it this way.

common-tasks/dispvm.md
```diff
@@ -20,7 +20,30 @@ Once a DispVM has been created it will appear in Qubes VM Manager with the name
...
+Thus if an AppVM uses sys-net as its NetVM, but the default system DispVM uses sys-whonix,
+any DispVM launched from this AppVM will have sys-whonix as its NetVM.
+The default system wide DispVM template can be changed with `qubes-prefs default_dispvm`.
+You can change this behaviour for individual VMs: in the Application Menu, open Qube Settings
+for the VM in question and go to the "Advanced" tab. 
+Here you can edit the "Default DispVM" setting to specify which DispVM template will be used to launch DispVMs from that VM.
+Disposable VMs will temporarily appear with the name `disp####`. 
+
+A Disposable VM launched from the Start Menu inherits the NetVM and firewall settings of the [DVM Template](https://www.qubes-os.org/doc/glossary/#dvm-template) from which it is built. 
+By default the DVM template is called `fedora-XX-dvm` (where `XX` is the Fedora version of the default TemplateVM).
```
andrewdavidwong: Do you mean "the default DVM Template" rather than "by default the DVM Template"? Since there can be more than one DVM Template now, there's no "*the* DVM Template" anymore, though there can be "the *default* DVM Template."

common-tasks/dispvm.md
```diff
@@ -20,7 +20,30 @@ Once a DispVM has been created it will appear in Qubes VM Manager with the name
...
+any DispVM launched from this AppVM will have sys-whonix as its NetVM.
+The default system wide DispVM template can be changed with `qubes-prefs default_dispvm`.
+You can change this behaviour for individual VMs: in the Application Menu, open Qube Settings
+for the VM in question and go to the "Advanced" tab. 
+Here you can edit the "Default DispVM" setting to specify which DispVM template will be used to launch DispVMs from that VM.
+Disposable VMs will temporarily appear with the name `disp####`. 
+
+A Disposable VM launched from the Start Menu inherits the NetVM and firewall settings of the [DVM Template](https://www.qubes-os.org/doc/glossary/#dvm-template) from which it is built. 
+By default the DVM template is called `fedora-XX-dvm` (where `XX` is the Fedora version of the default TemplateVM). 
+Note that changing the "NetVM" setting for the DVM Template *does* affect the NetVM of DispVMs launched from the Start Menu.
```
andrewdavidwong: Before moving on to 3.2, it would be good to say a bit about how having multiple dedicated DVM Templates works.

awokd:
>  Will the same pattern of behavior occur if I launch a DispVM from anon-whonix while my default DVM Template uses sys-net? That would be incredibly dangerous.

The answer to this is "yes", by design (and from a test I just did to confirm.)

>    Why was my DispVM based on the default DVM Template rather than the AppVM from which I launched it?

Because that's R4.0's design? Not sure how you want me to answer this one...

>    How do I base the DispVM on the AppVM instead?

I can answer this one! 

> Some of these questions are answered below, but in a way that at least partially contradicts what you wrote above (e.g., because it depends on the settings explained below).

I see your point on that "behaviour" sentence, but maybe I'm being slow because I'm not seeing any other settings/answers in your comments that contradict what I wrote? I'll go through and attempt to address one by one, maybe it will fix itself.

awokd:
I need to do a fresh install to confirm anon-whonix's default settings, unless either of you know already. Can't recall what its `default_dispvm` defaults to. Tried to address the other points raised.

common-tasks/dispvm.md
```diff
@@ -11,16 +11,56 @@ redirect_from:
...
+----------
+
+Similarly to how AppVMs are based on their underlying [TemplateVM](https://www.qubes-os.org/doc/glossary/#templatevm), DispVMs are based on their underlying [DVM Template](https://www.qubes-os.org/doc/glossary/#dvm-template).
+
+On a fresh installation of Qubes, the DVM template is called `fedora-XX-dvm` (where `XX` is the Fedora version of the default TemplateVM). 
+
+Disposable VMs and Networking (R4.0 and later)
+-----------------------------
+
+R4.0 introduces the concept of multiple disposable VM templates, whereas R3.2 was limited to only one.
```
andrewdavidwong: Why shift from "DVM Templates" to "disposable VM templates" here? That just invites confusion, because now it's not clear whether you're talking about the thing you just introduced, or something slightly different.

andrewdavidwong:
>>  Will the same pattern of behavior occur if I launch a DispVM from anon-whonix while my default DVM Template uses sys-net? That would be incredibly dangerous.
>>
> The answer to this is "yes", by design (and from a test I just did to confirm.)

Ok, that must be an oversight. Created https://github.com/QubesOS/qubes-issues/issues/3561 to address it.

>>    Why was my DispVM based on the default DVM Template rather than the AppVM from which I launched it?
>>
> Because that's R4.0's design? Not sure how you want me to answer this one...

You just did! The key is to explain how the behavior has changed and anticipate things that might be counterintuitive to the reader.

>>    How do I base the DispVM on the AppVM instead?
>>
> I can answer this one! 

:slightly_smiling_face: 

>> Some of these questions are answered below, but in a way that at least partially contradicts what you wrote above (e.g., because it depends on the settings explained below).
>>
> I see your point on that "behaviour" sentence, but maybe I'm being slow because I'm not seeing any other settings/answers in your comments that contradict what I wrote? I'll go through and attempt to address one by one, maybe it will fix itself.

Ok, thanks! Maybe that was the only one. I'll check to see whether I can find any others.

andrewdavidwong:
> I need to do a fresh install to confirm anon-whonix's default settings, unless either of you know already. Can't recall what its default_dispvm defaults to. Tried to address the other points raised.

Ok, pointed out the need for confirmation on the issue.

andrewdavidwong:
Looks like there's a small conflict that needs to be resolved. (Not sure how that happened.)

common-tasks/dispvm.md
```diff
@@ -11,16 +11,56 @@ redirect_from:
...
 See [this article](https://blog.invisiblethings.org/2010/06/01/disposable-vms.html) for more on why one would want to use a Disposable VM.
 
-Disposable VMs and Networking
+
+DVM Templates
+----------
+
+Similarly to how AppVMs are based on their underlying [TemplateVM](https://www.qubes-os.org/doc/glossary/#templatevm), DispVMs are based on their underlying [DVM Template](https://www.qubes-os.org/doc/glossary/#dvm-template).
+
+On a fresh installation of Qubes, the DVM template is called `fedora-XX-dvm` (where `XX` is the Fedora version of the default TemplateVM).
```
andrewdavidwong: How about "the default DVM Template," since there can be more than one? (Also note the capitalization.)

common-tasks/dispvm.md
```diff
@@ -11,16 +11,56 @@ redirect_from:
...
+R4.0 introduces the concept of multiple disposable VM templates, whereas R3.2 was limited to only one.
+You can set any AppVM to have the ability to act as a DVM Template with:
+
+    qvm-prefs <vmname> template_for_dispvms true
+
+The default system wide DVM template can be changed with `qubes-prefs default_dispvm`.
+By combining the two, choosing `Open in Disposable VM` from inside an AppVM will open the document in a DispVM based on the default DVM template you specified.
+
+You can change this behaviour for individual VMs: in the Application Menu, open Qube Settings for the VM in question and go to the "Advanced" tab. 
+Here you can edit the "Default DispVM" setting to specify which DispVM template will be used to launch DispVMs from that VM.
```
andrewdavidwong: "DispVM template" should be "DVM Template."

common-tasks/dispvm.md
```diff
@@ -11,16 +11,56 @@ redirect_from:
...
+This can also be changed from the command line with:
+
+    qvm-prefs <vmname> default_dispvm <dvmtemplatename>
+
+You can even set an AppVM that has also been configured as a DVM template to use itself, so DispVMs launched from within the AppVM/DVM Template would inherit the same settings.
+
+NetVM and firewall rules for DVM templates can be set as they can for a normal VM. 
+By default a DispVM will inherit the NetVM and firewall settings of the DVM Template on which it is based.
+Launching a DispVM from an AppVM will result in it using the DispVM's network/firewall settings (which default to the DVM template on which it is based).
```
andrewdavidwong: Is it really meaningful to say that the DispVM uses the DispVM's network and firewall settings? It doesn't have those settings before it's created. It has to get them from somewhere. If it gets them from the DVM Template, then the DispVM is really using the DVM Template's network and firewall settings, not its own.

common-tasks/dispvm.md
```diff
@@ -92,8 +137,8 @@ However, it is possible to start an arbitrary application in a DispVM directly f
 Customizing Disposable VMs
 --------------------------
 
-You can change the template used to generate the Disposable VM, and change settings used in the Disposable VM savefile. 
-These changes will be reflected in every new Disposable VM. 
+You can change the template used to generate the Disposable VMs, and change settings used in the Disposable VM savefile. 
+These changes will be reflected in every new Disposable VM spawned from that template.
```
andrewdavidwong: Is "spawned from" different from "launched from" and "based on"? "Spawned from" suggests that "that template" (a DVM Template?) is the parent, and the DispVM is the child, which is usually how we describe the relationship between TemplateVMs and TemplateBasedVMs. So, I'm guessing that "spawned from" is synonymous with "based on," and both are distinct from "launched from."

common-tasks/dispvm.md
awokd: It's meaningful because they can be changed after the fact. Will add somewhere.

awokd: ... but after testing I'm not sure it takes immediate effect, which would make it useless information.
 What does "NetVM and firewall rules for Disposable VMs can be set as they can for a normal VM." mean in the R3.2 section?

andrewdavidwong: > What does "NetVM and firewall rules for Disposable VMs can be set as they can for a normal VM." mean in the R3.2 section?

I interpret that as meaning that you can chose a NetVM and add or remove firewall rules for a DispVM in the same way you can for a regular TemplateBasedVM.

awokd:
>>    What does "NetVM and firewall rules for Disposable VMs can be set as they can for a normal VM." mean in the R3.2 section?
> I interpret that as meaning that you can chose a NetVM and add or remove firewall rules for a DispVM in the same way you can for a regular TemplateBasedVM.

OK; I removed the equivalent I wrote for R4.0 with my last commit because it doesn't make sense there then. Hopefully that's everything!

andrewdavidwong:
Thanks!

PR #539, opened 2018-01-28
Refactor USB.md into two documents

awokd:
This splits the usb.md document into two in preparation for
updating them with R4.0 information. It was already lengthy,
and expanding it with R4.0 would have made it worse. The "common tasks" of
attaching and detaching USB drives and miscellaneous
devices are left in common-tasks/usb.md. The rarely used items
such as creation/deletion of a USB qube and attaching mouse and
keyboard are moved to configuration/usb-qube.md. Finally,
the master document list is updated to reflect this split.

### DIFF
--- a/common-tasks/usb.md
+++ b/common-tasks/usb.md
@@ -11,195 +11,15 @@ redirect_from:
 - /en/doc/external-device-mount-point/
 - /doc/ExternalDeviceMountPoint/
 - /wiki/ExternalDeviceMountPoint/
-- /doc/usbvm/
-- /en/doc/usbvm/
-- /doc/USBVM/
-- /wiki/USBVM/
-- /doc/sys-usb/
 ---
 
 Using and Managing USB Devices
 ==============================
 
-Creating and Using a USB qube
------------------------------
-
-**Warning:** This has the potential to prevent you from connecting a keyboard to Qubes via USB. There are problems with doing this in an encrypted install (LUKS). If you find yourself in this situation, see this [issue][2270-comm23].
-
-Connecting an untrusted USB device to dom0 is a security risk since dom0,
-like almost every OS, reads partition tables automatically. The whole
-USB stack is put to work to parse the data presented by the USB device in order
-to determine if it is a USB mass storage device, to read its configuration, etc.
-This happens even if the drive is then assigned and mounted in another qube.
-
-To avoid this risk, it is possible to prepare and utilize a USB qube.
-
-A USB qube acts as a secure handler for potentially malicious USB devices,
-preventing them from coming into contact with dom0 (which could otherwise be
-fatal to the security of the whole system). With a USB qube, every time you
-connect an untrusted USB drive to a USB port managed by that USB controller, you
-will have to attach it to the qube in which you wish to use it (if different
-from the USB qube itself), either by using Qubes VM Manager or the command line
-(see instructions above). 
-You can create a USB qube using the management stack by performing the following
-steps as root in dom0:
-
- 1. Enable `sys-usb`:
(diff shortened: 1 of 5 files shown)


awokd:
Adjusting linked document shortcuts to new location

andrewdavidwong:
We used to have multiple USB documents like this, then we decided to combine them in order to have a single page devoted to all USB-related topics. I'm not sure it's a good idea to go back on that now. I have a feeling that one day we'll decide we have "too many USB pages" and consolidate them all again. We should avoid this kind of flip-flopping.

@awokd, it sounds like your primary motivation for splitting the pages is length:
>  It was already lengthy, and expanding it with R4.0 would have made it worse.

I don't see that this is a sufficient reason to split the pages. It seems to me that there are good reasons for the page to be long:

 * There's a lot of important information to convey.
 * All the information is about the same topic (USB).
 * Most readers should see all the information.

The most debatable of these points is the last one. Is it true that most readers should see all the information you've moved to `usb-qube`? Well, it depends. Users who already have a USB qube don't need to see the steps required to create one. So, if most users start out with one, most users won't need to see that. But what about the security warning about USB input devices? Yes, probably every Qubes user should be aware of that.

There were other motivations for previously unifying the pages. For example, it's easier for readers to find and link to the USB information if it's all in one place. It's easier to avoid having duplicate information on the pages (which is especially bad if some of it goes stale, because then we have inconsistent docs). It's easier to maintain, because we know there's only one page that requires updating when USB things change.

I'm not saying that we should *never* split this (or any) document up. I just want to make sure that the benefits are really worth the costs, *especially* since we made the opposite move not too long ago. In this case, it's unclear to me that just having two shorter documents instead of one longer one (with the total size remaining the same) is enough of a benefit.

@marmarek, any thoughts? IIRC, you were the one who suggested having a single USB document in the first place.

andrewdavidwong:
@marmarek: Review request is for the question above, not for the actual content, which is mostly just existing content that has moved.

awokd:
Yes, length primarily (it was going to make even editing annoying) and because it's in the "common tasks" section. Creating a USB qube or deciding how to connect a USB keyboard are not common tasks, maybe done once a system if at all. Can link to the warning in the blurb at the top.
Otherwise, if y'all want to stick with the one page, how about moving the rarely used stuff to the bottom?

andrewdavidwong:
> Yes, length primarily (it was going to make even editing annoying) and because it's in the "common tasks" section. Creating a USB qube or deciding how to connect a USB keyboard are not common tasks, maybe done once a system if at all.

I think you and I have a different understanding of "common task." To me, if most Qubes users perform a task at least once (or would choose to perform it if they knew about it), then it's a common task. It doesn't have to be a task that they perform repeatedly.

(Of course, now that USB qubes are created automatically by the installer, that may not be a common task anymore. I don't know how often they get manually created now.)

> Can link to the warning in the blurb at the top.
> Otherwise, if y'all want to stick with the one page, how about moving the rarely used stuff to the bottom?

That would be fine with me.

awokd:
ok

marmarek:
> @marmarek, any thoughts? IIRC, you were the one who suggested having a single USB document in the first place.

Yes, I think we should keep it this way. And organize the page to make it easy to navigate - first common tasks (use USB devices), then advanced, less common tasks (manually creating USB qube etc).
Having multiple USB related pages will make it hard finding information. For example where to look for "USB keyboard" documentation? "USB devices"? Setting up "USB qube"? Somewhere else?

> Otherwise, if y'all want to stick with the one page, how about moving the rarely used stuff to the bottom?

+1

PR #543, opened 2018-01-29
Update DispVM customization guide for Qubes 4.0

marmarek:
Complementary to #538

@andrewdavidwong I've shifted structure one level down to add "Qubes 4.0" and "Qubes 3.2" sections, but this made ToC less usable (only those 4.0/3.2 entries there). Any idea how to improve this? I guess this will be common problem...

### DIFF
--- a/customization/dispvm-customization.md
+++ b/customization/dispvm-customization.md
@@ -15,7 +15,7 @@ Disposable VM Customization
 Qubes 4.0
 ----------
 
-Disposable VM in Qubes 4.0 can be based on any template-based AppVM. You can also choose to use different AppVMs for different Disposable VMs. To prepare AppVM to be a base for Disposable VM, you need to set `template_for_dispvms` property, for example:
+Disposable VM in Qubes 4.0 can be based on any TemplateBasedVM. You can also choose to use different AppVMs for different Disposable VMs. To prepare AppVM to be a base for Disposable VM, you need to set `template_for_dispvms` property, for example:
 
     [user@dom0 ~]$ qvm-prefs fedora-26-dvm template_for_dispvms True
 
@@ -25,13 +25,14 @@ Additionally, if you want to have menu entries for starting applications in Disp
 
 ### Creating new Disposable VM base AppVM ###
 
-You may want to use multiple AppVMs for different Disposable VMs. To create new one, lets say `custom-dvm`, based on `debian-9` template, use following commands:
+In Qubes 4.0, you're no longer restricted to a single DVM Template. Instead, you can create as many as you want. Whenever you start a new Disposable VM, you can choose to base it on whichever DVM Template you like.
+To create new DVM Template, lets say `custom-dvm`, based on `debian-9` template, use following commands:
 
     [user@dom0 ~]$ qvm-create --template debian-9 --label red custom-dvm
     [user@dom0 ~]$ qvm-prefs custom-dvm template_for_dispvms True
     [user@dom0 ~]$ qvm-features custom-dvm appmenus-dispvm 1
 
-Additionally you may want to set it as default Disposable VM base:
+Additionally you may want to set it as default DVM Template:
 
     [user@dom0 ~]$ qubes-prefs default_dispvm custom-dvm
 
@@ -51,7 +52,7 @@ It is possible to change the settings for each new Disposable VM (DispVM). This 
 2.  Change the qube's settings and/or applications, as desired. Some examples of changes you may want to make include:
     -   Changing Firefox's default startup settings and homepage.
     -   Changing Nautilus' default file preview settings.
-    -   Changing the DispVM's default NetVM. For example, you may wish to set the NetVM to "none." Then, whenever you start a new DispVM, you can choose your desired ProxyVM manually (by changing the newly-started DipsVMs settings). This is useful if you sometimes wish to use a DispVM with a TorVM, for example. It is also useful if you sometimes wish to open untrusted files in a network-disconnected DispVM.
+    -   Changing the DispVM's default NetVM. For example, you may wish to set the NetVM to "none." Then, whenever you start a new DispVM, you can choose your desired ProxyVM manually (by changing the newly-started DipsVMs settings). This is useful if you sometimes wish to use a DispVM with a Whonix Gateway, for example. It is also useful if you sometimes wish to open untrusted files in a network-disconnected DispVM.
 
 4.  Shutdown the qube (either by `poweroff` from qube's terminal, or `qvm-shutdown` from dom0 terminal).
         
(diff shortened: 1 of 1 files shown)


marmarek:
Also, I use "DispVM base AppVM" term to avoid common confusion about using TemplateVM as a template for DispVM. But our glossary define "DVM Template". I'm not really sure which one is better.

andrewdavidwong:
@marmarek:
> I've shifted structure one level down to add "Qubes 4.0" and "Qubes 3.2" sections, but this made ToC less usable (only those 4.0/3.2 entries there). Any idea how to improve this? I guess this will be common problem...

Can we change the ToC code to include the next level down?

> Also, I use "DispVM base AppVM" term to avoid common confusion about using TemplateVM as a template for DispVM. But our glossary define "DVM Template". I'm not really sure which one is better.

I prefer "DVM Template." See https://github.com/QubesOS/qubes-issues/issues/2486.

marmarek:
> I prefer "DVM Template." See QubesOS/qubes-issues#2486.

Hmm, DVM is even more obscure than DispVM, see #2671...

andrewdavidwong:
Well, "DVM" just stands for "Disposable Virtual Machine," so I would be fine with any of the following:

 * DispVM Template
 * Disposable VM Template
 * Disposable qube template

marmarek:
> Can we change the ToC code to include the next level down?

Not sure. In theory https://github.com/dafi/tocmd-generator (source of [jquery.toc.js](https://github.com/QubesOS/qubesos.github.io/blob/master/js/jquery.toc.js)) should support two levels of ToC, but the one we use seems to support only one... Changing `h1` and `h2` according to latest commits in original repo doesn't fix this - there is still only one level, just different one. I'm not an expert in JS.

andrewdavidwong:
Me neither. Do we know anyone who can help?

Alternatively, we could use different upstream ToC code. The single heading level was a somewhat annoying limitation even before the 3.2/4.0 issue.

marmarek:
> Me neither. Do we know anyone who can help?

@bnvk ?



> Alternatively, we could use different upstream ToC code.

That's also an option.

customization/dispvm-customization.md
```diff
@@ -10,10 +10,64 @@ redirect_from:
...
 Disposable VM Customization
-====================
+============================
 
-Changing the DVM Template
--------------------------
+Qubes 4.0
+----------
+
+Disposable VM in Qubes 4.0 can be based on any template-based AppVM. You can also choose to use different AppVMs for different Disposable VMs. To prepare AppVM to be a base for Disposable VM, you need to set `template_for_dispvms` property, for example:
```
andrewdavidwong: We should use "TeplateBasedVM" instead of "template-based AppVM."

customization/dispvm-customization.md
```diff
@@ -10,10 +10,64 @@ redirect_from:
...
+
+    [user@dom0 ~]$ qvm-prefs fedora-26-dvm template_for_dispvms True
+
+Additionally, if you want to have menu entries for starting applications in Disposable VM based on this AppVM (instead of in the AppVM itself), you can achieve it with `appmenus-dispvm` feature:
+
+    [user@dom0 ~]$ qvm-features fedora-26-dvm appmenus-dispvm 1
+
+### Creating new Disposable VM base AppVM ###
+
+You may want to use multiple AppVMs for different Disposable VMs. To create new one, lets say `custom-dvm`, based on `debian-9` template, use following commands:
```
andrewdavidwong: > You may want to use multiple AppVMs for different Disposable VMs.

What does this mean, exactly?

Does it mean that I can start Disp1 based on AppVM1 and Disp2 based on AppVM2, rather than being forced to start both Disp1 and Disp2 based on AppVM1?

If that's what it means, I think it would be clearer to say:

"You may want to use different AppVMs for different Disposable VMs."

andrewdavidwong: It might be even better to say something like:

"In Qubes 4.0, you're no longer restricted to  a single DVM Template. Instead, you can create as many as you want. Whenever you start a new Disposable VM, you can choose to base it on whichever DVM Template you like."

customization/dispvm-customization.md
```diff
@@ -10,10 +10,64 @@ redirect_from:
...
+    [user@dom0 ~]$ qubes-prefs default_dispvm custom-dvm
+
+The above default is used whenever a qube request starting a new Disposable VM and do not specify which one (for example `qvm-open-in-dvm` tool). This can be also set in qube settings and will affect service calls from that qube. See [qrexec documentation](/doc/qrexec3/#extra-keywords-available-in-qubes-40-and-later) for details.
+
+If you wish to use the `fedora-minimal` template as a DVM Template, see the "DVM Template" use case under [fedora-minimal customization](/doc/templates/fedora-minimal/#customization).
+
+
+### Customization of Disposable VM ###
+
+It is possible to change the settings for each new Disposable VM (DispVM). This can be done by customizing the base AppVM:
```
andrewdavidwong: > Disposable VM (DispVM)

This should be introduced at the beginning, not the middle.

customization/dispvm-customization.md
```diff
@@ -10,10 +10,64 @@ redirect_from:
...
+
+1.  Start a terminal in the `fedora-26-dvm` qube (or another base for DispVM) by running the following command in a dom0 terminal. (If a qube have `appmenus-dispvm` feature set, there is no menu entry to start applications directly in it, instead of Disposable VM based on it).
```
andrewdavidwong: This sentence confuses me. I'm not sure what it means.

customization/dispvm-customization.md
```diff
@@ -10,10 +10,64 @@ redirect_from:
...
+
+        [user@dom0 ~]$ qvm-run -a fedora-26-dvm gnome-terminal
+
+2.  Change the qube's settings and/or applications, as desired. Some examples of changes you may want to make include:
+    -   Changing Firefox's default startup settings and homepage.
+    -   Changing Nautilus' default file preview settings.
```
andrewdavidwong: Not sure if this is a good example anymore, since we always have previews off by default now, right?

customization/dispvm-customization.md
```diff
@@ -10,10 +10,64 @@ redirect_from:
...
+    -   Changing the DispVM's default NetVM. For example, you may wish to set the NetVM to "none." Then, whenever you start a new DispVM, you can choose your desired ProxyVM manually (by changing the newly-started DipsVMs settings). This is useful if you sometimes wish to use a DispVM with a TorVM, for example. It is also useful if you sometimes wish to open untrusted files in a network-disconnected DispVM.
```
andrewdavidwong: We should probably say `sys-whonix` or "Whonix Gateway" now instead of `TorVM`.

customization/dispvm-customization.md
marmarek: What do you mean? Just DispVM term?

customization/dispvm-customization.md
marmarek: One may want to _enable_ it in DispVM.

customization/dispvm-customization.md
marmarek: If you enable `appmenus-dispvm` feature (as explained at the top), applications menu for this VM (`fedora-26-dvm`) will be "Disposable: fedora-26-dvm" (*instead of* "Domain: fedora-26-dvm") and entries there will start new DispVM based on that VM (`fedora-26-dvm`). Not in that VM (`fedora-26-dvm`) itself.
How can I rephrase it to make it clearer?

customization/dispvm-customization.md
andrewdavidwong: Yes. This is introducing "DispVM" as an abbreviation of "Disposable VM." The problem is that the abbreviation "DispVM" is used *before* it is introduced here. Introducing it on first use solves the problem.

customization/dispvm-customization.md
andrewdavidwong: I think you just did. :slightly_smiling_face:

customization/dispvm-customization.md
andrewdavidwong: Sure, but that seems uncommon. A more common example might be easier to understand, but it's not a big deal.

customization/dispvm-customization.md
marmarek: Where is it used before? This is the first place where DispVM word is used.

andrewdavidwong: Well, it's used in `template_for_dispvms` near the beginning and in several commands after that. Granted, those aren't the same as normal words, but it would be easier for a reader to understand that part if they knew what the `dispvm` part meant.

PR #544, opened 2018-01-30
formatting fixes

adrelanos:


andrewdavidwong:
This does not achieve what you want. Code blocks must be indented eight spaces (not just four) within numbered lists (if using indentation to designate code blocks), or else they will not render as code blocks. This also does not preserve the numbering. Each number restarts at 1. I'll fix the formatting of this page in a separate commit.

PR #546, opened 2018-01-30
usb 4.0 updates

awokd:
Will be breaking up the update work across multiple commits but am opening the PR early for visibility and to avoid duplication of efforts. Will add a comment when ready for review.

### DIFF
--- a/common-tasks/usb.md
+++ b/common-tasks/usb.md
@@ -35,6 +35,95 @@ qube's settings page in Qubes VM Manager or by using the
 [qvm-pci][Assigning Devices] command. For guidance on finding the correct USB
 controller, see [here][usb-controller].)
 
+ * R4.0
+
+USB drive mounting is integrated into the Connection Widget. This is the tooltray
+icon with a yellow square located in the top right of your screen by default.
+Simply insert
+your USB drive and click on the widget. You will see multiple entries for your
+USB drive; typically, `sys-usb:sda`, `sys-usb:sda1`, and `sys-usb:2-1` for example.
+The simplest (but slightly less secure, see note below about attaching individual
+partitions) option is to attach the entire block drive. In our example, this is `sda`,
+so hover over it.
+This will pop up a submenu showing running VMs to which the USB drive can be connected.
+Click on one and your USB drive will be attached!
+
+Note that attaching individual partitions can be slightly more secure because it doesn't
+force the target AppVM to parse the partition table. However, it often means the 
+AppVM won't detect the new partition and you will need to manually mount it inside
+the AppVM. To do this with the GUI,
+you'd select the `sda1` entry in our example and proceed to connect to an AppVM.
+Once the USB drive has been attached to the AppVM, it will
+appear as `/dev/xvd*` (usually `xvdi` but sometimes with higher letters if you
+have multiple devices attached.) Follow the below steps if you need to manually mount
+the partition:
+    ```
+    cd ~
+    mkdir mnt
+    sudo mount /dev/xvdi mnt
+    ```
+    And when done:
+    `sudo umount mnt`
+    
+The command-line tool you may use to mount whole USB drives or their partitions
+is `qvm-block`. This tool can be used to assign a USB drive to a qube as
+follows:
+
+ 1. Insert your USB drive.
+
+ 2. In a dom0 console (running as a normal user), list all available block
+    devices:
+
+        qvm-block
+
+    This will list all available block devices connected to any USB controller
+    in your system, no matter which qube hosts the controller. The name of the
+    qube hosting the USB controller is displayed before the colon in the device
+    name. The string after the colon is the name of the device used within the
+    qube, like so:
+
+        dom0:sdb1     Cruzer () 4GiB
+        
+        usbVM:sdb1    Disk () 2GiB
+
+    **Note:** If your device is not listed here, you may refresh the list by
+    calling (from the qube to which the device is connected):
+
+        sudo udevadm trigger --action=change
+
+ 3.  Assuming your USB drive is attached to dom0 and is `sdb`, we attach the
+     device to a qube with the name `personal` like so:
+
+         qvm-block a personal dom0:sdb
+
+     This will attach the device to the qube as `/dev/xvdi` if that name is not
+     already taken by another attached device, or `/dev/xvdj`, etc.
+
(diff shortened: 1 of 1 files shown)


common-tasks/usb.md
```diff
@@ -21,199 +21,108 @@ redirect_from:
...
+sticks, this includes things like USB external hard drives.)
 
-How to use a USB keyboard
--------------------------
+Qubes OS supports the ability to attach a USB drive (or just one or more of its
+partitions) to any qube easily, no matter which qube actually handles the USB
+controller. (The USB controller may be assigned on the **Devices** tab of a
+qube's settings page in Qubes VM Manager or by using the
+[qvm-pci][Assigning Devices] command. For guidance on finding the correct USB
+controller, see [here][usb-controller].)
```
marmarek: This suggests that connecting USB controller to a VM is a common action, and may even encourage users to do it each time USB drive is connected. I'd clearly separate "creating USB qube" section (one time setup) and "using USB qube" (common actions).

common-tasks/usb.md
```diff
@@ -21,199 +21,108 @@ redirect_from:
...
 
-**Caution:** Please carefully read the [Security Warning about USB Input Devices] before proceeding.
+ * R4.0
+
+USB drive mounting is integrated into the Connection Widget. This is the tooltray
```
marmarek: This widget is called "Devices Widget". Although it is indeed hard to get this name without looking into code... We should add some description there (tooltip?): https://github.com/QubesOS/qubes-issues/issues/3509

common-tasks/usb.md
```diff
@@ -21,199 +21,108 @@ redirect_from:
...
+    qube, like so:
 
-(Change `sys-usb` to your desired USB qube.)
+        dom0:sdb1     Cruzer () 4GiB
+        
+        usbVM:sdb1    Disk () 2GiB
 
-You can now use your USB keyboard.
+    **Note:** If your device is not listed here, you may refresh the list by
+    calling (from the qube to which the device is connected):
```
marmarek: I'd clarify that "the qube to which the device is connected" is usbVM in most cases..

common-tasks/usb.md
```diff
@@ -21,199 +21,108 @@ redirect_from:
...
-```
+        sudo udevadm trigger --action=change
 
-How to use a USB mouse
-----------------------
+ 3.  Assuming your USB drive is attached to dom0 and is `sdb`, we attach the
+     device to a qube with the name `personal` like so:
 
-**Caution:** Please carefully read the [Security Warning about USB Input Devices] before proceeding.
+         qvm-block a personal dom0:sdb
```
marmarek: While valid example, better give one with usbVM here (using default sys-usb name). Because this is what users will get in default installation.
Also, `a` action is convenient, but using full `attach` name will more readable.

common-tasks/usb.md
```diff
@@ -21,199 +21,108 @@ redirect_from:
...
+     You may also mount one partition at a time by using the same command with
+     the partition number after `sdb`.
 
-Add a line like this to the top of the file:
+     **Warning:** when working with single partitions, it is possible to assign
+     the same partition to multiple qubes. For example, you could attach `sdb1`
+     to qube1 and then `sdb` to qube2. It is up to the user not to make this
+     mistake. The Xen block device framework currently does not provide an easy
+     way around this. Point 2 of [this comment on issue 1072][1072-comm2] gives
+     details about this.
```
marmarek: Actually, in Qubes 4.0 it is already prevented. When you attach `sdb1`, `sdb` is removed from list of available devices.

common-tasks/usb.md
```diff
@@ -21,199 +21,108 @@ redirect_from:
...
 
-For a confirmation dialog each time the USB mouse is connected, change this line to:
-```
-sys-usb dom0 ask,default_target=dom0
-```
+ 6.  In a dom0 console, detach the stick
 
-How to attach USB drives
-------------------------
+         qvm-block d <vmname> <device>
```
marmarek: And here - `d` -> `detach`.

common-tasks/usb.md
```diff
@@ -383,6 +293,185 @@ When you finish, detach the device:
...
+You can create a USB qube using the management stack by performing the following
+steps as root in dom0:
+
+ 1. Enable `sys-usb`:
+
+        sudo qubesctl top.enable qvm.sys-usb
+
+ 2. Apply the configuration:
+
+        sudo qubesctl state.highstate
```
marmarek: Better compress this to one line action: `sudo qubesctl state.sls qvm.sys-usb`
This will also not leave `qvm.sys-usb` enabled and applied on every next `state.highstate` call.

common-tasks/usb.md
```diff
@@ -383,6 +293,185 @@ When you finish, detach the device:
...
+
+You can now use your USB keyboard.
+
+For a confirmation dialog each time the USB keyboard is connected, change this line to:
+```
+sys-usb dom0 ask,default_target=dom0
+```
+
+How to use a USB mouse
+----------------------
```
marmarek: Maybe worth noting that this is also done as part of `qubesctl state.sls qvm.sys-usb` (or if user selected USB qube during first boot).

awokd:
Still working through this document, @marmarek but thanks for the comments, will go back and incorporate them.

awokd:
Not quite finished; does AEM apply to R4.0? Going to add EFI vs. grub too.

common-tasks/usb.md
awokd: I was trying to clearly separate them with https://github.com/QubesOS/qubes-doc/pull/539. How else would you suggest?

awokd:
OK, I think that should about do it, assuming no changes are needed for AEM & 4.0. Sorry for the noise. Please review when possible. Detaching a USB device with the widget didn't work quite right, so I commented on a related issue already out there (3215).

common-tasks/usb.md
```diff
@@ -322,7 +369,7 @@ steps as root in dom0:
 
  2. Apply the configuration:
 
-        sudo qubesctl state.highstate
+        sudo qubesctl state.sls qvm.sys-usb
```
marmarek: When using `state.sls`, `top.enable` is not needed at all.

awokd:
Done.

PR #548, opened 2018-01-31
hvm 4.0rc4 updates

awokd:
Add mention and link of PVH.

### DIFF
--- a/managing-os/hvm.md
+++ b/managing-os/hvm.md
@@ -17,8 +17,14 @@ What are HVM domains?
 
 HVM domains (Hardware VM), in contrast to PV domains (Paravirtualized domains), allow one to create domains based on any OS for which one has an installation ISO.  For example, this allows one to have Windows-based VMs in Qubes.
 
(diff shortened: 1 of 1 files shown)


awokd:
Regarding this section towards the bottom:
```
One problem at the moment however, is that after the whole system gets suspended into S3 sleep and
subsequently resumed, some attached devices may stop working and should be restarted within the VM. 
```
Should this content be moved to one of the Win7 specific pages? I'd like to add the Linux equivalent about blacklisting drivers that don't support sleep, but am not sure if it's best to add here or on LinuxHVMTips.

marmarek:
> Should this content be moved to one of the Win7 specific pages? 

TBH I'm not sure if it is still needed. There were some related fixes in 4.0, but I haven't tested win7 suspend after that. Anyway I'd leave this generic sentence here.

> I'd like to add the Linux equivalent about blacklisting drivers that don't support sleep, but am not sure if it's best to add here or on LinuxHVMTips.

I'd say LinuxHVMTips, which is also linked from this page.

PR #550, opened 2018-02-01
reinstall-template 4.0 update

awokd:
add 4.0 clone template procedure because dummy template no longer works

### DIFF
--- a/managing-os/reinstall-template.md
+++ b/managing-os/reinstall-template.md
@@ -11,13 +11,35 @@ How to Reinstall a TemplateVM
 
 If you suspect your [TemplateVM] is broken, misconfigured, or compromised, you can reinstall any TemplateVM that was installed from the Qubes repository.
 
-The procedure varies by Qubes version; see the appropriate section below.
+The procedure varies by Qubes version and UpdateVM's distribution; see the appropriate section below.
 
 
-Manual Reinstallation Method (R4.0)
+Manual Reinstallation Method (Fedora based UpdateVM, R3.1+)
 ----------------------------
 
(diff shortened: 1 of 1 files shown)


awokd:
Please pend this PR until https://github.com/QubesOS/qubes-issues/issues/3553, or let me know if I should close this one and open anew. I need to rewrite the reinstall portion. Thought it was R4.0 vs. R3.2 but it's actually Debian vs. Fedora updateVM.

awokd:
Renamed sections so this should accurately describe how one would reinstall a template today. Can touch again later if #3553 changes anything, but I think this is good as is right now.

managing-os/reinstall-template.md
```diff
@@ -9,72 +9,106 @@ redirect_from:
...
 =============================
 
-If you suspect your [TemplateVM] is broken, misconfigured, or compromised, you
-can reinstall any TemplateVM that was installed from the Qubes repository.
-Starting in Qubes 3.1, the process is greatly simplified.
+If you suspect your [TemplateVM] is broken, misconfigured, or compromised, you can reinstall any TemplateVM that was installed from the Qubes repository.
 
-First, copy any files that you wish to keep from the TemplateVM's `/home` and
-`/rw` folders to a safe storage location. Then, in a dom0 terminal, run:
+The procedure varies by Qubes version and UpdateVM's distribution; see the appropriate section below.
```
marmarek: IMO it would be good to point here how to check what is set as UpdateVM.

awokd:
Added, but I'm just going to have to remove it again once --action=reinstall works on Debian!

PR #551, opened 2018-02-01
security-guidelines 4.0 update

awokd:
yum -> dnf
Replace R3.x specific description of treatment of template's home directory with link to /doc/templates


awokd:
Same question on this one- does AEM apply to R4.0? Haven't attempted it in any version.

andrewdavidwong:
> Same question on this one- does AEM apply to R4.0? Haven't attempted it in any version.

@marmarek

marmarek:
Yes, AEM should work on R4.0 too. But note only legacy boot is supported right now, no UEFI. Regardless of Qubes version.

PR #552, opened 2018-02-02
firewall 4.0 updates

awokd:
Depending on the accuracy of the following statement, I may need to go back and tweak the scripts in this document. Please let me know.

+Where to put firewall rules (R4.0)
+---------------------------
+
+Implicit in the above example [scripts](/doc/config-files/), but worth 
+calling attention to: for all qubes *except* NetVMs, iptables commands 
+should be added to the `/rw/config/rc.local` script. For NetVMs 
+(`sys-firewall` inclusive), iptables commands should be added to 
+`/rw/config/qubes-firewall-user-script`. This is because a NetVM is 
+constantly adjusting its firewall, and therefore initial settings from 
+`rc.local` do not persist.

### DIFF
--- a/security/firewall.md
+++ b/security/firewall.md
@@ -374,12 +374,11 @@ Where to put firewall rules (R4.0)
 ---------------------------
 
 Implicit in the above example [scripts](/doc/config-files/), but worth 
-calling attention to: for all qubes *except* NetVMs, iptables commands 
-should be added to the `/rw/config/rc.local` script. For NetVMs 
-(`sys-firewall` inclusive), iptables commands should be added to 
-`/rw/config/qubes-firewall-user-script`. This is because a NetVM is 
-constantly adjusting its firewall, and therefore initial settings from 
-`rc.local` do not persist.
+calling attention to: for all qubes *except* AppVMs supplying networking,
+iptables commands should be added to the `/rw/config/rc.local` script. For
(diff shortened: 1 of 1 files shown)


tasket:
A few points:
- The `qubes-firewall-user-script` is currently not executed on a 'netVM' startup in R4.0-rc. The `qubes-ip-change-hook` script is run instead. See issue https://github.com/QubesOS/qubes-issues/issues/3260
- Last sentence '...settings from rc.local do not persist' is more accurate for R3.x than R4.0. New behavior is that iptables/nftables remain fairly static, except for Qubes-specific chains like QBS-FORWARD. (The shift to `qubes-ip-change-hook` filename seems like a contradiction, but that's how it is now.)
- Using 'NetVM' here may be confusing in some cases. Internally, domUs still distinguish between netvm and proxyvm states even though the Create VM dialog no longer does. The difference seems to be whether an upstream netvm is defined (netvm = None results in the domU starting in a 'NetVM' state).

awokd:
Thanks for jumping in, @tasket . Read through that issue and made a couple edits per your comment. If we assume `qubes-firewall-user-script` will be addressed, would the documentation be OK as is? If so, suggest we just park this PR until then.
I'm still a bit unclear on how to phrase that last part. If we re-word the deprecated (for R4.0) term ProxyVM to "appVM supplying networking" like I did, does that make it correct or does it not apply to sys-net for example?

PR #556, opened 2018-02-02
vpn.md: disconnect-resistant fail closed configuration

hyperfekt:
If the connection to the VPN server drops long enough for it to be reinitiated (60s by default), the DNS lookup for the server will fail and prevent reestablishing the connection. `persist-remote-ip` reuses the IP from the previous connection.

### DIFF
--- a/configuration/vpn.md
+++ b/configuration/vpn.md
@@ -102,6 +102,11 @@ It has been tested with Fedora 23 and Debian 8 templates.
    Make sure it already includes or add:
 
        redirect-gateway def1
+   
+   Since we will be routing DNS queries to go through the VPN, the client will not be able to look up the server's IP if it disconnects, preventing it from reconnecting again.
+   To fix this, add the following to remember the initially resolved IP address:
+   
+       persist-remote-ip
 
    The VPN client may not be able to prompt you for credentials when connecting to the server.
    Create a file in the `/rw/config/vpn` folder with your credentials and using a directive.


tasket:
Thanks. This will likely be addressed in the upcoming revision by using a service-based configuration which restarts openvpn as necessary and doesn't override network policies.

Typically, when openvpn fails here it tries to re-use the IP anyway before exiting with a spurious error like auth-failure. Additionally, the supplied config file might have some `user=` or `group=` priv reduction which could prevent DNS or any other upstream access. In this case it may help to override user and/or group on the ExecStart command line, so that is the change I'd opt for; It also can be provided in the doc as just a part of the existing script, instead of as another option that they have to consider and then make a special edit.

hyperfekt:
Alright, that seems like a much more general solution, looking forward to it!
(In my case openvpn did not reuse the IP.)

tasket:
@hyperfekt - If you would like to start using a service-based configuration now, my [VPN project](https://github.com/tasket/Qubes-vpn-support) has one ready to use; note the `qubes4` branch.

adrelanos:
Will https://github.com/tasket/Qubes-vpn-support be suitable for installation by default? (Enabling through qvm-service or so?)

tasket:
On the `qubes4` branch, I think so. It works well.

My first reservation for putting it in Qubes (besides the naming "vpn...") is that it uses an interactive install script which seems not very Qubes-like behavior; I imagined this kind of install/config activity to be handled with qubesctl and I'm not to the point with saltstack where I can do this.

Another reservation is my preference to migrate the firewall settings into qubes-firewall, and to do it in a way where a 3rd-party client like Bitmask could enable it internally. Its also more elegant than having a piece of the OS try to populate a user-oriented script for what is now an OS function.

But the service config yes I'd like that to be the same as Qubes-vpn-support, just with a better name.

FWIW, I've started overhauling the vpn.md starting with the part that will be for R3.2: https://github.com/tasket/qubes-doc/tree/tasket03 - The thing is that Qubes-vpn-support could also be dropped into R3.2 codebase more easily than making the script-bearing doc seem user-friendly (and would correct the 3.2 DNS-blocking bug in the process, whereas the correcting code is a bit large to include in the doc version).

---

So there is:
* The naming (e.g. qubes-tunnel-somesuch)
* Integrate into qubes-firewall w internal on/off
* Keep setup script? Enables very brief instructions.
* For R3.2: Update doc scripts, or add to Qubes codebase?

ref https://github.com/QubesOS/qubes-issues/issues/3520

adrelanos:
It boils down to the question how do we stack iptables / nftables rules?

I think a good approach would be using drop-in folder
/usr/share/netfilter-persistent/plugins.d which is parsed by
netfilter-persistent.

It's available in Debian.

https://packages.debian.org/search?keywords=netfilter-persistent

Probably not available in Fedora but I don't see why this package should
be hard to port to Fedora. Stackable iptables / nftables rules is
something all distributions lack.

netfilter-persistent doesn't have much bugs / hard to fix bugs either:

https://bugs.debian.org/cgi-bin/pkgreport.cgi?pkg=netfilter-persistent;dist=unstable

So the tunnel firewall script in
/usr/share/netfilter-persistent/plugins.d could do config file (or even
.d folder) parsing. Doing nothing by default. And being enabled by
switching an option. This is then also something easily doable using salt.

I think long term also Qubes should port to netfilter-persistent.

Bitmask should also drop a snippet into
/usr/share/netfilter-persistent/plugins.d rather than modifying iptables
rules globally.

PR #560, opened 2018-02-03
secondary-storage 4.0 updates

awokd:
add 4.0 content (sourced from https://mail-archive.com/qubes-users@googlegroups.com/msg14871.html)
add version headings
adjust R3.2 procedure for consistency with rest of document
remove "#" from R3.2 procedure because it works as regular user if permissions are set correctly
remove "known issues" heading so bullet points fall under R3.2 section instead of appearing to apply to both versions
remove note about R3.1 because loop notifications also occur in R3.2

### DIFF
--- a/configuration/secondary-storage.md
+++ b/configuration/secondary-storage.md
@@ -12,19 +12,59 @@ Storing AppVMs on Secondary Drives
 ==================================
 
 Suppose you have a fast but small primary SSD and a large but slow secondary
-HDD.  You want to store a subset of your AppVMs on the HDD. In dom0:
+HDD.  You want to store a subset of your AppVMs on the HDD.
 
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
> remove note about R3.1 because loop notifications also occur in R3.2

Was that a regression? I'm fairly certain that that was true when I wrote it.

awokd:
Not sure if it's a regression, but I've only been running an AppVM off secondary for the past 3 months or so. I read through the linked thread (https://groups.google.com/forum/?_escaped_fragment_=topic/qubes-users/nDrOM7dzLNE/discussion#!topic/qubes-users/nDrOM7dzLNE/discussion) and saw where those loop notifications were discussed but not anything about a fix for them. Doesn't mean there's not a commit somewhere out there!

PR #561, opened 2018-02-03
resize-disk-image and resize-root-disk-image 4.0 updates/consolidation

awokd:
Please see commit notes


configuration/resize-disk-image.md
```diff
@@ -1,47 +1,72 @@
...
+qvm-grow-private <vm-name> <size>
+~~~
+
+Note: Size is the target size (i.e. 4096MB or 16GB, ...), not the size to add to the existing disk. 
+
+### Shrinking private disk image (Linux VM, R4.0)
+
+1.  Create a new qube with smaller disk using Qube Manager or `qvm-create`
+2.  Move data to the new qube using `qvm-copy` or OS utilities
+3.  Delete old qube using Qube Manager or `qvm-remove`
```
marmarek: This procedure will also work on R3.2, is less risky and simpler. I suggest deprecating the R3.2 instruction and having this one for both versions.

marmarek: It is also possible to prepare similar (the risky one) instruction for R4.0, which do not require duplicating the data. But as said before, I'm not sure if that's a good idea.

configuration/resize-disk-image.md
```diff
@@ -63,51 +88,28 @@ Now you can resize the image:
...
-First, stop/shutdown the HVM.
-
-Then, from a Dom0 terminal (in KDE: System Tools -\> Terminal Emulator) do the following:
-
-~~~
-cd /var/lib/qubes/appvms/<yourHVM>/
-ls -lh root.img  (<--verify current size of disk image)
-truncate -s 30GB root.img
-ls -lh root.img  (<--verify new size of disk image)
-~~~
```
marmarek: IMO it's more logical to merge those two pages - have one about disk resize. @andrewdavidwong what do you think?

awokd: FWIW, merging the two makes sense to me too and I thought about doing it but didn't want to make this PR even busier. The first two procedures only differ by a couple lines (`:root` vs. `:private`)  so I could collapse them entirely into each other. If @andrewdavidwong agrees, I'll revisit these after this PR's done and do so.

configuration/resize-disk-image.md
awokd: Will address on the next round!

configuration/resize-disk-image.md
andrewdavidwong: Merging these two documents is fine with me.

PR #565, opened 2018-02-04
assigning-devices 4.0 updates

awokd:
add introduction explaining why this can be needed and multi-function device quirks
add 4.0 content
add introduction to USB controller section explaining why
add strict reset options for R3.2 and R4.0


awokd:
Of course, that makes sense.

configuration/assigning-devices.md
```diff
@@ -11,71 +11,121 @@ redirect_from:
...
+USB qube during install.
+While this covers most use cases, there are some occasions when you may want to
+manually assign one NIC to `sys-net` and another to a custom NetVM,
+or have some other type of PCI controller you want to manually assign.
+
+Note that one can only assign full PCI or PCI Express devices by default.
+This limit is imposed by the PC and VT-d architectures.
+This means if a PCI device has multiple functions, all instances
+of it need to be assigned to the same qube unless you have disabled FLR with the
+`no-strict-reset` (R4.0) or `pci_strictreset` (R3.2) option.
```
marmarek: Strictly speaking, if device support FLR (Function Level Reset), it _is_ possible to assign a different functions of the same device to different VMs. But in practice very few devices support FLR and by default toolstack refuse such configuration is FLR is not supported.
The end effect is similar, but it isn't about disabling FLR, but about not refusing to continue if FLR is missing.

configuration/assigning-devices.md
```diff
@@ -11,71 +11,121 @@ redirect_from:
...
+In the steps below, you can tell if this is needed if you see the BDF for the
+same device listed multiple times with only the number after the "." changing.
+
+While a device can only be attached to one VM at a time, it *is* possible to
+*assign* the same device to more than one VM at a time.
```
marmarek: Should we define "assigned" and "attached" somewhere additionally? This is very subtle difference and it can be confusing in some cases.

awokd:
I did see Andrew's link prior to my last commit but it said nothing about breaks in the middle of a sentence, only "Insert a newline at the end of each sentence". However, I went back and removed them and will keep it in mind. I'll go back and check my other outstanding PRs for them too.

configuration/assigning-devices.md
awokd: Thanks, meant to re-word that but forgot.

configuration/assigning-devices.md
awokd: Edited, hopefully more clear now.

andrewdavidwong:
> I did see Andrew's link prior to my last commit but it said nothing about breaks in the middle of a sentence, only "Insert a newline at the end of each sentence".

Clarified that point to avoid misunderstanding.

andrewdavidwong:
I'm assuming that this particular Travis failure is already (or going to be) resolved separately, since we've already discussed it elsewhere.

PR #569, opened 2018-02-07
qubes-r3-building 4.0 updates

awokd:
make document title generic
add build VM recommendations
update dependencies
add typical gpg sequence of fpr, trust, 5, q
add key retrieval for Whonix templates (please verify process and key ID!)
add .setup steps
misc spelling/grammar

### DIFF
--- a/building/qubes-r3-building.md
+++ b/building/qubes-r3-building.md
@@ -1,6 +1,6 @@
 ---
 layout: doc
-title: Qubes R3 Building
+title: Qubes ISO Building
 permalink: /doc/qubes-r3-building/
 redirect_from:
 - /en/doc/qubes-r3-building/
@@ -8,13 +8,27 @@ redirect_from:
 - /wiki/QubesR3Building/
 ---
 
-Building Qubes OS 3.0 ISO
+Building Qubes OS ISO
 =========================
 
-Ensure your system is rpm-based and that you have necessary dependencies installed (see [QubesBuilder](/doc/qubes-builder/) for more info):
+In `dom0`, install the Fedora 26 template if you don't already have it.
+Other rpm-based operating systems may also work, but Fedora 26 has been successfully used to build Qubes R3.2 and R4.0 with the below steps.
 
 ~~~
-sudo yum install git createrepo rpm-build make wget rpmdevtools pandoc
+sudo qubes-dom0-update qubes-template-fedora-26
+~~~
+
+Create a standalone appVM from the Fedora 26 template.
+You may choose your own name, but this document will refer to it as `dev26`.
+Set private storage to at least 60000MB if you will be building only the default templates; 100000MB if you plan on additional.
+It's not required, but if you allocate additional CPU cores, the build process can utilize them at some steps such as the kernel build.
+Likewise, more memory (up to 16GB) can help.
+Last, you may want to disable memory balancing on `dev26` but keep in mind the impact on your other qubes.
+
+Once you've built `dev26`, open a Terminal window to it and install the necessary dependencies (see [QubesBuilder](/doc/qubes-builder/) for more info):
+
+~~~
+$ sudo dnf install git createrepo rpm-build make wget rpmdevtools dialog rpm-sign gnupg dpkg-dev debootstrap python2-sh
 ~~~
 
 Get the necessary keys to verify the sources:
@@ -24,36 +38,71 @@ $ wget https://keys.qubes-os.org/keys/qubes-master-signing-key.asc
 $ gpg --import qubes-master-signing-key.asc 
 $ gpg --edit-key 36879494
 # Verify fingerprint!, set trust to *ultimate*
+# (Typical sequence is fpr, trust, 5, q)
 $ wget https://keys.qubes-os.org/keys/qubes-developers-keys.asc
 $ gpg --import qubes-developers-keys.asc
 ~~~
 
-Note we do *not* relay above on the security of our server (keys.qubes-os.org) nor the connection (ssl, cert) -- we only rely on you getting the Qubes Master Signing Key fingerprint *somehow* and ensure they match!
+And if you will be building the Whonix templates:
 
-Now lets bootstrap the builder. Unfortunately the builder cannot verify itself (the classic Chicken and Egg problem), so we need to verify the signature manually:
+~~~
+$ gpg --keyserver pgp.mit.edu --recv-keys 916B8D99C38EAF5E8ADC7A2A8D66066A2EEACCDA
+~~~
+
+**Note** In the above process, we do *not* rely on the security of our server (keys.qubes-os.org) nor the connection (ssl, cert) -- we only rely on you getting the Qubes Master Signing Key fingerprint *somehow* and ensuring they match!
+Likewise, the Whonix signing key is available from multiple sources.
+
+Now let's bootstrap the builder. Unfortunately, the builder cannot verify itself (the classic Chicken and Egg problem), so we need to verify the signature manually:
 
 ~~~
 $ git clone git://github.com/QubesOS/qubes-builder.git
+$ mkdir qubes-builder/keyrings
+$ mkdir qubes-builder/keyrings/git
+$ cp .gnupg/pubring.gpg qubes-builder/keyrings/git/
+$ cp .gnupg/trustdb.gpg qubes-builder/keyrings/git/
 $ cd qubes-builder
-$ git describe --exact-match HEAD
-<some tag>
-$ git tag -v <some tag>
+$ git tag -v `git describe`
 ~~~
 
-Assuming the verification went fine, we're good to go with all the rest without ever thinking more about verifying digital signatures on all the rest of the components, as the builder will do that for us, for each component, every time we, even for all aux files (e.g. Xen or Linux kernel sources).
+Assuming the verification went fine, we're good to go with all the rest without ever thinking more about verifying digital signatures on all the rest of the components.
+The builder will do that for us for each component, every time we build, even for all auxiliary files (e.g. Xen or Linux kernel sources).
 
-Let's configure the builder first (we can use one of the example configs, either for R2 or "master", which currently means pre-released R3):
-
-~~~
-cp example-configs/qubes-os-master.conf builder.conf
-~~~
-
-You can take a loot at the `builder.conf.default` for a description of all available options. Nevertheless, the default config should be enough for start:
+Let's configure the builder first (see Note at bottom if you would prefer to manually configure):
 
 ~~~
-$ make get-sources qubes
-$ make sign-all # this requires setting SIGN_KEY in the builder.conf, can be skipped for test builds.
+$ ./setup
+# Select 3.2 or 4.0 for version
+# Stable
+# No (we want a full build)
+# Select builder-fedora, builder-debian, template-whonix, mgmt-salt
+# Yes (to download)
+# Select fc26, stretch, whonix-gateway, whonix-workstation (for the currently shipping templates)
+~~~
+
+Once the download is complete, continue the build process with:
+
+~~~
+$ make install-deps
+$ make get-sources
+~~~
+
+Finally, if you are making a test build, use:
+
+~~~
+$ make qubes
+$ make iso
+~~~
+
+Or for a fully signed build (this requires setting SIGN_KEY in the builder.conf):
+
+~~~
+$ make sign-all
 $ make iso
 ~~~
 
 Enjoy your new ISO!
+
+
+**Note** Instead of using `./setup`, you can manually configure the build by doing `cp example-configs/qubes-os-master.conf builder.conf` and editing `builder.conf`.
+Take a look at `builder.conf.default` for a description of all available options.
+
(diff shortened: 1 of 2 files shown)


awokd:
Am jumping ahead to this document because there have been a few requests on qubes-users lately. Doubt I have to mention it, but please verify that Whonix key ID and retrieval process. Assume I am malicious or unwittingly passing along a compromised key.

awokd:
@adrelanos I saw you commented on another of my PRs- could you please check out this one too? Prestaging the Whonix key like this was more reliable on my builds, and could also give an opportunity to review it but maybe there are additional steps that should be called out.

adrelanos:
Thank you for working on Qubes / Qubes-Whonix build documentation!

See qubes-builder TravisCI.

https://github.com/QubesOS/qubes-issues/issues/3484#issuecomment-361372701

So I think if you are not using `setup`, you may have to use:

```
export GNUPGHOME=~/qubes-builder/keyrings/git
mkdir --parents "$GNUPGHOME"
chmod --recursive 700 "$GNUPGHOME"
```

(This I learned from qubes-builder TravisCI log.)

`gpg --keyserver pgp.mit.edu --recv-keys 916B8D99C38EAF5E8ADC7A2A8D66066A2EEACCDA` alone is not secure as only the short key id is checked. One has to run `gpg --fingerprint 916B8D99C38EAF5E8ADC7A2A8D66066A2EEACCDA` to check if the correct full fingerprint has been received.

Better use https://github.com/QubesOS/qubes-builder-debian/blob/master/keys/whonix-developer-patrick.asc. From within qubes-builder source directory:

`gpg --import keys/whonix-developer-patrick.asc`

Then it is as trusted as qubes-builder source directory.

Anyhow. Also gpg ownertrust has to be adjusted (qubes-builder needs that, not Whonix specific).

    echo '916B8D99C38EAF5E8ADC7A2A8D66066A2EEACCDA:6:' | gpg --import-ownertrust

You may also need to adjust gpg ownertrust for Qubes key.

    echo '427F11FD0FAA4B080123F01CDDFA1A3E36879494:6:' | gpg --import-ownertrust

Not sure that is still required. `setup` may do that for you.

----

Cannot comment on the rest. I am not well versed with qubes-builder. For building Qubes-Whonix, please check, try and document (perhaps in later pull request to not increase complexity of this one):
https://github.com/QubesOS/qubes-issues/issues/3441#issuecomment-359447019

awokd:
Thanks, @adrelanos . It's that key handling step that's the most sensitive so I wanted extra eyes on it. My builds worked when using `./setup` without the `--import-ownertrust` but I added them and the `$GNUPGHOME` steps to the Note about manual builds.

building/qubes-r3-building.md
```diff
@@ -24,36 +38,80 @@ $ wget https://keys.qubes-os.org/keys/qubes-master-signing-key.asc
...
 $ wget https://keys.qubes-os.org/keys/qubes-developers-keys.asc
 $ gpg --import qubes-developers-keys.asc
 ~~~
 
-Note we do *not* relay above on the security of our server (keys.qubes-os.org) nor the connection (ssl, cert) -- we only rely on you getting the Qubes Master Signing Key fingerprint *somehow* and ensure they match!
+And if you will be building the Whonix templates:
+
+~~~
+$ wget https://github.com/QubesOS/qubes-builder-debian/blob/master/keys/whonix-developer-patrick.asc
+$ gpg --import whonix-developer-patrick.asc
```
marmarek: I'd defer this step to later point, where builder-debian is already downloaded and its signature is verified. Otherwise you rely here on just https + github.

building/qubes-r3-building.md
```diff
@@ -24,36 +38,80 @@ $ wget https://keys.qubes-os.org/keys/qubes-master-signing-key.asc
...
+$ mkdir qubes-builder/keyrings
+$ mkdir qubes-builder/keyrings/git
+$ cp .gnupg/pubring.gpg qubes-builder/keyrings/git/
+$ cp .gnupg/trustdb.gpg qubes-builder/keyrings/git/
 $ cd qubes-builder
-$ git describe --exact-match HEAD
-<some tag>
-$ git tag -v <some tag>
+$ git tag -v `git describe`
+~~~
```
marmarek: This whole step can be simplified to the old version: you don't need to use `qubes-builder/keyrings/git/pubring.gpg` for manual verification. After all, you've imported master signing key to your default keyring, not here.
Instead, I'd add extra step:
```
gpg --import qubes-developers-keys.asc
```
This is because tags signatures on qubes-builder (and every other component) are made using some of those keys. And those keys are signed by the master key. Since the master key is already set to trust 'ultimate', gpg will use that when evaluating trustworthiness of the keys here. It's very important to check if the verification message contains "Good signature from ..." and _do not contains_ "WARNING: This key is not certified with a trusted signature!".

building/qubes-r3-building.md
```diff
@@ -24,36 +38,80 @@ $ wget https://keys.qubes-os.org/keys/qubes-master-signing-key.asc
...
+$ make iso
 ~~~
 
-You can take a loot at the `builder.conf.default` for a description of all available options. Nevertheless, the default config should be enough for start:
+Or for a fully signed build (this requires setting SIGN_KEY in the builder.conf):
 
 ~~~
-$ make get-sources qubes
-$ make sign-all # this requires setting SIGN_KEY in the builder.conf, can be skipped for test builds.
+$ make sign-all
```
marmarek: This suggests that `make qubes` isn't needed for signed build, which isn't true.

building/qubes-r3-building.md
```diff
@@ -24,36 +38,80 @@ $ wget https://keys.qubes-os.org/keys/qubes-master-signing-key.asc
...
+# Select fc26, stretch, whonix-gateway, whonix-workstation (for the currently shipping templates)
 ~~~
 
-Assuming the verification went fine, we're good to go with all the rest without ever thinking more about verifying digital signatures on all the rest of the components, as the builder will do that for us, for each component, every time we, even for all aux files (e.g. Xen or Linux kernel sources).
+Once the download is complete, continue the build process with:
+
+~~~
+$ make install-deps
+$ make get-sources
+~~~
```
marmarek: setup script will take care about importing Whonix signing key (there will be a prompt for that).
But if going with manual builder configuration, this is a good place for importing that key - it is already downloaded to `qubes-src/builder-debian/keys/whonix-developer-patrick.asc`, like this:
```
gpg --homedir keyrings/git --import qubes-src/builder-debian/keys/whonix-developer-patrick.asc
```
You may need to execute `make get-sources` again to resume the operation (if any component failed because of missing key).

building/qubes-r3-building.md
```diff
@@ -24,36 +38,80 @@ $ wget https://keys.qubes-os.org/keys/qubes-master-signing-key.asc
...
-$ make sign-all # this requires setting SIGN_KEY in the builder.conf, can be skipped for test builds.
+$ make sign-all
 $ make iso
 ~~~
 
 Enjoy your new ISO!
+
+
+**Note** Instead of using `./setup`, you can manually configure the build by doing `cp example-configs/qubes-os-master.conf builder.conf` and editing `builder.conf`.
+Take a look at `builder.conf.default` for a description of all available options.
```
marmarek: Actually, better place is `doc/Configuration.md` file.

building/qubes-r3-building.md
```diff
@@ -24,36 +38,80 @@ $ wget https://keys.qubes-os.org/keys/qubes-master-signing-key.asc
...
+On manually configured builds, you may also need to:
+
+~~~
+export GNUPGHOME=~/qubes-builder/keyrings/git
+mkdir --parents "$GNUPGHOME"
+chmod --recursive 700 "$GNUPGHOME"
+echo '427F11FD0FAA4B080123F01CDDFA1A3E36879494:6:' | gpg --import-ownertrust
+~~~
```
marmarek: This will be done automatically on first `make get-sources` call. Unless you create the directory manually but forget to import keys there, which is exactly what you've done here...

building/qubes-r3-building.md
awokd: How is this different in concept than the 
```
$ wget https://keys.qubes-os.org/keys/qubes-master-signing-key.asc
$ gpg --import qubes-master-signing-key.asc 

$ wget https://keys.qubes-os.org/keys/qubes-developers-keys.asc
$ gpg --import qubes-developers-keys.asc
```
steps?
I'm trying to do all the keyring stuff at once.

building/qubes-r3-building.md
awokd: `gpg --import qubes-developers-keys.asc` is already done in the previous section. Are you saying this one can be simplified to:
```
$ git clone git://github.com/QubesOS/qubes-builder.git
$ cd qubes-builder
$ git tag -v `git describe`
```
?

building/qubes-r3-building.md
awokd: Original doc omitted that step then.

building/qubes-r3-building.md
awokd: That's why I'm trying to pre-stage it in the previous section, before we start copying keyrings around and before the user starts hitting errors related to it being missing.

building/qubes-r3-building.md
awokd: Don't these accomplish the same thing?
```
$ cp .gnupg/pubring.gpg qubes-builder/keyrings/git/
$ cp .gnupg/trustdb.gpg qubes-builder/keyrings/git/
```
But I'll remove.

building/qubes-r3-building.md
marmarek: Here you're importing key without verification if it is really the right one. When you have it downloaded by qubes-builder, it has verified that the file is covered by properly signed git tag.
Note that for the master key we have specific step of fingerprint verification. If we add such step also for the other key, it would be ok.

marmarek: Also, keys in qubes-developers-keys.asc are signed by master key (and we use that information later), but whonix key is not.

awokd: I understand both your points, thank you. Maybe the better way to handle this would be to make the script more robust instead of (or in addition to) in the documentation. I'll take a look and update this when I have a better idea.

building/qubes-r3-building.md
marmarek: Yes. Plus info how to interpret result - see previous comment.

building/qubes-r3-building.md
marmarek: Hmm, I wonder if this step is required at all, since I review updates to qubes-template-whonix and add my tag there too. It is probably only needed for development builds directly from Patrick's repository.

building/qubes-r3-building.md
marmarek: Yes, indeed. But generally IMO better to simplify the instruction as much as possible. The critical stuff is to validate if downloaded qubes-builder is properly signed, that should be the last thing one need to do with gpg manually. All further steps should be handled by scripts in qubes-builder. Less space for human error.

building/qubes-r3-building.md
marmarek: Ah, you've already added it :)

building/qubes-r3-building.md
marmarek: In reverse order, should be:
```
make qubes
make sign-all
make iso
```

rustybird:
> `gpg --keyserver pgp.mit.edu --recv-keys 916B8D99C38EAF5E8ADC7A2A8D66066A2EEACCDA` alone is not secure as only the short key id is checked. One has to run `gpg --fingerprint 916B8D99C38EAF5E8ADC7A2A8D66066A2EEACCDA` to check if the correct full fingerprint has been received.

Nah, they [fixed](https://dev.gnupg.org/T1579) that in 2014.

awokd:
Not sure if `echo '916B8D99C38EAF5E8ADC7A2A8D66066A2EEACCDA:6:' | gpg --import-ownertrust` is necessary in the manual procedure, so I left it out.
We could also split that manual procedure out to a separate doc if we want to keep this one clean, but seemed a bit overkill.

awokd:
Was running through the manual build steps and noticed the file at https://keys.qubes-os.org/keys/qubes-developers-keys.asc has an old date and is 40085 bytes vs. the same file from Git which is 50087 bytes. It caused Rafał's commits to not validate. Can it be updated, please?

marmarek:
Updated.

awokd:
Am missing a manual step or sequence somewhere, getting:
~~~
--> Downloading additional sources for vmm-xen...
Wrong signature on grub-0.97.tar.gz.UNTRUSTED!
make[1]: *** [Makefile:78: grub-0.97.tar.gz] Error 1
make: *** [Makefile:192: vmm-xen.get-sources-extra] Error 2
~~~
Will keep working on it.

marmarek:
Looks like either grub or its signature download failed. Try removing those files (or whole vmm-xen directory) and retry.

awokd:
> removing those files (or whole vmm-xen directory

That did it, running `make qubes` now. I got a Makefile Error 4 on the run prior to it complaining about the GRUB file. I had originally unset GNUPGHOME before `make get-sources`, but looked at `scripts/get-sources-extra` and saw it did something with that variable so moved the unset after. Not sure if the Error 4 meant a broken download or had something to do with that sequence.

I think these are good then as long as there are no surprises later on in the build.

awokd:
`make qubes` finished with no errors.

PR #573, opened 2018-02-07
rxvt yum -> dnf

awokd:
and line breaks


awokd:
Not seeing anything that would block this from working in R4.0 but untested.

PR #576, opened 2018-02-09
windows VM qvm-prefs should be HVM and no kernel

adubois:


adubois:
Too many outstanding issues. Will submit a single PR.

PR #580, opened 2018-02-10
consolidate private and root disk resize docs

awokd:


configuration/resize-disk-image.md
```diff
@@ -1,113 +1,89 @@
...
+### Template disk image
 
-1048576 MiB is the maximum size which can be assigned to private storage through Qube Manager.
+If you want install a lot of software in your TemplateVM, you may need to increase the amount of disk space your TemplateVM can use. 
+*Make sure changes in the TemplateVM between reboots don't exceed 10G.*
 
-To grow the private disk image of an AppVM beyond this limit, `qvm-volume` can be used:
+1.  Make sure that all the VMs based on this template are shut down (including netvms etc).
+2.  Resize the *root image* using Qubes version specific procedure below.
+   You may instead resize the *private image* with the appropriate commands if your software installs to `/home` for example.
```
marmarek: This isn't the best place (or wording) for such suggestion. Those points are clearly about TemplateVM, and installing software in `/home` there doesn't make much sense, because it won't be propagated to TemplateBaseVMs. Resizing private image makes more sense for non-TemplateVMs.

configuration/resize-disk-image.md
```diff
@@ -1,113 +1,89 @@
...
+3.  If any netvm/proxyvm used by this template is based on it, set template's netvm to none.
```
marmarek: Again, this is definitely not needed in 4.0, and probably also not needed in 3.2.

configuration/resize-disk-image.md
```diff
@@ -1,113 +1,89 @@
...
 
-**It is critical to use the same (or bigger for some safety margin) size in truncate call compared to resize2fs call.
-Otherwise you will lose your data!** 
-Then reset kernel options back to default:
+### Resize a StandaloneVM Root Image
 
-~~~
-qvm-prefs -s <vm-name> kernelopts default
-~~~
+Another way to increase the size of is to turn your TemplateVM into a StandaloneVM.
```
marmarek: Well, creating StandaloneVM does not make it larger... But it may be more convenient if you need a lot of extra software in just one VM. I think this sentence needs to be reworded.

configuration/resize-disk-image.md
awokd: Yes, I don't know what I was thinking here.

configuration/resize-disk-image.md
```diff
@@ -1,113 +1,89 @@
...
 
-### Private disk image (R4.0)
+### Template disk image
 
-1048576 MiB is the maximum size which can be assigned to private storage through Qube Manager.
+If you want install a lot of software in your TemplateVM, you may need to increase the amount of disk space your TemplateVM can use. 
+*Make sure changes in the TemplateVM between reboots don't exceed 10G.*
 
-To grow the private disk image of an AppVM beyond this limit, `qvm-volume` can be used:
+1.  Make sure that all the VMs based on this template are shut down (including netvms etc).
```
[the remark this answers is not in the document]
awokd: Verified as needed on R3.2- with a running AppVM the template's root was locked and the resize wouldn't register (`resize2fs` claimed there was nothing to do) until I shutdown the AppVM.

configuration/resize-disk-image.md
awokd: I left it in the R3.2 section in case there was some other locking mechanism going on.

awokd:
This is our second pass through these documents, might as well finish up what's left on this PR!

PR #581, opened 2018-02-10
USB: delete duplicate content

awokd:
and add heading back


awokd:
Travis link check is successful. ^

PR #582, opened 2018-02-11
document https://github.com/QubesOS/qubes-app-yubikey

adrelanos:


andrewdavidwong:
Looks like there are some typos in the Travis log.

@marmarek:

1. The Travis log says, "If these are false-positives, simply commit as is and they will be ignored in the future." What is the correct way for me to fix these typos while merging a PR? I don't want them to be whitelisted as false positives, since they're genuine positives. If I check out the branch, add a commit on top fixing them, then merge the branch into master, will they be whitelisted?

2. The Travis log also says, "In addition, the following suspicious spellings were eliminated: [list of typos]." What does "eliminated" mean here? Does that mean the typos were fixed, or does it mean they were allowed through because they were previously whitelisted?

marmarek:
@andrewdavidwong the spellchecker works by comparing list of unknown words in master and PR. So, if anything that isn't in dictionary is already in master, spellchecker will not complain about it. The second point is about unknown words that were found in master, but not in PR. Which may mean that some typos were present in master and are fixed by the PR.

security/yubi-key.md
```diff
@@ -40,52 +40,67 @@ password (other than your login password!) and apply the configuration.
...
-   replace `PASSWORD-HASH` with SHA1 hash for your YubiKey-linked password (other
-   than your standard Qubes password). You can calculate it using this command:
+6. Paste your hashed password (other than your standard Qubes password)  into
+`/etc/qubes/yk-keys/yk-login-pass-hashed.hex` in dom0.
+
+   You can calculate your hashed password using this command:
 
        echo -n "PASSWORD" | openssl dgst -sha1
 
+   (Replace `PASSWORD` with your actual password.)
```
marmarek: Technically this is ok, but it will leave the password in shell history. I recommend this instead:
```
read password
echo -n "$password" | openssl dgst -sha1
```

andrewdavidwong:
> the spellchecker works by comparing list of unknown words in master and PR. So, if anything that isn't in dictionary is already in master, spellchecker will not complain about it. The second point is about unknown words that were found in master, but not in PR. Which may mean that some typos were present in master and are fixed by the PR.

Ok, so merging a PR that contains a typo won't whitelist that typo for another PR as long the typo is fixed by a subsequent commit before Travis checks the other PR, right?

marmarek:
> Ok, so merging a PR that contains a typo won't whitelist that typo for another PR as long the typo is fixed by a subsequent commit before Travis checks the other PR, right?

Exactly.

adrelanos:
All feedback resolved.

* https://github.com/QubesOS/qubes-doc/pull/582#pullrequestreview-98203644

andrewdavidwong:
@adrelanos, there are still typos in the Travis log:
```
    +Derivating
    +TempalteVM
    +commnand
    +reponse
    +yubikeys
```

adrelanos:
Typos fixed.

`YubiKeys` spelling taken from https://www.yubico.com/products/yubikey-hardware/.

PR #583, opened 2018-02-12
copy-from-dom0 line breaks, version clarification

awokd:
https://github.com/QubesOS/qubes-issues/issues/3571


awokd:
Also spelled out the `echo -n dom0 > /var/run/qubes/qubes-clipboard.bin.source` step.

PR #585, opened 2018-02-13
Add note to enable qubes-firewall when template is used as FirewallVM

taradiddles:
When a FirewallVM is based on a template that doesn't have the qubes-firewall feature enabled the qubes settings gui of VMs connected to this FirewallVM will show an error message:

"The 'work' qube is network connected to 'sys-firewall', which does not
support firewall!
You may edit the 'work' qube firewall rules, but these will not take any
effect until you connect it to a working Firewall qube." 

The system seems to be fully functional though.

fix: enable the template's qubes-firewall feature.

ref:  https://groups.google.com/forum/#!topic/qubes-users/i08ZSoQLPUQ


marmarek:
This message is intentional, when you do not have qubes-firewall service installed/enabled in the template. The problem is that installation of that service is not advertised to dom0. Fix here:
https://github.com/QubesOS/qubes-core-agent-linux/pull/93

taradiddles:
Closing  / not relevant anymore after Marek's fix

PR #586, opened 2018-02-13
config-files 4.0 updates

awokd:
Per https://github.com/QubesOS/qubes-issues/issues/3260

Not sure if there are any new scripts or `guid.conf` values for R4.0 but I can add if there is a list somewhere.


tasket:
@awokd - Per issue 3503 a `qubes-firewall.d' feature was also added soon after qubes-firewall-user-script was fixed.

The order of execution is:
1. /etc/qubes/qubes-firewall.d
2. /rw/config/qubes-firewall.d
3. /rw/config/qubes-firewall-user-script

PR #588, opened 2018-02-15
doc.md add links to R4 content

awokd:
Not sure if this is the best approach, or if the content should be pulled out into new /doc files.


andrewdavidwong:
Ideally, the content would be rewritten into a documentation page, since a blog post (or article) is different from a documentation page in presentation and audience. In reality, however, it's perhaps unlikely that anyone will take the time to do a good job of converting these into documentation pages. Even if someone were willing to take the time, these are probably already close enough to their  documentation versions that the time would be better spent on more neglected documentation areas.

andrewdavidwong:
These should be relative links, @awokd.

PR #590, opened 2018-02-15
qubes-builder link to ISO build instead of Archlinux template

awokd:
although the screenshots are nice


awokd:
Actually, going to merge this one with the other PR.

PR #592, opened 2018-02-16
disk-trim add systemd fstrim and efi initrd rebuild step

awokd:
from https://mail-archive.com/qubes-users@googlegroups.com/msg19082.html and followups

### DIFF
--- a/configuration/disk-trim.md
+++ b/configuration/disk-trim.md
@@ -32,16 +32,27 @@ Configuration
 ----------
 
 In all versions of Qubes, you may want to set up a periodic job in `dom0` to trim the disk.
+This can be done with either systemd (weekly only) or cron (daily or weekly).
 
-This can be done from a terminal as root, by creating a `trim` file in `/etc/cron.daily` (or `/etc/cron.weekly`).
-Add the following contents:
+ * **Systemd**
 
-```
-#!/bin/bash
-/sbin/fstrim --all
-```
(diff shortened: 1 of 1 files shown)


awokd:
Intentionally skipping the `dracut -H` option in the EFI section per the man page comment about it:
>This has the
drawback, that you can’t put the disk on another controller or machine,
and that you can’t switch to another root filesystem, without recreating
the initramfs image. The usage of the --hostonly option is only for
experts and you will have to keep the broken pieces. 

Left it in the GRUB config section because it was in **bold** so it seemed pretty important to someone at some time, but I don't know if it's still the best recommendation?

andrewdavidwong:
@marmarek: I know that you're very busy, and a lot of qubes-doc PRs are assigned to you. Please feel free to delegate them to other Qubes team members, if you like. (I assign them to you when I'm not certain who else has the required technical expertise to review the content.)

awokd:
And if there's something else I can/should be doing, please let me know. Travis claims a bunch of these are "failed". The ones that aren't spelling related (like this one) should be good with a re-run of Travis but it looks like the only way I can do that is to close and re-open the PRs which I imagine will generate spam.

awokd:
Added -H because it appears to be required for non-default keyboard layouts and LUKS: https://ask.fedoraproject.org/en/question/80389/cannot-open-luks-container-with-rescue-initramfs/

PR #593, opened 2018-02-17
Remove paragraph on KDE desktop switcher

pierwill:
I suggest this paragraph be removed. It is outdated and overly specific for a "Getting Started" page.


andrewdavidwong:
Since the same point is also made [here](https://www.qubes-os.org/doc/security-guidelines/#observing-security-contexts) (and probably elsewhere in our documentation), I think it's fine to remove this. (If this were the only place that it were mentioned, it would be important that we move the information to a more appropriate location rather than simply deleting it.)

PR #599, opened 2018-02-23
consistent dracut commands across docs

awokd:


marmarek:
Any specific reason for `-H` option?

awokd:
Per https://github.com/QubesOS/qubes-doc/pull/592#issuecomment-366460426
Haven't tested with a non-default keyboard layout so I'm not completely sure it's required, but looks like it was still an issue with Fedora 23.

marmarek:
Ah, so this (or related thing) might also fix  https://github.com/QubesOS/qubes-issues/issues/2971?

awokd:
Yes, that definitely looks related but I'm surprised, thought it was only the older Fedora versions that had the problem. Not sure it would be a good idea to default the installer to use `-H`; sometimes it's useful to be able to install Qubes on one PC then transfer the disk to another.

awokd:
So, from https://github.com/marmarek/qubes-core-admin-linux/commit/bcf7c9e978681991c776b8bb1c8172af83e0d606, we want to keep `-H` in here, correct?

marmarek:
In fact this all makes no difference. Because `/usr/lib/dracut/dracut.conf.d/01-dist.conf` contains `hostonly="yes"`.

awokd:
OK! Can this be merged as is then, to capture the `uname -r` changes, or should I remove `-H` first?

marmarek:
In the light of this finding, I'd remove -H from here. Otherwise it will be misleading (suggesting that -H is really needed to get hostonly mode).

PR #601, opened 2018-02-24
http-filtering-proxy 4.0 updates

awokd:
I've been attempting to spread out the work across multiple logically distinct commits. Hopefully that helps make them easier to review.

I'd still like replace the [link inside](https://groups.google.com/group/qubes-devel/attach/39c95d63fccca12b/proxy.tar.gz?part=0.1) this document with one pointing somewhere on qubes-os.org instead. Can I do that by changing the link to `archive` and uploading `proxy.tar.gz` to qubes-attachment (in a separate PR)?


andrewdavidwong:
> I'd still like replace the [link inside](https://groups.google.com/group/qubes-devel/attach/39c95d63fccca12b/proxy.tar.gz?part=0.1) this document with one pointing somewhere on qubes-os.org instead. Can I do that by changing the link to `archive` and uploading `proxy.tar.gz` to qubes-attachment (in a separate PR)?

Technically speaking, yes, but looking at https://github.com/QubesOS/qubes-issues/issues/1202, it looks like we were mostly guessing about the accuracy of this document when we added it. Perhaps the document and the contents of that tarball should be reviewed more carefully. (They really should have been in the first place.)

What do you think, @marmarek?

awokd:
I ran through the procedure on both 3.2 and 4.0 so it's verified working, but I know I'd be more comfortable with some extra scrutiny on the script etc. in that blob.

marmarek:
Documentation page itself looks ok. But that script archive indeed is problematic in the current form. I don't like linking directly to google archive, without any checksum. And in fact pointing to some specific instance of the archive (with given hash) makes sense only if we somehow verify that version... That's ~300 lines of python.
I don't think there is anything malicious (there is no communication or data processing from either AppVM side, or the internet). But there may be some bugs leading to leaking some traffic. And even if we (or someone) find one, and even fix it, there is no easy way to distribute this update, since it's just an archive, not a proper package.

To keep qubes-doc a trustworthy place, I'd either:
 - remove this doc
 - add yet another warning about the above, while putting the archive into qubes-attachment repository

Alternatively, this could be properly maintained in a separate repository, and distributed as a proper package instead of just tarball. But we'd need someone who would maintain it...

andrewdavidwong:
My vote is for removing the doc and adding it back if (and only if) someone volunteers to maintain the script as a proper package. What do you think, @awokd?

awokd:
Sounds good; I know I don't like linking to random off-site blobs in official docs either. Could maybe be a candidate for the community docs.

andrewdavidwong:
> Sounds good; I know I don't like linking to random off-site blobs in official docs either. Could maybe be a candidate for the community docs.

:+1: 

> One more thought on this though- could it be merged before you delete so if it's brought back online here at some point, my work isn't lost?

Sure.

PR #603, opened 2018-02-27
Add behind-proxy

awokd:
Did an "RFC" on qubes-users but got [no replies](https://mail-archive.com/qubes-users@googlegroups.com/msg19372.html), so putting this here.


configuration/behind-proxy.md
```diff
@@ -0,0 +1,89 @@
...
+    If not, perform the following so updates can be downloaded through the proxy.
+
+    First, determine which template your UpdateVM is using:  
+     a. Go to a `dom0` terminal prompt.  
+     b. Enter `qubes-prefs` and look for `updatevm`.  
+     c. Enter `qvm-prefs <UpdateVMName>` and look for `template`.  
+    Perform the following steps from inside that template.
+
+    Add `proxy=http://10.0.0.1:8080` to the bottom of `/etc/apt/apt.conf.d/71proxy` in Debian templates.
+    In Fedora, add it to `/etc/dnf/dnf.conf` instead, after the line that says `### QUBES END ###`.
```
marmarek: This will not work in Qubes 4.0, because templates have no direct network access at all. Better to configure proxy in updates-proxy service. See `Upstream` setting in `tinyproxy.conf(5)`. This would require modifying `/etc/tinyproxy/tinyproxy-updates.conf`, including making it persistent (a file in `/rw` + `cp ... /etc/tinyproxy/tinyproxy-updates.con` in `/rw/config/rc.local`?).

awokd: I thought adding the proxy setting here would result in it only getting used by the UpdateVM (e.g. sys-net) based on this template, and the templates themselves would actually continue to use qubes-rpc (4.0) or restricted networking (3.2) to pull updates from the UpdateVM. What am I missing?

awokd: Thinking about it a bit more, I could see this proxy setting might override the one for `tinyproxy-updates` when using a debian-9 based sys-net to supply updates for the debian-9 template. Would it work to combine the approaches and apply the `apt.conf.d / dnf.conf` modifications strictly to the UpdateVM (instead of the underlying template) via your persistence approach?

marmarek: When you update template, dnf/apt is used only inside that template. And dnf/apt in template needs to be configured to use updates proxy - overriding proxy setting there will break updates. On the updates proxy side (sys-net) it is just http proxy, nothing more. Modifying dnf/apt configuration there have no effect on what is used by template to download updates.

There are some cases where modifying apt/dnf config to use some (corporate?) proxy in TemplateBasedVM is desirable - for example if you want to (temporarily) install package directly there, instead of installing it permanently in template. Another case is just checking for updates to decide whether template update is needed.
For those cases, it might be ok to set proxy setting _before_ `### QUBES BEGIN ###` (or in file ordered before 71proxy). This way, the setting will be used only in TemplateBasedVM, but will be overridden in TemplateVM to updates proxy. This needs to be tested, I'm not sure about apt and dnf behavior with duplicated settings.

andrewdavidwong:
Note that there is a minor conflict in `doc.md`.

awokd:
Updated.

awokd:
Is this document more suited to the Community docs vs. official?

andrewdavidwong:
> Is this document more suited to the Community docs vs. official?

The topic seems suitable for the official docs, in my opinion.

andrewdavidwong:
> > Is this document more suited to the Community docs vs. official?
> 
> The topic seems suitable for the official docs, in my opinion.

Given the change in policy (https://github.com/QubesOS/qubes-issues/issues/4693) (and the fact that the Qubes devs haven't had time to re-review it in well over a year), I think this would now be more suitable for the Community docs.

PR #605, opened 2018-02-28
Update firewall.md documentation

adubois:
address issues related to the combined use of iptables and nftables as well as fix various typos and unclear parts.

### DIFF
--- a/security/firewall.md
+++ b/security/firewall.md
@@ -182,7 +182,7 @@ network 192.168.x.0/24.
 **1. Route packets from the outside world to the FirewallVM**
 
 From a Terminal window in sys-net VM, take note of the 'Interface name' and
-'IP address' on which you want to expose your service (i.e. eth0, 192.168.x.x)
+'IP address' on which you want to expose your service (i.e. ens5, 192.168.x.x)
 
 ` ifconfig | grep -i cast `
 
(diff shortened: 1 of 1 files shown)


security/firewall.md
```diff
@@ -204,6 +206,10 @@ the service
 
 > Note: If you want to expose the service on multiple interfaces, repeat the
   steps described in part 1 for each interface
+  
+> Note: On Qubes R4, nftables is also used which imply that nft rules also need to be set. Qubes OS has defined a `qubes-firewall` table with a forward chain.
+
+`nft add rule ip qubes-firewall forward meta iifname eth0 ip daddr 10.137.0.x tcp dport 443 ct state new counter accept`
```
marmarek: Hmm, qubes-firewall service (by design) should only take care about traffic outgoing from a VM, not incoming. So, I'd consider the need for this line a bug. Let me fix it right now. And then replace this line with reference for appropriate package version.

adubois: @marmarek, I suspect you're very busy. This line is to document this exact use case (expose service in AppVM and open flow inbound through sys-net then sys-firewall) The sub-section title is something like expose a service to the outside world. I took care to lock it as much as possible at every stage.

marmarek: The line as it is looks ok until the bug is resolved. Please add this ticket reference: https://github.com/QubesOS/qubes-issues/issues/3644

adubois:
@marmarek (mentioning you so you get an event in your queue)... also I’m not sure I still learn github workflow....

PR #606, opened 2018-02-28
remove mac-changer content

awokd:
per @tasket in https://www.mail-archive.com/qubes-users@googlegroups.com/msg19555.html


awokd:
Going to add note about NetworkManager instead

PR #607, opened 2018-03-01
anonymizing-your-mac-address version notes

awokd:
and mention disabling NetworkManager for macchanger to work.
From discussion [here](https://www.mail-archive.com/qubes-users@googlegroups.com/msg19566.html).


tasket:
I'd recommend against refreshing the macchanger-based method and implying it will work if-user-prepares-x until it can be tested and shown to work; alternately, display a comment that it may not work as intended. I tried this method a little over a year ago and it allowed the MAC address to revert to original when, for instance, the system woke from sleep and the system hooks didn't exist to address the issue. Subsequently after system updates, users started to report that the scripts stopped working at all.

The general macchanger approach is something I've tried on various distros for over 5 years now. No matter what the specifics were, it couldn't handle some state transitions (such as sleep/wake) resulting in the original address being leaked. It probably can work for special distros like TAILS that support a very limited envelope of system states (boot, running with no persistence, poweroff) but Qubes is not like that.

andrewdavidwong:
Related issues: https://github.com/QubesOS/qubes-issues/issues/938, https://github.com/QubesOS/qubes-issues/issues/3405

awokd:
Thanks @tasket and @andrewdavidwong . Looks like there was consensus in #938 to remove the macchanger content once the updated NM hit all the templates, so we should be in the clear.

PR #608, opened 2018-03-02
How to connect to Microsoft Exchange with Qubes OS

one7two99:
this howto describes how to setup an Email AppVM which can connects to your Microsoft Exchange Server and provides an gateway to use your favorite Linux (!) apps to access the exchange content.
Currently (1st draft) only describing the access to your mails, calendar sync will be added soon
The functionality is based on Davmail.

### DIFF
--- a/configuration/connect-to-ms-exchange.md
+++ b/configuration/connect-to-ms-exchange.md
@@ -1,0 +1,233 @@
+# How to connect to Microsoft Exchange with Qubes OS
+
+## Introduction
+This Howto describes how you can connect to your Exchange Server with Qubes OS.
+It is based on using a Exchange-to-IMAP-Gateway which is named Davmail.
+
+The howto covers:
+
+- Installation/Configuration of Davmail
+- Configuration to use Thunderbird
+- Configuration to use neomutt
+
+the following VMs will be mentioned in this howto:
+
+- t-mail = TemplateVM
+- my-vault = ApppVM which is used for SplitGPG
+- my-workmail = AppVM
+
+TODO:
+
+- Howto use SplitGPG with neomutt
+- Howto work with attachments (PDF/Pictures) and HTML mails in neomutt
+
+## Davmail
+
+install required packages for Davmail
+
+`sudo dnf install java-9-openjdk`
+
+Download Davmail in an AppVM and qvm-copy-to-vm it over to the mail template vm
+in the AppVM which has internet access:
+
+`wget https://kent.dl.sourceforge.net/project/davmail/davmail/4.8.3/davmail-linux-x86_64-4.8.3-2554.tgz`
+`qvm-copy-to-vm t-mail davmail-linux-x86_64-4.8.3-2554.tgz`  
+
+In the TemplateVM:
+
+`sudo tar -xvzf ~/QubesIncoming/<AppVM>/davmail.*.tgz -C /opt`
+`sudo mv /opt/davmail* /opt/davmail`
+
+tell your template VM where to find the GPG Vault VM 
+
+`export QUBES_GPG_DOMAIN=my-vault`
+
+test access via SplitGPG (this command should list your secret-keys in the Vault-AppVM)
+
+`qubes-gpg-client -K` 
+
+
+Shutdown TemplateVM and create a new AppVM based on the TemplateVM
+in dom0
+
+`qvm-shutdown --wait t-mail`
+`qvm-create --template t-mail --label blue my-workmail`
+
+Open Qubes Setting for the new AppVM and increase storage capacity
+Private storage max. size: 20480 MiB
+
+Start AppVM and continue configuration there
+
+`qvm-run --auto my-workmail gnome-terminal`
+
+Start davmail
+
+`/opt/davmail/davmail.sh &`
+
+- Main Tab: Echange Protocol: Auto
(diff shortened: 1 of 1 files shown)


configuration/connect-to-ms-exchange.md
```diff
@@ -0,0 +1,233 @@
...
+## Davmail
+
+install required packages for Davmail
+
+`sudo dnf install java-9-openjdk`
+
+Download Davmail in an AppVM and qvm-copy-to-vm it over to the mail template vm
+in the AppVM which has internet access:
+
+`wget https://kent.dl.sourceforge.net/project/davmail/davmail/4.8.3/davmail-linux-x86_64-4.8.3-2554.tgz`
```
marmarek: Is there any way to verify authenticity of the archive? Some signature? Checksum on a page hosted somewhere else (independently of sourceforge.net)? It would be good to include some method here. If there are none, then add a warning about that.

configuration/connect-to-ms-exchange.md
```diff
@@ -0,0 +1,233 @@
...
+`sudo tar -xvzf ~/QubesIncoming/<AppVM>/davmail.*.tgz -C /opt`
+`sudo mv /opt/davmail* /opt/davmail`
+
+tell your template VM where to find the GPG Vault VM 
+
+`export QUBES_GPG_DOMAIN=my-vault`
+
+test access via SplitGPG (this command should list your secret-keys in the Vault-AppVM)
+
+`qubes-gpg-client -K`
```
marmarek: TemplateVM shouldn't really need access to GPG Vault VM. This step should be performed where keys will be accessed - my-workmail.

configuration/connect-to-ms-exchange.md
```diff
@@ -0,0 +1,233 @@
...
+
+- Main Tab: Echange Protocol: Auto
+- Main Tab: OWA (Exchange) URL: https://owa.domain.com/owa/
+- Main Tab: [ ] Local POP port
+- Advanced Tab: Default windows domain: <YOUR-DOMAIN>
+- Advanced Tab: [x] Disable Update Check
+
+tell your AppVM VM where to find the GPG Vault VM 
+
+`echo "export QUBES_GPG_DOMAIN=my-vault" >> /home/user/.bashrc`
```
marmarek: Better write `my-vault` to `/rw/config/gpg-split-domain` (which is sourced by `/etc/profile.d/qubes-gpg.sh`). This way it will work seamlessly with `qubes-gpg-client-wrapper`.

configuration/connect-to-ms-exchange.md
```diff
@@ -0,0 +1,233 @@
...
+`qvm-run --auto my-workmail gnome-terminal`
+
+Start & configure davmail (see above)
+
+`/opt/davmail/davmail.sh &`
+
+### Setting up neomutt
+Clone mutt-wizard
+
+`git clone https://github.com/LukeSmithxyz/mutt-wizard.git ~/.config/mutt`
```
marmarek: Again, some warning about source verification could be useful. Also, you may want to link to https://www.qubes-os.org/doc/mutt/ for those unwilling to download random script from the internet...

configuration/connect-to-ms-exchange.md
```diff
@@ -0,0 +1,233 @@
...
+[Repository work-local]
+type = Maildir
+localfolders = ~/.mail/work
+```
+
+Create a link to the certfiles
+you could also put the correct path to the certfiles into .offlineimaprc but we're using the default paths which are also configures via mutt-wizard
+
+`cd /etc/ssl/certs/`
+`sudo ln -s ca-bundle.trust.crt ca-certificates.crt`
```
marmarek: Since this is in AppVM, the change will not persist after restart. Better fix path in the config.

andrewdavidwong:
(Updated PR title to be more descriptive.)

configuration/connect-to-ms-exchange.md
one7two99: Yes, thank you for pointing this out.
I have asked in the Davmail Mailinglist and they provide Checksums on the same sourceforge page where you download the packages.
For example for the current version:
https://sourceforge.net/projects/davmail/files/davmail/4.8.3/checksums.md5/download

configuration/connect-to-ms-exchange.md
one7two99: Ok, but this step was made in the template VM.
Fixing the path might be a better solution.

configuration/connect-to-ms-exchange.md
one7two99: As the idea was to provide a simple(r) approach getting (neo)mutt but also Exchange configured I think it might make sense to split this how-to into sections:

1) Qubes + Davmail -> to get Exchange connected
2) Qubes + Neomutt -> How to get neomutt working under Qubes OS using the neomutt-wizard script but also by creating an own configuration from scratch, covering three examples:
Googlemail / Outlook.com / Corporate Exchange (connected via DavMail, referencing to the 1st how-to).

There is no need describing everything in one How-to, what do you think?

marmarek: Sounds good.

andrewdavidwong:
This PR is being closed because there has been no activity on it in over a year. However, if anyone would like it to be reopened, simply leave a comment saying so, and we'll be happy to reopen it. Thank you.

PR #615, opened 2018-03-10
disk-trim downgrade "recommended" to "you may want"

awokd:
per https://github.com/QubesOS/qubes-issues/issues/3686

### DIFF
--- a/configuration/disk-trim.md
+++ b/configuration/disk-trim.md
@@ -59,7 +59,7 @@ However, inside App and Template qubes, the `discard` mount option is on by defa
 
 If you are using Qubes with LVM, you may also want to set `issue_discards = 1` in `/etc/lvm/lvm.conf`.
 Setting this option will permit LVM to issue discards to the SSD when logical volumes are shrunk or deleted.
-In R4.x, LVM Logical volumes are frequently deleted (every time a disposable VM is shut down, for example) so setting `issue_discards = 1` is recommended if using an SSD.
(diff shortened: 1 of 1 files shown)


awokd:
1. It's not security vs. convenience or SSD life. It's security vs. secure with a theoretical information disclosure that's been around for 10 years yet I'm unaware of any instance in which it was useful information for an attack (i.e. the hair splitting).
2. If I repeat the information and link covered in the introduction for this single setting, I should also do the same for each setting in the fstrim section as well as those in the LUKS section because they all can result in passing through discards to disk. For consistency, I should actually remove the extraneous mention of it from this one...

Should I do that and see if the introduction needs rewording? Maybe it's not being made clear there.

andrewdavidwong:
> It's not security vs. convenience or SSD life. It's security vs. secure with a theoretical information disclosure that's been around for 10 years yet I'm unaware of any instance in which it was useful information for an attack (i.e. the hair splitting).

I was just referring to what @marmarek said in https://github.com/QubesOS/qubes-issues/issues/3686 ("So, this is a trade-off between security and SSD life time. We choose to have safer default.") I don't really have a horse in this race. I think your disagreement on this point is with him, and should probably be debated in that issue (if at all).

Given the apparent disagreement, and the fact that @marmarek also said that this wording would be a good idea (https://github.com/QubesOS/qubes-issues/issues/3686#issuecomment-372022739), I'm going to merge it as is.

> If I repeat the information and link covered in the introduction for this single setting, I should also do the same for each setting in the fstrim section as well as those in the LUKS section because they all can result in passing through discards to disk. For consistency, I should actually remove the extraneous mention of it from this one...
Should I do that and see if the introduction needs rewording? Maybe it's not being made clear there.

Not big enough of a deal to spend any more time on right now, IMHO.

PR #619, opened 2018-03-23
installation-guide add note and link about size of R4 ISO

awokd:
Could possibly add to the downloads page too, but will leave that up to you.


andrewdavidwong:
Please use relative links, in accordance with the Documentation Guidelines.

PR #628, opened 2018-04-02
uefi-troubleshooting add efi=no-rs section

awokd:
The nice thing about standards like UEFI is there are so many ways to choose to implement them. Sadly, many are wrong...

### DIFF
--- a/troubleshooting/uefi-troubleshooting.md
+++ b/troubleshooting/uefi-troubleshooting.md
@@ -143,3 +143,40 @@ Ctrl-Alt-F2), or booting from installation media in "Rescue a Qubes system" mode
 
        efibootmgr -v -c -u -L Qubes -l /EFI/qubes/xen.efi -d /dev/sda -p 1 "placeholder /mapbs /noexitboot"
 
+Installation freezes before getting to Anaconda
+-----------------------------------------------
+
+On some early, buggy UEFI implementations, you may need to disable EFI under Qubes completely.
+This can sometimes be done by switching to legacy mode in your BIOS/UEFI configuration.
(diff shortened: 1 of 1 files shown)


troubleshooting/uefi-troubleshooting.md
```diff
@@ -143,3 +143,40 @@ Ctrl-Alt-F2), or booting from installation media in "Rescue a Qubes system" mode
...
+5. Execute the following commands:
+   ~~~
+   cp -R /mnt/sysimage/boot/efi/EFI/qubes /mnt/sysimage/boot/efi/EFI/BOOT
+   mv /mnt/sysimage/boot/efi/EFI/BOOT/xen.efi /mnt/sysimage/boot/efi/EFI/BOOT/BOOTX64.efi
+   mv /mnt/sysimage/boot/efi/EFI/BOOT/xen.cfg /mnt/sysimage/boot/efi/EFI/BOOT/BOOTX64.cfg
+   ~~~
+6. Go back to `tty6` (Ctrl-Alt-F6) and click `Reboot`.
+7. Continue with setting up default templates and logging in to Qubes.
+
+Whenever there is a kernel or Xen update for Qubes, you will need to follow these [other steps above](/doc/uefi-troubleshooting/#boot-device-not-recognized-after-installing) because your system is using the fallback UEFI bootloader in `[...]/EFI/BOOT` instead of directly booting to the Qubes entry under `[...]/EFI/qubes`.
```
marmarek: `/usr/lib/kernel/install.d/90-xen-efi.install` attempts to handle non-default dirs too. But it rely on ability to query EFI variables, which is impossible with `efi=no-rs`. So yes. this paragraph is needed.

PR #629, opened 2018-04-02
Add FAQ item about coreboot

vincentadultman:
As touched on in IRC with holger and marek. New and existing users are often confused about what hardware to buy and whether coreboot is superior to vendor bios.  I've tried to keep this entry concise (it could already be considered too long) what I've tried to do is include the most often asked questions without writing a tutorial which is probably out of scope for the Qubes website and better left to the coreboot wiki. I've ignored libreboot here as it is not part of the certified hardware requirements and I don't recall a report of success.

### DIFF
--- a/about/faq.md
+++ b/about/faq.md
@@ -115,6 +115,10 @@ At the same time, due to the smart use of Xen shared memory, our GUI implementat
 
 Please refer to [this page](/doc/vm-sudo/).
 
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
Minor point: Did you mean "trustworthy" rather than "trustable"?

PR #631, opened 2018-04-03
Documentation for awesome WM

3hhh:
incl. a guide for focus steal hardening

### DIFF
--- a/customization/awesome.md
+++ b/customization/awesome.md
@@ -1,0 +1,187 @@
+---
+layout: doc
+title: awesome
+permalink: /doc/awesome/
+redirect_from:
+- /en/doc/awesome/
+- /doc/awesome/
+---
+
+# Using awesome in dom0
+
+## Qubes-specific features
+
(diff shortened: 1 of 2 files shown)


woju:
BTW don't you want to package that focus-oriented snippet in our package for people to be able to just `require` it in their configs?

3hhh:
> BTW don't you want to package that focus-oriented snippet in our package for people to be able to just require it in their configs?

I even considered providing a patch file for all changes.

However even the snippet contains options that users may choose upon (do I want to clear focus on that event or not?) and considering the probably small awesome user base of technical people I thought it would be better to try to explain the approach and let everyone make his own decision in the end. That might even help people with other awesome versions.

If there's some demand for it in the future though, I'll consider creating a patch that enables/disables the hardening depending on some variable being set in the rc.lua.

woju:
In principle I have nothing against, but it would be up to i3 guys to write it.

PR #634, opened 2018-04-21
Update fedora-minimal.md

trueriver:
I found the existing doc misleading, as it claims sudo is not installed. This misled me to try to install sudo, which dnf said was already there. Also, the fact that the template does not have passwordless root is not a bug and therefore the relevant install should not be described as a "fix" in the doc.

I have also included explicit instructions to enable passwordless root, which will also serve as an example to newcomers of how to install the packages listed in the following section.

### DIFF
--- a/managing-os/templates/fedora-minimal.md
+++ b/managing-os/templates/fedora-minimal.md
@@ -66,17 +66,33 @@ A comprehensive guide to customizing the minimal template is available [here][GU
 Qubes 4.0
 ---------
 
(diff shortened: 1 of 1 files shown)


trueriver:
This is my first ever pull request, not only here but on any git based system. I will appreciate feedback if I am unaware of any cultural or technical practices.

andrewdavidwong:
Just a handy tip for future reference: You can close an issue from a commit by including a keyword and issue in the commit message. In your case, it would be:
```
Fixes QubesOS/qubes-issues#3833
```

PR #635, opened 2018-04-22
faq: add steps to edit update repositories

awokd:
per https://github.com/QubesOS/qubes-issues/issues/3737#issuecomment-383401560


andrewdavidwong:
Well, this won't work around the problem in normal cases (e.g., when the server has a temporary problem).

andrewdavidwong:
I'm going to wait on a response to https://github.com/QubesOS/qubes-issues/issues/3737#issuecomment-383418890 before merging this, and even then, it will probably be better to add the revised version (the same as what appears in the announcement) to avoid confusion.

awokd:
Makes sense. Should I close?

awokd:
ok

PR #637, opened 2018-04-25
Notice for qubes-tunnel testing

tasket:
Issues #3503 #3520

### DIFF
--- a/configuration/vpn.md
+++ b/configuration/vpn.md
@@ -12,10 +12,11 @@ redirect_from:
 How To make a VPN Gateway in Qubes
 ==================================
 
-    ## Note on new qubes-tunnel availability!
-    A streamlined VPN option is in testing phase and is recommended for most users on Qubes 3.2 and 4.0. It is currently [available here](https://github.com/tasket/qubes-doc) and qubes-users discussion is [here](https://groups.google.com/d/msgid/qubes-users/ee24f104-efbc-23f7-aca3-6be86104ddaf%40posteo.net).
+## Note on new qubes-tunnel availability!
+
+A streamlined VPN option is in testing phase and is recommended for most users on Qubes 3.2 and 4.0. It is currently [available here](https://github.com/tasket/qubes-doc) and qubes-users discussion is [here](https://groups.google.com/d/msgid/qubes-users/ee24f104-efbc-23f7-aca3-6be86104ddaf%40posteo.net).
     
-    The following instructions will be deprecated and replaced once the new solution is finalized.
+The following instructions will be deprecated and replaced once the new solution is finalized.
 
 ---
 ---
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
I'm *slightly* worried about the wording here:
> and is recommended for most users on Qubes 3.2 and 4.0

@marmarek, any opinion?

tasket:
I chose that wording because I feel that attempts to use the new scripts (in service already under different naming) will lead to fewer issues for users than the existing method... despite the testing status.

But please feel free to re-word it as you see fit.

tasket:
Also wondering, since qubes-tunnel appears to be more of a direct contribution (per Patrick's indication in #3503 ) instead of a 'Qubes-contrib' optional package, whether it should just be added to the Qubes testing repositories.

configuration/vpn.md
```diff
@@ -12,6 +12,15 @@ redirect_from:
 How To make a VPN Gateway in Qubes
 ==================================
 
+## Note on new qubes-tunnel availability!
+
+A streamlined VPN option is in testing phase and is recommended for most users on Qubes 3.2 and 4.0. It is currently [available here](https://github.com/tasket/qubes-doc) and qubes-users discussion is [here](https://groups.google.com/d/msgid/qubes-users/ee24f104-efbc-23f7-aca3-6be86104ddaf%40posteo.net).
```
andrewdavidwong: Is `https://github.com/tasket/qubes-doc` really the correct URL?

andrewdavidwong:
> I chose that wording because I feel that attempts to use the new scripts (in service already under different naming) will lead to fewer issues for users than the existing method... despite the testing status.
> 
> But please feel free to re-word it as you see fit.

I guess the reason for my uncertainty is that I don't know what our official position regarding that package is. It seems somewhat strange to recommend something from a non-official source to *all* users. I'll leave it for @marmarek to decide.

> Also wondering, since qubes-tunnel appears to be more of a direct contribution (per Patrick's indication in #3503 ) instead of a 'Qubes-contrib' optional package, whether it should just be added to the Qubes testing repositories.

Also for @marmarek to decide. If it's a contrib package, though, we should make sure to follow our procedure for such packages:
https://www.qubes-os.org/doc/package-contributions/

configuration/vpn.md
tasket: oops... that should be 'tasket/qubes-tunnel'

tasket:
For reference, Marek and Patrick comments have it as a direct Qubes inclusion:
https://github.com/QubesOS/qubes-issues/issues/3503#issuecomment-361114105

I was just wondering if qubes-testing was a better repository in this case. Then it wouldn't seem as strange to recommend it (or to use wording asking users to help test).

andrewdavidwong:

PR #640, opened 2018-04-30
dispvm-customization add static Disposable VMs for sys-*

awokd:


customization/dispvm-customization.md
```diff
@@ -57,6 +57,42 @@ It is possible to change the settings for each new Disposable VM (DispVM). This
 4.  Shutdown the qube (either by `poweroff` from qube's terminal, or `qvm-shutdown` from dom0 terminal).
 
 
+### Using static Disposable VMs for sys-* ###
+
+You can use a static DispVM for `sys-*` as long as it is stateless.
+For example, a `sys-net` using DHCP or `sys-usb` will work.
+`sys-firewall` will not work unless you have no custom rules set, because per VM rules are stored in a configuration file inside that AppVM.
```
marmarek: I'd phrase it the other way around. Right now it suggest stateless sys-firewall is rare case, which I think is not true - custom rules are rare case.

customization/dispvm-customization.md
```diff
@@ -57,6 +57,42 @@ It is possible to change the settings for each new Disposable VM (DispVM). This
...
+To create one:
+
+~~~
+qvm-create -C DispVM -l red <sys-VMName>
+qvm-prefs <sys-VMName> virt_mode hvm
```
marmarek: This is needed only for sys-net and sys-usb, but not for sys-firewall.

customization/dispvm-customization.md
```diff
@@ -57,6 +57,42 @@ It is possible to change the settings for each new Disposable VM (DispVM). This
...
+qvm-service <sys-VMName> meminfo-writer off
```
marmarek: And this too.

customization/dispvm-customization.md
```diff
@@ -57,6 +57,42 @@ It is possible to change the settings for each new Disposable VM (DispVM). This
...
+qvm-pci attach --persistent <sys-VMName> dom0:<BDF>
```
marmarek: I think this require a little explanation - how to get those devices. For example by listing old VM. Also, usb controllers needs `-o no-strict-reset=True`.

one7two99:
Hello Marek,

thank you for the interesting post, I think having "non-persistent" sys-vms is a good thing as for example the sys-net VM is basically the entry point for an atacker.

On 05/03 02:42, Marek Marczykowski-Górecki wrote:

> +You can use a static DispVM for `sys-*` as long as it is stateless.
> +For example, a `sys-net` using DHCP or `sys-usb` will work.
> +`sys-firewall` will not work unless you have no custom rules set, because per VM rules are stored in a configuration file inside that AppVM.
> [...]

what about storing the firewall configuration in dom0 and "pushing" it to the disposable sys-firewall-VM:
Just a quick idea (haven't tried this):


1) edit your Firewall configuration in a file in dom0
   => vi /home/user/FW-Config.iptables

2) qvm-copy the iptables configuration file to the disposable sys-firewall AppVM
   => qvm-copy-to-vm sys-firewall /home/user/FW-Config.iptables

3) Restore firewall configuration in sys-firewall
   => qvm-run --auto --user root sys-firewall 'iptables-restore /home/user/QubesIncoming/FW-Config.iptables


The downsize is that you might find to add a mechanism that adds additional rules as soon as more AppVMs are started up.
Maybe it is possible to create a custom firewall file for each AppVM which will be added to the sys-fireall iptable after the AppVM has been started?

[799]

awokd:
Haven't had a chance yet to fully review the comments, but is the above still on-topic for my "dispvm-customization add static Disposable VMs for sys-*" PR?

awokd:
>>    what about storing the firewall configuration in dom0 and "pushing" it to the disposable sys-firewall-VM:

> This is exactly how standard firewall rules works in qubes (technical details are different, but the idea is the same).

Does this mean we can also use a static DispVM to replace `sys-firewall` even when AppVM rules have been set, only exception being if someone has edited them manually inside `sys-firewall` itself?

Updated doc per Marek's earlier comments, but would like to add some explanation of this too if it's correct.

marmarek:
> Does this mean we can also use a static DispVM to replace sys-firewall even when AppVM rules have been set, only exception being if someone has edited them manually inside sys-firewall itself?

Yes, exactly. Rules set in VM settings (firewall tab, qvm-firewall tool) are stored in dom0 and sent to appropriate VM when needed.

awokd:
Updated.

PR #649, opened 2018-05-12
Add FAQ to Statistics page

andrewdavidwong:
Requested by: QubesOS/qubes-issues#3858

@woju, I'd like to improve the answer to the question: "What kinds of data do you collect about Qubes users?" In order to do this, I need a bit more information:

 - How are the IPv4 addresses kept? (For example, are they hashed or encrypted? Does a human ever see them?)
 - How long do we keep the IPv4 addresses that we collect?
 - How do we count the number of requests from each IPv4 address?

CC: @marmarek, @rootkovska

### DIFF
--- a/about/statistics.md
+++ b/about/statistics.md
@@ -23,7 +23,7 @@ Since the graph is updated daily, the bar for the current month will be very low
 
 ### How is the userbase estimated?
 
-We simply count the number of unique IPv4 addresses that connect to the Qubes update servers each month (except for Tor connections; see below).
+We simply count the number of unique IPv4 addresses that connect to the Qubes update servers each month (except for Tor connections; see [below][tor-methodology]).
 
 ### How has the methodology for counting Tor users changed?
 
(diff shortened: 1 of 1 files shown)


about/statistics.md
```diff
@@ -10,10 +10,52 @@ redirect_from:
...
+We cross-reference the list of connecting IP addresses with the official database of Tor exit nodes in order to distinguish Tor and clearnet IPs and requests.
+For this purpose, an IP address counts as belonging to a Tor exit node if there was a Tor exit node active for that address in the past 24 hours.
+
+### What kinds of data do you collect about Qubes users?
+
+The only data we collect are:
+ - The IPv4 addresses that connect to the Qubes update servers
+ - The number of requests from each IPv4 address
+
+We do not collect any other kinds of data about Qubes users.
```
woju: Well, we keep standard httpd-access and -error log and if someone skilled wanted, they could guess something from it (at least which templates are in use, possibly how many VMs are running). We keep the log for both statistics and audit reasons. I'm not sure how long we keep them, but I think at least 2 months, but in any case the statistics script does not touch logs older than 3 months.

The log is kept unencrypted and that includes IP addresses. They are looked at by human only in case of audit/postmortem (this happened at least once). There are no other non-volatile places where IP addresses are stored.

Graph plotting is separate and it does not touch the log, only .json file, which, as mentioned, contains no IP addresses.

**How the counting works:** Once in a day (on the first day of month, twice, to finish the previous month) the script parses the log and updates .json file, which is linked from /statisticts/. The script looks at each line of the log, and if the URL is right (repo manifest) and the timestamp is from the specified month, then it extracts release and IP address. IP address is looked up in local cache of exit node database. If it matches exit node, the "tor" request counter is incremented. If not, the "plain" request counter is incremented _and_ the address is added to a in-memory set. When finished, the Tor estimate is computed and the two numbers: "plain" (the number of IPs in the set) and "tor" (the estimate above) are updated and the .json file is written to disk.

The source code is available. I'd welcome the public scrutiny of both methodology and any privacy issues. For the privacy problems, maybe the bug report could be sent in private through security@.

andrewdavidwong: > We keep the log for both statistics and audit reasons. I'm not sure how long we keep them, but I think at least 2 months, but in any case the statistics script does not touch logs older than 3 months.
> 
> The log is kept unencrypted and that includes IP addresses. They are looked at by human only in case of audit/postmortem (this happened at least once).

I think it's very important that we improve our practices in this area. We should set clear policies for data access and retention, store the data securely, and securely destroy it as soon as it's no longer needed. I'll make a separate issue about this.

andrewdavidwong:
@woju: Thanks. Updated to take your feedback into account. Let me know if you see any more room for improvement before I merge.

PR #651, opened 2018-05-16
edit security-critical-code.md

pierwill:
- move each sentence to a new line
- make edits for clarity

### DIFF
--- a/doc.md
+++ b/doc.md
@@ -230,7 +230,7 @@ System
 ------
  * [Qubes OS Architecture Overview](/doc/architecture/)
  * [Qubes OS Architecture Spec v0.3 [PDF]](/attachment/wiki/QubesArchitecture/arch-spec-0.3.pdf)
- * [Security-critical elements of Qubes OS](/doc/security-critical-code/)
+ * [Security-critical Code in Qubes OS](/doc/security-critical-code/)
  * [Qubes Core Admin](https://dev.qubes-os.org/projects/core-admin/en/latest/)
  * [Qubes Core Admin Client](https://dev.qubes-os.org/projects/core-admin-client/en/latest/)
  * [Qubes Admin API](/news/2017/06/27/qubes-admin-api/)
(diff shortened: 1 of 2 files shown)


andrewdavidwong:
Thanks, @pierwill! The page still needed some more work, so I've added a further revision.

---

@marmarek, I found this section very confusing:
> Additionally when we consider attacks that originate through a compromised network domain and target the VMs connected to it. Those attacks do not apply to domains connected to other network domains (Qubes allows more than one network domains), or those with networking disabled (Dom0 is not connected to any network by default).
> 
>  - Xen network PV frontends
>  - VM’s core networking stacks (core TCP/IP code)

I couldn't tell whether the document was saying that these *are* or *are not* security-critical components. If you check the current state of the page (without applying this PR), you'll see that they're clearly included in the "Security-Critical Third-Party Components", but my understanding is that they're not security-critical (at least VMs' core networking stacks aren't).

pierwill:
I'm wondering if the section "Cooperative Covert Channels Between Domains" belongs in this document. It seems out of place in a doc that introduces and develops the concept of Trusted Code Base. It is an important note on security philosophy, of course—

Though it's not immediately clear where, maybe another place in the docs could be found for this section? Maybe as a note in `architecture.md`?

andrewdavidwong:
> I'm wondering if the section "Cooperative Covert Channels Between Domains" belongs in this document. It seems out of place in a doc that introduces and develops the concept of Trusted Code Base. It is an important note on security philosophy, of course—
> 
> Though it's not immediately clear where, maybe another place in the docs could be found for this section? Maybe as a note in `architecture.md`?

I think the natural place for it would be [Understanding and Preventing Data Leaks](https://www.qubes-os.org/doc/data-leaks/) (which is why I also included a link to this page at the end).

What do you think, @marmarek?

marmarek:
> I think the natural place for it would be Understanding and Preventing Data Leaks (which is why I also included a link to this page at the end).
> 
> What do you think, @marmarek?

I agree.

andrewdavidwong:
Actually, when I went to move that section to [Understanding and Preventing Data Leaks](https://www.qubes-os.org/doc/data-leaks/), I realized that all of the content is already explained more clearly and in more detail in the latter document, so I've simply removed the section from [Security-critical Code](https://www.qubes-os.org/doc/security-critical-code/) instead.

andrewdavidwong:
@marmarek, I think this is ready for your review now.

PR #653, opened 2018-05-20
Make dd bs flag work for BSD/MacOS

soycode:
BSD `dd` uses lowercase letters for the bs argument. https://www.unix.com/man-page/FreeBSD/1/dd/

This way of noting seemed most concise, but if you prefer having it more descriptive or as a full second command/section let me know. Thanks!


andrewdavidwong:
This way of writing it will confuse a lot of people. Perhaps we could just use `bs=1024` or something so that it's cross-platform.

soycode:
Good call, changed.

PR #657, opened 2018-05-25
Fix typing error qubes-prefs instead qvm-prefs

donob4n:
https://groups.google.com/d/msgid/qubes-users/LDJHlUH--3-0%40keemail.me?utm_medium=email&utm_source=footer


marmarek:
qubes-prefs is for global prefs, qvm-prefs is per-vm. Global `default_dispvm` is a default value for per-vm `default_dispvm`. While here indeed is a doc bug (qubes-prefs don't accept vm name), generally it might be useful to change the default one - if you want to delete default dispvm.
Pro tip: you can list `default_dispvm` for all qubes using `qvm-ls -O name,default_dispvm`

PR #667, opened 2018-06-25
explain how to publish the key

niccokunzmann:
Problem: The key needs to be published in order for the
  signature-checker to work. (my guess)
Solution: Add a section about uploading the key to a server.

### DIFF
--- a/basics_dev/code-signing.md
+++ b/basics_dev/code-signing.md
@@ -78,6 +78,14 @@ uid                  Bilbo Baggins <bilbo@shire.org>
 sub   4096R/69B0EA85 2013-03-13
 ~~~
 
+Upload the Key
+--------------
+
+For others to find the public key, please upload it to a server.
+
+```
+gpg --send-keys 69B0EA85
+```
 
 Using PGP with Git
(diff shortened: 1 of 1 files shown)


tasket:
You may also want to mention associating the key to your Github account (after uploading to keyserver step) as I think that is part of the checking process.

marmarek:
> You may also want to mention associating the key to your Github account (after uploading to keyserver step) as I think that is part of the checking process.

Currently not (we use trust-on-first-use approach here), but it is used for "Verified" badge, which is nice :)
We may add such check in the future.

As for actual command - it might be a good idea to specify keyserver too, because in some systems there is no default specified.

niccokunzmann:
@andrewdavidwong I think, I addressed the key server.

niccokunzmann:
@tasket Do you have a link in mind? I only found one for uploading the private key which might not be what is intended.

tasket:
@niccokunzmann For github the place for adding your Public key is here:
https://github.com/settings/keys

The `New GPG key` button will let you paste your Public key.

PR #669, opened 2018-06-26
NM: start VPN automatically, fail-close

gasull:
When configuring with NetworkManager, make VPN start automatically and fail-close the connection.

Editing the connection `VM uplink eth0` won't persist between VM reboots.

### DIFF
--- a/configuration/vpn.md
+++ b/configuration/vpn.md
@@ -68,6 +68,8 @@ Set up a ProxyVM as a VPN gateway using NetworkManager
    And substitute "XXXXXXXXXXXXXX" for the actual password.
    The contents of `passwd-file.txt` may differ depending on your VPN settings.  See the [documentation for `nmcli up`](https://www.mankier.com/1/nmcli#up).
    
+   Troubleshooting:  If you notice that the VPN doesn't autostart on hardware bootup but it does when restaring the VPN VM, try adding a slight delay in seconds in `/rw/config/rc.local` before the `nmcli` command, e.g. `sleep 2`.
+   
 5. (Optional) Make the network fail-close for the AppVMs if the connection to the VPN breaks.
 
    Edit `/rw/config/qubes-firewall-user-script` and add these lines:
--- a/configuration/vpn.md
+++ b/configuration/vpn.md
@@ -50,11 +50,39 @@ Set up a ProxyVM as a VPN gateway using NetworkManager
 
 3. Set up your VPN as described in the NetworkManager documentation linked above.
 
-4. Configure your AppVMs to use the new VM as a NetVM.
+4. Make your VPN start automatically.
+
+   Edit `/rw/config/rc.local` and add these lines:
+   
+   ```bash
+   # Automatically connect to the VPN
+   nmcli connection up file-vpn-conn passwd-file /rw/config/NM-system-connections/secrets/passwd-file.txt
+   ```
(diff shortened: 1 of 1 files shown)


tasket:
@gasull The submitted re-write covers the fail-close part. See:
https://github.com/tasket/qubes-tunnel
https://github.com/QubesOS/qubes-issues/issues/3503

I recommend using qubes-tunnel instead if the connection is openvpn (although its configurable enough to use different protocols).

IIRC, auto-connect can be set in the NM GUI. The last time I tried I recall it was erratic and often failed (but that was several NM revisions ago).

andrewdavidwong:
Assigning to @marmarek to decide whether this should be closed for the reasons @tasket gave or merged.

gasull:
@tasket wrote:
> I recommend using qubes-tunnel instead if the connection is openvpn (although its configurable enough to use different protocols).

If that's the recommended way of doing it, why isn't qubes-tunnel mentioned in this document?

> IIRC, auto-connect can be set in the NM GUI. The last time I tried I recall it was erratic and often failed (but that was several NM revisions ago).

No, it doesn't work at all.  I should have said that in the commit message.  You can edit the uplink vm0 connection, but it will be rewritten upon reboot, despite being in the /rw/ tree.  I tried editing manually, with the GUI, and even changing the file permissions of the file.  It was always rewritten regardless.

tasket:
@gasull I did a quick PR to the doc so it would mention the beta, but I think Qubes devs wanted to rely on the qubes-users list to let people know (I'm not really sure). Marek and Patrick were very busy with core 4.0 and Whonix release issues at the time so I didn't want to press the issue.

It would be great if you join in testing qubes-tunnel and report your experience.

BTW, if user testing is to be considered essentially complete I can add the deb and rpm stuff to qubes-tunnel this week. But I also have a few code tweaks to post before then, mainly to support non-password client cert configurations.

gasull:
Sorry, I'm just getting started with Qubes, so I want to keep things simple.  I would rather not test software in beta state right now since I'm still figuring out how to use Qubes.

I think my solution is simpler than downloading a Git project and installing it; it just involves a one-liner in `/rw/config/rc.local` and another one-liner secrets file.  Pretty simple.  `qubes-tunnel` might be great once it's part of Qubes and everything is done automagically.

tasket:
Understandable. FWIW, the parent project Qubes-vpn-support uses rc.local to start and doesn't require installation into template (though it still needs download and install into proxyVM). I'm doing the release candidate this week.

The Network Manager approach probably would have been documented more if there were more demand for it. Unfortunately NM's openvpn handling has been a turn off for many and not many VPN services provide working configs for it (often NM will mangle an ovpn config when importing it). Its easier just to have users download and use the configs directly, or download a custom client program.

configuration/vpn.md
```diff
@@ -50,11 +50,40 @@ Set up a ProxyVM as a VPN gateway using NetworkManager
...
 3. Set up your VPN as described in the NetworkManager documentation linked above.
 
-4. Configure your AppVMs to use the new VM as a NetVM.
+4. (Optional) Make your VPN start automatically.
+
+   Edit `/rw/config/rc.local` and add these lines:
+   
+   ```bash
+   # Automatically connect to the VPN once Internet is up
+   for i in {1..50}; do ping -c1 1.1.1.1 &> /dev/null && break; done
```
marmarek: There is nm-online tool for this.

gasull: Changed it.

PR #672, opened 2018-07-01
Add IPv6 failsafe

tasket:
While we're waiting for qubes-tunnel and new instructions to be approved, it would be a good idea to extend failsafe to IPv6. This is not an issue for R3.2, but R4.0 does allow IPv6 traffic.


marmarek:
> While we're waiting for qubes-tunnel and new instructions to be approved

I guess it's blocked on me, but can't find where exactly (I guess you're not talking about #637). Can you point me relevant issue?

andrewdavidwong:
I don't think it's you, @marmarek. See https://github.com/QubesOS/qubes-issues/issues/3503 and specifically https://github.com/QubesOS/qubes-issues/issues/3503#issuecomment-401633425.

PR #681, opened 2018-07-21
Make VPN work on suspend-resume

gasull:
Make VPN work after laptop is suspended and later resumed.


gasull:
It seems to have no effect.  I guess I didn't wait enough on previous tests to make the VPN key go stale.  I'm closing this.

PR #685, opened 2018-07-25
Fix auto-VPN, improve kill switch

gasull:
`nm-online` doesn't seem reliable and many times it didn't work, and piping it to `systemd-cat` would log a small binary blob.  The new method has worked for me 100% of the time.

The old `iptables` rules for the kill switch gets in the way for connections to my laptop's LAN, since they were processed before the rule `-A FORWARD -m conntrack --ctstate --ctstate RELATED,ESTABLISHED -j ACCEPT`, and therefore established connections to the LAN would break.  The new rule doesn't have this problem and it's shorter.

### DIFF
--- a/configuration/vpn.md
+++ b/configuration/vpn.md
@@ -56,8 +56,11 @@ Set up a ProxyVM as a VPN gateway using NetworkManager
    
    ```bash
    # Automatically connect to the VPN once Internet is up
-   nm-online --quiet --wait-for-startup
-   nmcli connection up file-vpn-conn passwd-file /rw/config/NM-system-connections/secrets/passwd-file.txt
+   while ! ping -c 1 -W 1.1.1.1; do
+      sleep 1
+   done
+   PWDFILE="/rw/config/NM-system-connections/secrets/passwd-file.txt"
+   nmcli connection up file-vpn-conn passwd-file $PWDFILE
    ```
    You can find the actual "file-vpn-conn" in `/rw/config/NM-system-connections/`.
    
@@ -73,8 +76,8 @@ Set up a ProxyVM as a VPN gateway using NetworkManager
 
    Edit `/rw/config/qubes-firewall-user-script` and add these lines:
    ```bash
-   #    Block forwarding of connections through upstream network device
-   #    (in case the vpn tunnel breaks):
+   # Block forwarding of connections through upstream network device
+   # (in case the vpn tunnel breaks)
    iptables -I FORWARD -o eth0 -j DROP
    iptables -I FORWARD -i eth0 -j DROP
    ```


tasket:
Configuring a sys-vpn to handle Mixed LAN & VPN traffic is not a wise idea unless you really know what you're doing. The simpler case that tightly binds a specific network to a qube is more secure and in keeping with Qubes philosophy; the different network/gateway/protocol can be accessed by choosing a different proxyVM. For that reason I don't think its a good generic example to have in a Qubes doc. Its better to show how to create an exception to network separation in technical footnotes.

Also, using a numerical-position insert here is somewhat unreliable. And the block for eth0 input is removed, which I also disagree with as a generic example.

gasull:
> Configuring a sys-vpn to handle Mixed LAN & VPN traffic is not a wise idea unless you really know what you're doing. The simpler (current) case that tightly binds a specific network to a qube is more secure and in keeping with Qubes philosophy

Do you have an article that explains this?  I'm very much interested.

In my case I only open 2 ports I need, and I forward them to a particular qube.

> For that reason I don't think its a good generic example to have in a Qubes doc.

There's no mention about the laptop's LAN in the document, only in my commit/PR message.

> using a numerical-position insert here is somewhat unreliable.

True.  I'll improve on that.

> the block for eth0 input is removed, which I also disagree with as a generic example.

I believe `iptables -I FORWARD -i eth0 -j DROP` was redundant since the default rule is `-P FORWARD DROP`.

tasket:
Qubes is designed to strongly isolate (or try to isolate) anything a VM can process -- its the one unique thing a qube does by default. Network examples go back to Joanna's 2011 paper that explains a Tor VM that "forward only the Tor traffic", in addition to separate firewall VMs for separate physical networks.

A VM configured for encrypted VPN + anti-leak firewall more than implies network separation; the example shouldn't be open to unencrypted clearnet in that case and there should be no opportunity for users to _unwittingly_ happen upon the mixed-net condition the proposal is allowing. This is especially true when multiple downstream VMs with widely varying applications, misfeatures and possible malware will be connecting through that proxyVM and users will leave them running in the background absentmindedly.

Its great that you can make these exceptions for yourself and operate your VPN VM in different access modes with that understanding. Many other users, however, won't arrive at the VPN doc with the same understanding or intentions so the wrinkle proposed here could enable a nasty breech for some.

Demand for this kind of configuration seems low... most who know they need the exception seem capable of adding it themselves; There exist a few old threads in qubes-users about making a LAN exception so maybe the topic could be left to the mail list or added as a footnote.

tasket:
> I believe iptables -I FORWARD -i eth0 -j DROP was redundant since the default rule is -P FORWARD DROP.

There are numerous ACCEPTs between that rule and the end of the chain that don't take tunneling into account. So this doesn't work.

gasull:
@tasket wrote;
> > I believe iptables -I FORWARD -i eth0 -j DROP was redundant since the default rule is -P FORWARD DROP.
> 
> There are numerous ACCEPTs between that rule and the end of the chain that don't take tunneling into account. So this doesn't work.

OK, I just added another rule to block incoming traffic too.

I'm not sure I'm understanding all of @tasket 's objections.  The text/code changed has 2 blocks.

The [first block](https://github.com/QubesOS/qubes-doc/pull/685/files#diff-b5237521e2a374e41f3f8a7cdad167daL59) makes autoconnect work 100% of the time.

The [second block](https://github.com/QubesOS/qubes-doc/pull/685/files#diff-b5237521e2a374e41f3f8a7cdad167daL74) improves on previous `iptables` commands so existing connections that don't use the VPN won't be disrupted.  This code is therefore more generic and works in most situations.

It wouldn't make sense to me that, because the VPN connection breaks, we have to disrupt the already existing connections that don't use the VPN (the `-A FORWARD -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT` rule).

tasket:
The first block seems like it would be OK (I wouldn't know for sure... neither of my VPN accounts work with Network Manager's wobbly openvpn plugin).

---

The second part is still risky and inappropriate for a Qubes security example. The mistake is in positioning `conntrack` as a guard function on equal footing with the VPN client, and not accounting for adverse conditions like MITM or appVMs opportunistically trying to establish and maintain clearnet links.

> It wouldn't make sense to me that, because the VPN connection breaks, we have to disrupt the already existing connections that don't use the VPN...

Yes, we have to disrupt non-VPN connections.... The doc shows the creation of a VM expressly for a VPN link with an automatic anti-leak firewall. In the Qubes context, that strongly implies the VM is _dedicated_ to the VPN. Having submitted issues about OS lapses in forwarding control, I know its possible for surreptitious clearnet links to form; I'd expect the VPN firewall to mitigate that by stopping the non-VPN traffic.

OTOH, submitter is advancing a threat model where clearnet links and all products of the configuration are intended by the user; that is insisting on the naive perspective.

This is poor form:
> Then, assuming you counted two lines in the output, edit...

Proposal here is convoluted with unnecessary numeric positions for inserts, and introduces an inconsistency with the scripted solution's firewall rules. Another problem is that no attempt has been made to explain how the new rules can be utilized as the author intends; I infer enough inconsistency between what the submitter suggests in the discussion and the code to surmise it couldn't be explained properly.

tasket:
> Otherwise we are actually attacking the user by disrupting (denial of service) their connections to their home appliances.

You're way off base. No one is disrupting user-initiated connections because the blocking rules are enabled at startup; user would have no _routine_ access to LAN devices through a VPN VM ... unless the VPN terminated on that LAN (which is an option BTW). And its worth repeating they can use sys-firewall for this purpose.

Furthermore, its clear you don't use the firewall rules in the way they're configured in your submission. Your dialogue indicates you don't auto-run them at VM startup, and instead wait to manually run them after making LAN connections. So you've introduced complex behavior into the firewall without trying to describe it to the user, and because the code submission doesn't fit the use case I don't think you could even describe it if you tried. For this reason the firewall part of the submission should be disregarded; its insistent and earnest without being serious.

tasket:
Marek has personally recommended overriding that rule for securing a tunnel a number of times... specifically mentioning the rules you're trying to replace. And those rules have been vetted by more than one Qubes project developer.

`conntrack` is not a strong security function and its use here is all downsides. It doesn't even enable the use case you claim.

marmarek:
Main documentation for VPN VM should be about routing everything (connected to that VM) through VPN. If VPN is down, no traffic is allowed. If you want LAN access, you can connect a VM to another network provider (like sys-firewall). If you want to mix those two, that's non-standard configuration, that should be handled separately. The documentation can mention where to put custom rules about exceptions etc. But by default it should assume that everything go through VPN in the least error-prone way.

gasull:
> If VPN is down, no traffic is allowed. If you want LAN access, you can connect a VM to another network provider (like sys-firewall).

I would swear I read in the docs that sys-firewall was supposed to connect to the VPN ProxyVM.  So it would be:
```
sys-net <-> VPN <-> sys-firewall <-> AppVMs
```
But I can't find this now.  Was this configuration ever recommended or am I completely wrong and sys-firewall shouldn't connect to VPN?

I can think of many situations when someone would want LAN and Internet at the time, like printing a webpage, or using Kodi with Kore as a remote.  What's the Qubes way of doing this?

I will revert the second block of changes as requested by @tasket and you, but please answer my two questions because I'm confused now.  Thanks.

andrewdavidwong:
> I can think of many situations where someone would want LAN and Internet at the same time, like printing a webpage, or using Kodi with Kore as a remote. What's the Qubes way of doing this?

In the printer case, it would probably be something like:

1. Print to PDF in browsing VM.
2. Copy PDF to printer VM.
3. Print.

jpouellet:
> In the printer case, it would probably be something like:
> 
> 1. Print to PDF in browsing VM.
> 2. Copy PDF to printer VM.
> 3. Print.

Hopefully with a qubes.PdfConvert between 2 and 3 if you're concerned about PDF parsing bugs in your printer - which is probably a worthwhile thing to consider depending on your threat model. Printers have a large attack surface, potentially some crappy microcontroller-class CPU without hardware memory protection, very likely the ability for an attacker to persist in firmware, potentially network access, and over time probably access to lots of sensitive documents.

PR #686, opened 2018-07-25
software-update-dom0: replace uname -r with version number

awokd:
Think this was my error.

### DIFF
--- a/common-tasks/software-update-dom0.md
+++ b/common-tasks/software-update-dom0.md
@@ -138,9 +138,9 @@ If the update process does not automatically do it (you should see it mentioned 
 from the update command), you may need to manually rebuild the EFI or grub config depending on which
 your system uses.
 
-EFI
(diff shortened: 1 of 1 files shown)


jpouellet:
Why? Unless I'm missing something, `uname -r` expands to the correct thing without the user needing to know (and without the docs getting outdated).

andrewdavidwong:
> Why? Unless I'm missing something, `uname -r` expands to the correct thing without the user needing to know (and without the docs getting outdated).

Presumably, we all agree that `uname -r` expands to the *current* release name. The question is whether, at that particular point in the process, the current release name has already been updated or whether it's still the old one. If it's the new one -- i.e., the one to which the user wishes to upgrade -- then `uname -r` should be fine. If it's the old one, then `uname -r` would be incorrect.

awokd:
Right, what @andrewdavidwong said. At that point in the process, the user's looking to build the "future" release, not the current one.

PR #690, opened 2018-08-07
Qubes-Whonix 14

adrelanos:


andrewdavidwong:
The page won't work when you remove the YAML header. We probably want to make this a redirect instead, and update the URL in the doc index.

PR #691, opened 2018-08-07
Whonix 14

adrelanos:


andrewdavidwong:
This would mean that Qubes OS no longer supports Whonix 13, which is probably not true and different from saying that Whonix 13 has reached EOL. We generally don't *replace* versions of supported TemplateVMs; we only *add* them. Note that there are several EOL distro versions on this list, because Qubes OS still supports them, even though they're no longer supported by their own maintainers.

CC: @marmarek

andrewdavidwong:
To clarify, I think there are two meanings of "support" here:

- "Qubes OS supports Whonix 13" means something like, "Qubes OS accommodates Whonix 13 existing as a TemplateVM in Qubes OS such that things are secure from the Qubes side of things."
  (Note that this says nothing about the internals of Whonix 13, which may have already reached EOL.)

- "The Whonix Project supports Whonix 14" means something like, "The Whonix Project fixes bugs and adds features internal to Whonix 14 in order to ensure that it functions according to the goals of the Whonix Project."

This distinction allows all of the following to be true statements:

- "Qubes 3.2 supports Whonix 13."
- "Qubes 4.0 supports Whonix 13."
- "Qubes 3.2 supports Whonix 14."
- "Qubes 4.0 supports Whonix 14."
- "Whonix no longer supports Whonix 13."
- "Whonix supports Whonix 14."

PR #693, opened 2018-08-09
Small add in installing and updating

p-suedo:
Hello Qubes! I am pretty new to this, but was reading about StandaloneVM, and think the command is only for 3.x, not 4.x. I tried copying from other material, hope I did it right! Figure is better to try myself since its little thing. Thanks!

### DIFF
--- a/common-tasks/software-update-vm.md
+++ b/common-tasks/software-update-vm.md
@@ -189,6 +189,12 @@ In order to create a standalone VM you can use a command line like this (from co
 qvm-create <vmname> --standalone --label <label>
 ~~~
 
+or in 4.0:
+
+```
+qvm-create --class StandaloneVM --label <label> --property virt_mode=hvm <vmname>
+```
+
 ... or click appropriate options in the Qubes Manager's Create VM window.
 
 Using more than one template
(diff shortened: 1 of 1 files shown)


p-suedo:
Hello! I looked at docs, but want knowing if it is better to put 3.2 and 4.0 after title 'Standalone Vms' - according to docs, it looks like the whole page should be split into 3.2 and 4.0, but this is one small difference. I see earlier on same page the did this way ('Reverting VM' has 3.2 and 4.0 sections), and I think makes sense since just small differences on long page.

andrewdavidwong:
That's fine, thanks.

common-tasks/software-update-vm.md
```diff
@@ -168,7 +168,29 @@ However, a compromise of a template affects only a subset of all your AppVMs (in
...
+-   VMs used for development (devel environments require a lot of \*-devel packages and specific devel tools)
+
+-   VMs used for installing untrusted packages.
+    Normally you install digitally signed software from Red Hat/Fedora repositories, and it's reasonable that such software has non malicious *installation* scripts (rpm pre/post scripts).
+    However, when you would like to install some packages from less trusted sources, or unsigned, then using a dedicated (untrusted) standalone VM might be a better way.
+
+In order to create a standalone VM you can use a command line like this (from console in Dom0):
+
+```
+qvm-create --class StandaloneVM --label <label> --property virt_mode=hvm <vmname>
```
marmarek: Technically, `virt_mode=hvm` isn't needed for every StandaloneVM. But it makes sense if one want to use kernel from within.

PR #696, opened 2018-08-19
Fix implication that Windows support isn't done

strugee:


services/qmemman.md
```diff
@@ -19,7 +19,7 @@ Traditionally, Xen VMs are assigned a fixed amount of memory. It is not the opti
 The [tmem](https://oss.oracle.com/projects/tmem/) project provides a "pseudo-RAM" that is assigned on per-need basis. However this solution has some disadvantages:
 
 -   It does not provide real RAM, just an interface to copy memory to/from fast, RAM-based storage. It is perfect for swap, good for file cache, but not ideal for many tasks.
--   It is deeply integrated with the Linux kernel. When Qubes will support Windows guests natively, we would have to port *tmem* to Windows, which may be challenging.
+-   It is deeply integrated with the Linux kernel. Since Qubes supports Windows guests natively, we would have to port *tmem* to Windows, which may be challenging.
```
andrewdavidwong: It's not clear to me which sense of the word "natively" is being used here, so I can't tell whether the new version of this sentence is accurate. @marmarek?

marmarek: Well, the whole tmem section is mostly irrelevant here, as the technology doesn't exists for us:
 - it is [not security supported](https://lists.xen.org/archives/html/xen-announce/2012-09/msg00006.html)
 - it never gets out of [experimental state](https://wiki.xenproject.org/wiki/Xen_Project_Release_Features), for at least 5 years

So, the argument about Windows support is very minor. Anyway, we do have Qubes Windows Tools (although unmaintained now), which can be considered "Windows guest native support". And it doesn't support neither of tmem or qmemman, so the argument is simply false (but indeed adding support for qmemman would be much easier than for tmem).

strugee: I'm not sure what's desired here - it feels like the Windows remark is a bit of a red herring and we want to minimize it? Perhaps something like:

    It is deeply integrated with the Linux kernel. This would cause complications since Qubes natively supports Windows guests.

awokd: Or just delete "When Qubes will support Windows guests natively, we would have to port *tmem* to Windows, which may be challenging." entirely?

andrewdavidwong:

PR #698, opened 2018-08-21
Setup example of an encrypted second HDD on R4.0

Galland:
This is too complex and error prone not to facilitate users' lives with an example of the few commands needed to perform this.


tasket:
You may want to consider adding a GUI example as well. The `gnome-disks` tool very nicely handles both LUKS and volume groups.

andrewdavidwong:
Please don't use `#` for comments when giving command-line instructions. It's ambiguous with a root prompt. You can instead make those comments text outside and preceding each code block instruction line.

Galland:
@tasket there is no gnome-disks in dom0

tasket:
@Galland gnome-disks isn't pre-installed in dom0, but it is available from the repository.

PR #699, opened 2018-08-25
Add Desktop Environment Section

asddsaz:


andrewdavidwong:
We already have pages for these:

- https://www.qubes-os.org/doc/kde/
- https://www.qubes-os.org/doc/xfce/

Please submit PRs for the existing pages if you want to modify them.

andrewdavidwong:
Oh, I see now that you were trying to duplicate those existing pages into new subdirectories. Please don't.

- They have the same URLs as the existing pages, which will probably break things.
- There's no good reason to do that.
- Even if there is a reason, you should move, not copy, so that we don't end up with duplicates.

PR #707, opened 2018-09-20
Update dispvm.md

e6lk7dqzm83p:
Adding documentation, per the suggestion in https://github.com/QubesOS/qubes-issues/issues/4160

### DIFF
--- a/common-tasks/dispvm.md
+++ b/common-tasks/dispvm.md
@@ -60,6 +60,17 @@ A Disposable VM launched from the Start Menu inherits the NetVM and firewall set
 Note that changing the "NetVM" setting for the system default DVM Template *does* affect the NetVM of DispVMs launched from the Start Menu.
 Different DVM Templates with individual NetVM settings can be added to the Start Menu. 
 
+**Important Notes:**
+Some DVM Templates will automatically create a menu item to launch a DVM, if you do not see an entry and want to add one please use the command:
+
+    qvm-features deb-dvm appmenus-dispvm 1
+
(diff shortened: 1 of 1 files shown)


common-tasks/dispvm.md
```diff
@@ -60,6 +60,17 @@ A Disposable VM launched from the Start Menu inherits the NetVM and firewall set
 Note that changing the "NetVM" setting for the system default DVM Template *does* affect the NetVM of DispVMs launched from the Start Menu.
 Different DVM Templates with individual NetVM settings can be added to the Start Menu. 
 
+**Important Notes:**
+Some DVM Templates will automatically create a menu item to launch a DVM, if you do not see an entry and want to add one please use the command:
```
marmarek: Those having automatically created menu entries are prepared during system installation, together with other default qubes. Basically the below command is run at the initial configuration time for those DVM Templates. It doesn't happen to any DVM template created manually.
But I think "Some DVM Templates" in this place is enough, instead of lengthy explanation.
Note that this step is already covered in https://www.qubes-os.org/doc/dispvm-customization/, but I'm ok with duplicating it here.

e6lk7dqzm83p: Hi,
Thank you for approving this.

As I noted in QubesOS/qubes-issues#4160 some DVM Templates I created manually after installation had menu entries automatically created to launch a DVM (this is what confused me) and some did not.  That's why I chose the language I did, it's not clear (on my system at least) that there is consistent behavior with menu item behavior for manually created DVM Templates.

PR #709, opened 2018-09-22
Replace ccleaner with sdelete from Sysinternals.

100111001:
For nulling the free space currently ccleaner is recommended. As this software can contain adware and has to be installed first, I'd recommend SDelete from Sysinternals.

Advantage: It's an official tool from Microsoft and it's a command line utility.

Additionally, this guide should be updated anyway, because as of R4.0 qubes does not use root.img but lvm thin provisioning.

### DIFF
--- a/customization/windows-template-customization.md
+++ b/customization/windows-template-customization.md
@@ -133,19 +133,19 @@ Manual tasks that can/should be started in the template
     1. Run windows drive cleaner as Administrator.
     2. Enable all the task and run the cleaner
 
- * CCleaner file cleaning
-    1. Install CCleaner free
-    2. Copy the attached ccleaner configuration file in CCleaner program file folder
-    3. Run ccleaner with all option set except "wipe free space" (it will also remove user history and preferences)
-    4. Run ccleaner only with the option "wipe free space".
-
-        It will write zeros in all unused space. This will allow you to strip the root.img file later
+ * SDelete file cleaning (Not sure whether this is usefull for R4.0 because of lvm thin provisioning)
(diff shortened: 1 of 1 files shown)


customization/windows-template-customization.md
```diff
@@ -133,19 +133,19 @@ Manual tasks that can/should be started in the template
...
     2. Enable all the task and run the cleaner
 
- * CCleaner file cleaning
-    1. Install CCleaner free
-    2. Copy the attached ccleaner configuration file in CCleaner program file folder
-    3. Run ccleaner with all option set except "wipe free space" (it will also remove user history and preferences)
-    4. Run ccleaner only with the option "wipe free space".
-
-        It will write zeros in all unused space. This will allow you to strip the root.img file later
+ * SDelete file cleaning (Not sure whether this is usefull for R4.0 because of lvm thin provisioning)
```
marmarek: The whole instruction assumes file based disk volumes (see "copy root.img using the following command" below), so comment about LVM is irrelevant here.
So, better mark the whole section as 3.2 only (or 4.0 with file-based volumes).

As for 4.0 with LVM, ideally would be convincing Windows to use TRIM operation (if it doesn't happen already). Otherwise, procedure like this should be used, but the second part (the one in dom0) should be adjusted for LVM (not necessary in this PR).

tasket: I'm not sure which Windows versions enable TRIM be default. However, the doc section could be updated to merely point to Microsoft instructions for enabling it, in addition to using `fallocate --dig-holes` in dom0. This should be both easier and applicable to various 3.2 + 4.0 storage layers.

andrewdavidwong:

PR #713, opened 2018-10-10
Update multimedia.md

one7two99:
updated documentation to install spotify (link to recent pubkey) and some reformatting

### DIFF
--- a/configuration/multimedia.md
+++ b/configuration/multimedia.md
@@ -33,13 +33,10 @@ In the last step we will create an AppVM from this template.
 
 Launch a Terminal in the new template VM:
 
-`qvm-run --auto t-multimedia gnome-terminal`
+`qvm-run --auto --user root t-multimedia xterm`
 
 Important:
 Enter all the following commands in the terminal of the template VM
-Become the root user to run all following command without the need to use sudo in the multimedia template VM
-
-`sudo -i`
 
 This howto assumes that you have xclip available in the AppVM where you download the Repository Signing keys.
 xclip will be used to paste the content of the clipboard to a file.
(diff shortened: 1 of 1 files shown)


configuration/multimedia.md
```diff
@@ -57,58 +54,42 @@ Import GPG-Key for spotify
...
 
-This should look like:
-
-    [user@t-multimedia ~]$ `gpg --with-fingerprint spotify.pubkey`
-
-    pub  4096R/341D9410 2017-07-25 Spotify Public Repository Signing Key <tux@spotify.com>
-
-         Key fingerprint = 0DF7 31E4 5CE2 4F27 EEEB  1450 EFDC 8610 341D 9410
-
+Compare it for example to https://keyserver.ubuntu.com/pks/lookup?op=vindex&search=0xA87FF9DF48BF1C90&fingerprint=on
```
marmarek: Instead of verification with another mean of downloading the key based on its ID, better verify if that key really belong to spotify. For example here: https://www.spotify.com/de/download/linux/

one7two99: I agree, I'll add this to the howto, as I have also planned to rewrite it, to make it easier to follow.

marmarek: The long one 932..C90 is a fingerprint. Key ID (in gpg at least) is simply last 8/16 chars of the fingerprint.
For key verification, you need something that externally confirm given key really belong to spotify. Keyserver on its own doesn't give it to you - it only decode information contained in the key and anyone can generate key with the same display name, but not the fingerprint.
Anyway, spotify website pointed above do contain key fingerprint, so have extra info: someone controlling https://www.spotify.com certify that this key really belongs to spotify.

andrewdavidwong:

PR #716, opened 2018-10-15
Update hvm.md

MadsRC:
It's fairly easy to miss the fact that some of the text is for R3 and some of it is for R4. In the source it's pretty straightforward (Since it contains a linebreak) but when rendered, said linebreak is removed. Adding a few third-level headers clears this up, at least in opinion.

### DIFF
--- a/managing-os/hvm.md
+++ b/managing-os/hvm.md
@@ -22,7 +22,8 @@ Interested readers might want to check [this article](https://blog.invisiblethin
 Creating an HVM domain
 ----------------------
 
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
This is definitely better, so I'm merging. However, it would be even better if this page followed our [Version-specific Documentation](https://www.qubes-os.org/doc/doc-guidelines/#version-specific-documentation) guidelines. Please feel free to open a new PR if you're willing to do that too.

PR #717, opened 2018-10-16
Document Split GPG error

haplo:
Split GPG doesn't work when the private key is protected by a passphrase in the GPG backend. It took me a couple hours to find the problem and how to fix it, first because the ioctl error is not common and then because of the pinentry-ncurses limitation. I'm hoping that future users that search for "qubes gpg ioctl error" will find this information.

### DIFF
--- a/security/split-gpg.md
+++ b/security/split-gpg.md
@@ -74,6 +74,14 @@ signed before the operation gets approved. Perhaps the GPG backend domain
 could start a Disposable VM and have the to-be-signed document displayed
 there? To Be Determined.
 
+- The Split GPG client will fail to sign or encrypt if the private key in the
+GnuPG backend is protected by a passphrase, it will give a *"Inappropriate ioctl
+for device"* error. Avoid setting passphrases for the private keys in the GPG
+backend domain, it won't provide extra security anyway, if an attacker gains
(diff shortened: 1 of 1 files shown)


haplo:
I thought I had read the whole document in search for a solution to the problem I was having, but I missed that bit, because it's in the QubesOS 3.2 section and I'm running 4.0.

I propose removing that paragraph from the 3.2 section and consolidating into the one I'm adding, given that it affects all QubesOS versions. I can reference the other sections that further explain why adding a passphrase doesn't add to the security of the keys.

haplo:
@andrewdavidwong I cleaned up the mentions to passphrases not working. Please see
my commit message, as I mentioned it's not specific to any QubesOS version, so I don't
think it's worth it to mention it twice in those sections.

Also I think I can do a cleanup of the whole document, would that be worthy of a separate PR?

andrewdavidwong:
> I cleaned up the mentions to passphrases not working. Please see my commit message, as I mentioned it's not specific to any QubesOS version, so I don't think it's worth it to mention it twice in those sections.

Ok, that's fine.

> Also I think I can do a cleanup of the whole document, would that be worthy of a separate PR?

That would be great. Thanks!

PR #724, opened 2018-10-24
Fix misformating in Split GPG doc

haplo:
An error was introduced in pull request #717.

qubes doc split gpg misformat


andrewdavidwong:
This is because the `<` appears at the beginning of a line. As long as it's not the first character in a line, it'll be fine. For example, this would be another way to fix it:

```
you have a private key that already has a passphrase set use `gpg2
--edit-key <key_id>`, then `passwd` to set an empty passphrase. Be aware that
```

haplo:
Thank you @andrewdavidwong , I wasn't sure so I went with the safe option.

andrewdavidwong:
You should always check your changes by [serving the site on your local machine](https://github.com/QubesOS/qubesos.github.io#instructions) before submitting.

haplo:
Thank you Andrew, I will run and check it next time for sure.

PR #744, opened 2018-11-19
Update i3 doc, add config link, minor format fixes

anadahz:


customization/i3.md
```diff
@@ -11,18 +11,25 @@ redirect_from:
 
 # i3 installation in dom0
 
-i3 is part of the testing repository (as of Qubes R3.1) and can be installed from there
-using the dom0 update mechanism.
+i3 is part of the stable repository (as of Qubes R3.1) and can be installed by
```
andrewdavidwong: Is it really true that this was in the stable repo at any point while R3.1 was supported?

anadahz: It is true, you can find the i3 rpm on the stable (current) repository of R3.1 here:
https://yum.qubes-os.org/r3.1/current/dom0/fc20/rpm/i3-4.8-3.fc20.x86_64.rpm

PR #746, opened 2018-11-23
Added documentation for qvm-device and derived commands

GammaSQ:
After raising the [issue of missing documentation](https://github.com/QubesOS/qubes-issues/issues/4530#issuecomment-441106070) for qvm-device, qvm-pci, qvm-usb and qvm-block, I decided to write down what the manpage told me plus some information from the mailinglist (mainly the losetup-trick documented in qvm-block).

However, there is a lot I don't know which would need someone more knowledgeable than myself to fill in:
 - What are the possible device-classes? I know of at least `mic` which isn't documented in the manpage. (in qvm-device.md)
 - What's the idea behind the `--all` option? Isn't this default-behavior? Why is it required when using `--exclude`? (in qvm-device.md)
 - How does `persistent` behave? Especially, when does Qubes try to attach a device again? Only when both VMs and device are available? When both VMs are available? What do the fail-states look like? (in qvm-device.md)
 - What's the difference between `no-strict-reset` and `permissive`? (in qvm-pci.md)
 - I'm unsure as to what's the threat-model / security considerations behind using qvm-block over qvm-usb over qvm-pci. Using qvm-usb compared to qvm-pci is convenient, but is there a disadvantage of using qvm-usb over qvm-block?
 - I created the files by copying existing 3.2 documentation. I'm not sure weather the "redirect_from" headers are version-specific, so maybe they need to be removed.

This does NOT close the mentioned issue, since the qvm-device interface still should be reworked to be more intuitive.

P.S.: Someone should probably check spelling.

### DIFF
--- a/reference/tools/4.0/dom0.md
+++ b/reference/tools/4.0/dom0.md
@@ -12,3 +12,8 @@ Reference pages for these tools are available here:
  * <https://dev.qubes-os.org/projects/core-admin/en/latest/manpages/>
  * <https://dev.qubes-os.org/projects/core-admin-client/en/latest/manpages/>
 
+ * [qvm-device](/doc/tools/4.0/dom0/qvm-device/)
+ * [qvm-pci](/doc/tools/4.0/dom0/qvm-pci/)
+ * [qvm-usb](/doc/tools/4.0/dom0/qvm-usb/)
+ * [qvm-block](/doc/tools/4.0/dom0/qvm-block/)
+
--- a/reference/tools/4.0/dom0/qvm-block.md
+++ b/reference/tools/4.0/dom0/qvm-block.md
@@ -1,0 +1,74 @@
+---
+layout: doc
+title: qvm-block
+permalink: /doc/tools/4.0/dom0/qvm-block/
+redirect_from:
+- /doc/dom0-tools/qvm-block/
+- /doc/dom0-tools/qvm-block/
+- /en/doc/dom0-tools/qvm-block/
+- /doc/Dom0Tools/QvmBlock/
+- /wiki/Dom0Tools/QvmBlock/
+---
+
+#qvm-block
+
+##Synopsis
+
+    qvm-block [COMMAND] [OPTIONS] <target-vm-name> <block-vm-name>:<device>
+
+`qvm-block` is an alias for `qvm-device block`. [`qvm-device`](/doc/tools/4.0/dom0/qvm-device/) offers a generalised interface for attachment and detachment of devices (in the case of `qvm-block` block devices), only the options specific to `qvm-block` will be discussed here.
+
+##Usage
+
+Please follow the [`qvm-device`](/doc/tools/4.0/dom0/qvm-device/) documentation for general usage.
+
+See also [handling USB-devices](/doc/usb/#r40) for a quick guide of the most common use-case.
+
+##Options Specific to Block-Devices
+
+These options can be specified using the `-o` or `--option` handle. The general syntax is
+
+    qvm-block a -o <option=value>[ -o <option=value>[...]] <target-vm> <source-vm>:<device-id>
+
+###frontend-dev
+Specify device node in target domain.
+
+Possible values are non-space strings.
+
+Default is first available, starting from `xvdi`.
+
+###read-only
+Attach block-device in read-only mode. If device itself is read-only, only read-only attach is allowed.
+
+Possible values are `True` and `False`.
+
+Default is `False`, if read-write attachment is possible.
+
+###devtype
+Type of device
+
+Possible Values are `disk` adn `cdrom`.
+
+Default: `disk`
+
+##Deprecated --attach-file option
+Up until Qubes 3.2, `qvm-block` exposed an option to attach a [file as block-device](/doc/tools/3.2/dom0/qvm-block/) to another VM. Qubes 4.0 does _not_ support this behavior. As a workaround, `losetup` can be used in linux VMs.
+
+In the linux-source VM run
+
+    sudo losetup /dev/loop0 /path/to/file
+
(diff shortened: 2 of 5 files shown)


andrewdavidwong:
> P.S.: Someone should probably check spelling.

This is (largely) covered by Travis CI. Please take a look at the log for any new unknown spellings introduced.

marmarek:
>     * What are the possible device-classes? I know of at least `mic` which isn't documented in the manpage. (in qvm-device.md)

No programmatic way to list them yet. Right now supported are: `block`, `pci`, `usb`, `mic`.

>     * What's the idea behind the `--all` option? Isn't this default-behavior? Why is it required when using `--exclude`? (in qvm-device.md)

There is a little difference between bare `list` command and the one with explicit qubes list (including `--all`) - in the latter case, devices attached to those qubes are listed too _even if not currently present in the system_ (in that case description will be "Unknown device").
To be honest, I think this difference is unintentional (in other words: a bug).

>     * How does `persistent` behave? Especially, when does Qubes try to attach a device again? Only when both VMs and device are available? When both VMs are available? What do the fail-states look like? (in qvm-device.md)

It basically means "attach device at qube start". If device is not there for any reason (not plugged in, backend unavailable etc), qube startup will fail.

> What's the difference between `no-strict-reset` and `permissive`? (in qvm-pci.md)

`no-strict-reset` means do not require the device to support proper PCI reset (already described in doc)
`permissive` is about PCI config space access - by default only selected whitelisted registers are writable, with `permissive` option set, most registers are writable (besides few blacklisted). Details: https://www.qubes-os.org/doc/assigning-devices/#pci-passthrough-issues

>     * I'm unsure as to what's the threat-model / security considerations behind using qvm-block over qvm-usb over qvm-pci. Using qvm-usb compared to qvm-pci is convenient, but is there a disadvantage of using qvm-usb over qvm-block?

See https://www.qubes-os.org/doc/usb/#attaching-a-single-usb-device-to-a-qube-usb-passthrough
In short, preferred order (use first that suit the particular use): qvm-block, then qvm-usb, then qvm-pci.
The smaller part of device you connect, the smaller the attack surface is.

The source of actual qvm-device manual page is here: https://github.com/QubesOS/qubes-core-admin-client/blob/master/doc/manpages/qvm-device.rst

GammaSQ:
I've added explanations, link to the invisiblethingslab-blogpost on usb v. block and linked the commend referencing `--all` behavior might be erroneous.

I'm still wondering why `--exclude` requires `--all` and when devices stay available in their source-vm. I.e. can I attach a block-device somewhere, yet still use it in the VM it originated in? I know I can't use PCI-devices in two VMs.
Maybe even what should happen when I try to attach an already attached device to another VM? I believe pci, usb and block behave differently? What about mic?

andrewdavidwong:
@GammaSQ, I just noticed that you're trying to change these man pages in the wrong repo. There's a note about this at the bottom of every page (on the web version):

> Note: The Markdown source of this page in qubes-doc was generated by running the update-manpages-4-0 script on qubes-core-agent-linux/doc/vm-tools/. If you wish to update the contents of this page as it appears on the Qubes OS website, please submit a pull request to change the appropriate file in qubes-core-agent-linux/doc/vm-tools/. Do not attempt to change the Markdown source of this page in qubes-doc directly. All direct changes to the Markdown file will be overwritten the next time this page is regenerated.

Please submit your changes to the appropriate repo(s).

Also, use links very sparingly, since these are man pages that will usually be viewed in dom0 (without network access). The relative links you're trying to use will not make sense there.

GammaSQ:
I'm very sorry, but I believe you are wrong. The above Note is valid for Q3.2 dom0 command reference, the documentation I've written is for Q4 and didn't exist before. (Since the Q4-reference is hosted on dev.qubes-os.org while 3.2 is on www.qubes-os.org) Therefore, I don't think this would be overwritten by autogenerated manpages.

If what you are saying is it's better to host manpage - copies in this part of the documentation, I'd ask weather this would fit somewhere else in the documentation, since I specifically intended it as documentation and not as manpage. As it turned out, having [separate manpages seems to be difficult](https://github.com/QubesOS/qubes-issues/issues/4530#issuecomment-445353118).

andrewdavidwong:
> I'm very sorry, but I believe you are wrong.

This is one of those rare cases in which I'm right. :wink: 

> The above Note is valid for Q3.2 dom0 command reference, the documentation I've written is for Q4 and didn't exist before.

The note is valid for all man pages in qubes-doc generated from another source, regardless of Qubes version. (Even though these man pages are not generated from another source, they _should_ be. See below.)

> (Since the Q4-reference is hosted on dev.qubes-os.org while 3.2 is on www.qubes-os.org)

Sure, but yours would be hosted on www.qubes-os.org instead of dev.qubes-os.org.

> Therefore, I don't think this would be overwritten by autogenerated manpages.

I'm the one who overwrites it with autogenerated man pages. I'm telling you that I don't want to overwrite or delete your stuff, but I will if it's in the wrong place. :slightly_smiling_face: 

> If what you are saying is it's better to host manpage - copies in this part of the documentation,

I'm saying that it would be better _not_ to host the original source files in `qubes-doc`. The general policy is for the original source files of man pages to be in the repos of the tools for which they are man pages, so that the tools and man pages are kept together. The Markdown versions in `qubes-doc` are generated from those.

> I'd ask weather this would fit somewhere else in the documentation, since I specifically intended it as documentation and not as manpage.

In that case, yes, these should be elsewhere in the documentation and not presented as man pages (in which case, it's appropriate for the Markdown source to be in `qubes-doc`). However, you should drop all of the stuff that is (or should be) covered by the man pages for these tools. Find the appropriate existing document(s) to add the other information you want to add. Please submit a new PR for this.

PR #751, opened 2018-12-04
split-gpg + enigmail: qubes-gpg-client-wrapper is always needed

ThomasWaldmann:
the section applies to qubes 3.2 (did not check, but it said so)
and to qubes 4.0 (verified, was not obvious before).


ThomasWaldmann:
that was quick! thanks.

andrewdavidwong:
Thank *you*.

PR #752, opened 2018-12-04
Windows 10 Qube for Dummies

ThomasWaldmann:
The motivation to write this down was not that it is very
complicated (obviously it is not), but rather that the docs
I have read (before just trying it) somehow left a bad
impression about Windows on Qubes.

I had the impression that:

- if one wants Windows, one must use Windows 7 (which is
  EOL 1/2020, so only a short term option, if any)
- newer Windows don't work or are unstable because the
  tools needed are not tested / not working on them.

Both is not true IF one can live with a live with a very
basic featureset, which I also described.

### DIFF
--- a/managing-os/windows/windows-vm.md
+++ b/managing-os/windows/windows-vm.md
@@ -8,6 +8,64 @@ permalink: /doc/windows-vm/
 Installing a Windows VM
 =======================
 
+Simple Windows 10 install (Qubes 4.0)
+-------------------------------------
+
+If you just want something simple and you can live without some features.
+
+Works:
+- display (1440x900 or 1280x1024 are a nice fit onto FHD hw display)
+- keyboard (incl. correct mapping), pointing device
+- network (emulated Realtek NIC)
+
+Does not work:
+- copy & paste (the qubes way)
+- copying files into / out of the VM (the qubes way)
+- assigning USB devices (the qubes way via the tray applet)
+- audio output and input
+- PCI device 5853:0001 (Xen platform device) - no driver
+- all other features/hardware needing special tool/driver support
+
+Installation procedure:
+- Have the Windows 10 ISO image (I used the 64-bit version) downloaded in some qube.
+- Create a new Qube:
+  - Name: Win10, Color: red
+  - Standalone Qube not based on a template
+  - Networking: sys-firewall (default)
+  - Launch settings after creation: check
+  - Click "OK".
+- Settings:
+  - Basic:
+    - System storage: 30000+ MB
+  - Advanced:
+    - Include in memory balancing: uncheck
+    - Initial memory: 4096+ MB
+    - Kernel: None
+    - Mode: HVM
+  - Click "Apply".
+  - Click "Boot from CDROM":
+    - "from file in qube":
+      - Select the qube that has the ISO.
+      - Select ISO by clicking "...".
+    - Click "OK" to boot into the windows installer.
+- Windows Installer:
+  - Mostly as usual, but automatic reboots will halt the qube - just restart
+   it again and again until the installation is finished.
+  - Install on first disk.
+  - Windows license may be read from flash via root in dom0:
+
+    `strings < /sys/firmware/acpi/tables/MSDM`
+
+    Alternatively, you can also try a Windows 7 license key (as of 2018/11
+    they are still accepted for a free upgrade).
+    
+    I first installed Windows and all updates, then entered the license key.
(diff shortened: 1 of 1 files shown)


managing-os/windows/windows-vm.md
```diff
@@ -8,6 +8,61 @@ permalink: /doc/windows-vm/
...
+    - Mode: HVM
+  - click "Apply"
+  - click "Boot from CDROM"
+    - "from file in qube":
+      - select the qube that has the ISO
+      - select ISO by clicking "..."
+    - click "OK" to boot into the windows installer
+- Windows Installer:
+  - Mostly as usual, but automatic reboots will halt the qube - just restart
+   it again and again until the installation is finished.
```
ThomasWaldmann: is there a better way?

managing-os/windows/windows-vm.md
```diff
@@ -8,6 +8,61 @@ permalink: /doc/windows-vm/
...
+  - Install on first disk.
```
ThomasWaldmann: I couldn't get rid of 2nd and 3rd disk via the GUI.

managing-os/windows/windows-vm.md
taradiddles: IMO to avoid duplication with the [HVM page](https://www.qubes-os.org/doc/hvm/) it would make more sense to write that Windows 10 works as a HVM (like most other OSes) and link to the HVM page. You could then list stuff that is specific to Windows 10 (unreliable QWT so no qubes integration, ...).
My 2c :)

ThomasWaldmann: my question referred to the halting, see line 52.

taradiddles: Yeah I was replying to the whole PR (FWIW I wrote most of the windows doc). Overlooked the "review changes" button.

Re- your question: I don't know of any other way - once a VM shuts down (even after a reboot "event") it has to be restarted manually. Alternatively a loop and a few commands in bash can autorestart a VM after it stops running. But having to manually restarting the VM allows people to `qvm-clone` every now and then.

marmarek: Right now there is no way to avoid this manual restart. The code to automatically restart qubes-related services (gui-daemon, qrexec-daemon etc) at such restart is missing.

managing-os/windows/windows-vm.md
marmarek: There is none. Second disk is "private" volume - especially useful for TemplateVM, the third one is "volatile" volume - fresh at each qube start - could be used for swap file, but not sure if Windows support having it separately.

managing-os/windows/windows-vm.md
```diff
@@ -8,6 +8,61 @@ permalink: /doc/windows-vm/
...
+  - name: Win10, colour: red
+  - standalone Qube not based on a template
+  - networking: sys-firewall (default)
+  - launch settings after creation
+  - click "OK"
+- Settings:
+  - Basic:
+    - system storage: 30000MB+
+  - Advanced:
+    - include in memory balancing: uncheck
```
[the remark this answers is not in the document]
marmarek: correct

ThomasWaldmann:
@marmarek i thought the section i added belongs just there:

```
Installing a Windows VM
====================
 
Simple Windows 10 install (Qubes 4.0)
```

The main headline is "Installing a Windows VM", so describing how to do that, belongs just right there and stuff like "how to move a qubes windows vm from 3.2 to 4.0" belongs after this.

So, maybe rather the other stuff needs moving?

About other windows versions: I did not try them and maybe they are not that important:
- Windows 7 will run out of support 1/2020 anyway and has frequent issues with the updater stopping to work.
- Windows 8/8.1 isn't much loved due to its strange UI.
- Everybody who has a 7/8.x key can currently upgrade to 10 for free just by doing a win10 install and giving that key.

ThomasWaldmann:
OK, so how are we proceeding with this?

A guy on IRC wondering about windows and qubes reminded me to look at my PR.

andrewdavidwong:
I was about to close this due to staleness, but then I saw that everyone seems enthusiastic about merging this content. The questions were just over relatively minor details like the optimal location of the content. We shouldn't allow the perfect to be the enemy of the good, so I've merged this PR with a couple of quick minor alterations per @marmarek's comments. We can easily move it around and further improve it later.

ThomasWaldmann:
Cool, thanks! :)

PR #754, opened 2018-12-05
Add separators between mailing list sections

tokideveloper:
For better visual distinction, add separators between the
mailing list sections. The visual differences between the
many headings and subheadings on the page are very small
such that overlooking the beginning of a new mailing list
section becomes likely. As a result, people could send their
emails to a wrong list by accident which already happened.

### DIFF
--- a/about/support.md
+++ b/about/support.md
@@ -223,6 +223,7 @@ guidelines.
    Project does not maintain a forum. The curious can find these by searching
    the list archives.)
 
+---
 
 qubes-announce
 --------------
@@ -244,6 +245,7 @@ here.
  * This list can also be browsed via an optional [Google Groups web
    interface][qubes-announce-web].
 
+---
 
 qubes-users
 -----------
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
> For better visual distinction, add separators between the
mailing list sections. The visual differences between the
many headings and subheadings on the page are very small
such that overlooking the beginning of a new mailing list
section becomes likely. 

This is better addressed by improving the way information is presented on the page as a whole, which I'll do in a separate commit.

> As a result, people could send their emails to a wrong list by accident which already happened.

It's not clear that this was the cause. No matter how clearly we present things, some people will always get confused. :slightly_smiling_face:

tokideveloper:
[**@andrewdavidwong**](https://github.com/andrewdavidwong) commented on [6. Dez. 2018, 04:59 MEZ](https://github.com/QubesOS/qubes-doc/pull/754#issuecomment-444740598 "2018-12-06T03:59:36Z - Replied by Github Reply Comments"):
> This is better addressed by improving the way information is presented on the page as a whole, which I'll do in a separate commit.
> 

Thanks, Andrew! The page looks a lot better now! :-)

> > As a result, people could send their emails to a wrong list by accident which already happened.
> 
> It's not clear that this was the cause. No matter how clearly we present things, some people will always get confused.

Including me … sometimes. ;-)

PR #756, opened 2018-12-11
Note that minimal fedora NetVM needs polkit

jpouellet:
Otherwise nm-applet is not able to control NetworkManager over dbus, and you get the following error message:

```
(nm-applet:19808): nm-applet-WARNING **: 17:24:56.716: disconnect_cb: Device disconnect failed: (1) org.freedesktop.NetworkManager.network-control request failed: not authorized
```

This is unfortunate, but I don't see a better way around it.

See also: https://groups.google.com/d/topic/qubes-users/b_yEFCMazuM/discussion


jpouellet:
Also, it's not clear to me that `gnome-keyring` is actually required (my sys-net seems to work fine without it), but I've left it in since I don't know why it was there, and as @andrewdavidwong has pointed out before: https://en.wikipedia.org/wiki/Wikipedia:Chesterton%27s_fence

PR #761, opened 2019-01-02
Update Firewall.md

unman:
Make it clear that limit on number of rules only applies prior to R4.0


andrewdavidwong:
Looks like this is related to https://github.com/QubesOS/qubes-issues/issues/1570.

PR #762, opened 2019-01-05
Create howto-use-expressvpn-with-qubes.md

one7two99:
This howto covers the complete process how to use ExpressVPN with Qubes by creating a new template, then an AppVM and finally the ExpressVPN configuration.
ExpressVPN allows you to run your traffic through a VPN and protect it from spying eyes, like your ISP.

### DIFF
--- a/configuration/howto-use-expressvpn-with-qubes.md
+++ b/configuration/howto-use-expressvpn-with-qubes.md
@@ -76,7 +76,7 @@ Configure ExpressVPN
 - Import  the downloaded OpenVPN configuration file (QubesIncoming/...)
 - Beautify the Name, like ExpressVPN-<Locationname>
 - Copy & Paste the Username and Password from your ExpressVPN page into correct fields
-  (User key password can be left empty)
+  (User key password should also contain the password)
 - Important: Click on the right Icon in the password field and choose:
 - (X) Store the passwords for all users
 - Save the dialog
--- a/configuration/howto-use-expressvpn-with-qubes.md
+++ b/configuration/howto-use-expressvpn-with-qubes.md
@@ -1,0 +1,145 @@
+This Howto describes how you can use ExpressVPN with Qubes to run all traffic through a VPN Proxy-VM.
+More info about ExpressVPN can be found here:
+https://www.expressvpn.com/what-is-vpn/browse-anonymously
+
+This howto will cover:
+- Creating a new vpn-template which is based on fedora-28-minimal (named: t-fedora-28-vpn)
+- Create an Proxy-AppVM (named: sys-vpn)
+- Configure ExpressVPN in the Proxy-AppVM
+
+If you run into any problems with this Howto, do not hesitate to contact me.
+The benefit of this setup is:
+you can decide which AppVMs should run their traffic through ExpressVPN,
+simple by setting their netvm to sys-vpn.
+
+Install fedora-28-minimal template in dom0
+```
+sudo qubes-dom0-update qubes-template-fedora-28-minimal
+```
+
+Clone this template to a new template, which will be used for the VPN AppVM later
+Hint: keep the original templates as baseline templates if you run into problems
+```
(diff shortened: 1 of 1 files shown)


tasket:
It's not up to me, but instructions for specific vpn providers (there are hundreds) seems more appropriate for a blog or mailing list.

Also, given the provider names, this tends to be confusing for non-expert users (Express vpn setup - 'This is the faster method?' or Private Internet Access - 'I choose these instructions for privacy?').

marmarek:
I don't think the official Qubes OS documentation should recommend specific VPN provider(s), unless we have explicit reasons for it (like believing it provides better privacy than others).

On the other hand, this instruction looks like pretty generic OpenVPN setup. The only provider specific things are URL to download configuration, suggested names and DNS leak tests URL at the end. Other than that, it should work with other providers supporting OpenVPN like NordVPN, Mullvad etc.
And we already have generic documentation here: https://www.qubes-os.org/doc/vpn/#set-up-a-proxyvm-as-a-vpn-gateway-using-networkmanager, which also have screenshots for most steps. It could be appropriate to add those few provider specific items there ("how to get openvpn configuration for: ExpressVPN: ..., some-other-vpn: ..."). But the whole page about specific provider belongs rather to https://github.com/Qubes-Community/ or some blog.

andrewdavidwong:
I agree with @marmarek and @tasket. Absolutely no advertisements in Qubes documentation.

We also shouldn't have more than one general VPN document, so any new general information worth adding should be added to https://github.com/QubesOS/qubes-doc/blob/master/configuration/vpn.md.

one7two99:
Hello,

Am Sa., 5. Jan. 2019, 04:10 hat Andrew David Wong <notifications@github.com>
geschrieben:

> W also shouldn't have more than one general VPN document, so any new
> general information worth adding should be added to
> https://github.com/QubesOS/qubes-doc/blob/master/configuration/vpn.md.
>
But this howto is not that detailed like mine and is covering Qubes 3.2
from the screenshots.

Also it is not necessary to put the password in a file as it is mentioned
in the howto.

I agree that there should be no advertisement but it would be good to have
one specific page about those kind of VPN services.
The reason:
People who are not looking for a VPN setup might read that using a VPN
provider will improve their setup and therefore newbies should think about
adding this layer to their qubes setup.

I also included building a whole template from a Fedora minimal template
which is missing in the VPN docs.
I am afraid that putting this info into the Wuebs Community Docs will
likely not be read by beginners.

Maybe we can add information for some more providers like NordVPN and
others.
We can make clear that this is not an advertisement.

- O

andrewdavidwong:
> But this howto is not that detailed like mine and is covering Qubes 3.2
> from the screenshots.
> 
> Also it is not necessary to put the password in a file as it is mentioned
> in the howto.
> 

These are good reasons to improve the page, not reasons to create a separate page that leaves the existing one deficient.

> I agree that there should be no advertisement but it would be good to have
one specific page about those kind of VPN services.
The reason:
People who are not looking for a VPN setup might read that using a VPN
provider will improve their setup and therefore newbies should think about
adding this layer to their qubes setup.

This is a reason to add a new section to the existing page.

> I am afraid that putting this info into the Wuebs Community Docs will
likely not be read by beginners.

The folks at https://github.com/Qubes-Community are in a better position to judge whether this concern is justified or not.

> Maybe we can add information for some more providers like NordVPN and
others.
We can make clear that this is not an advertisement.

Making the instructions more generic without mentioning specific VPN providers is probably the way to go, as @marmarek described above, and this would make it even more appropriate for the content to go into the existing VPN document.

one7two99:
Hi Andrew,

Thanks for the answer, I have already added the info the Community docs.

>> The folks at https://github.com/Qubes-Community are in a
>> better position to judge whether this concern is justified or not

I know this projects as I was also involved in it from the beginning and I
have already transferred some info from my personal repo to the Qubes
community repo  ;-)

- O.

PR #767, opened 2019-01-21
Update disposablevm-customization.md

goldiqubes:
After creating a disp-sys-usb VM. I had problems with my mouse not working. After many hours I discovered the fix and to help others documented the fix here.


marmarek:
Having https://github.com/QubesOS/qubes-issues/issues/3904#issuecomment-389718164 would make it easier to find...

PR #768, opened 2019-01-22
Added info on Intel TXT requirement

tomschlenkhoff:
Hope this helps others not to waste time if the initial requirements are not met.


marmarek:
For Coreboot compatible machines, there is an alternative: [Heads](https://github.com/osresearch/heads). Functionally similar thing, but built into coreboot+linux based firmware.

PR #787, opened 2019-02-10
Add solution to make persistent rules in /etc/hosts

jamke:
Add solution to make persistent rules in /etc/hosts. 
I checked that it works, but I hope the approach will be checked before approval for being qubes-way and clean enough.
If it's not, different solution should be provided to modify /etc/hosts permanently for TemplateBasedVMs, because, it seems, there is no one in the documentation.


andrewdavidwong:
@jamke: FYI, it looks like your other PR being merged (#786) has caused a minor conflict in this PR that you'll want to resolve.

jamke:
@andrewdavidwong I see, now it looks to be solved. 
Also removed unnecessary new line several lines above.

PR #790, opened 2019-02-15
Replace old HTTPS repo warning with APT vuln warning

andrewdavidwong:
My understanding is that the old HTTPS repo warning is no longer applicable to users who are installing Qubes 4.0.1. Meanwhile, @v6ak has [suggested](https://groups.google.com/forum/#!msg/qubes-users/5D8AxG3jtdw/7K1bm2GIAgAJ) adding a warning about QSB 46.

### DIFF
--- a/installing/installation-guide.md
+++ b/installing/installation-guide.md
@@ -31,8 +31,16 @@ Furthermore, if you are installing Qubes on a potentially compromised system, we
 Qubes 4.0.1 Warning
 -------------------
 
-After installing Qubes 4.0.1, please install fresh APT-based (e.g., Debian and Whonix) TemplateVMs in order to be protected from the APT update mechanism vulnerability that was patched after the release of 4.0.1.
-For instructions and further details, please see [QSB #46].
+Immediately after installing Qubes 4.0.1, please upgrade all of your APT-based (e.g., Debian and Whonix) TemplateVMs by executing the following command in a dom0 terminal for each such TemplateVM:
+
+    $ sudo qubes-dom0-update --action=upgrade <template-package-name>
+
+For example, the command for the `debian-9` TemplateVM would be:
+
+    $ sudo qubes-dom0-update --action=upgrade qubes-template-debian-9
+
+These upgrades are required in order to be protected from the APT update mechanism vulnerability that was announced and patched in [QSB #46], which was after the release of Qubes 4.0.1.
+This method is simpler than the method recommended in [QSB #46], but it is just as safe and effective so long as it is performed immediately after installing Qubes OS.
 
 
 Hardware Requirements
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
In the [discussion thread](https://groups.google.com/d/topic/qubes-users/5D8AxG3jtdw/discussion), it was suggested that users use the GUI tools to update the templates installed from the ISO. However, I opted to instead advise users to install fresh templates, since this is what we advised as the safer route in the QSB. A user who reads a recommendation in the installation guide to use the GUI tool, then reads the QSB, will be under the impression that the installation guide is being sloppy with respect to security.

installing/installation-guide.md
```diff
@@ -28,37 +28,11 @@ Therefore, we *strongly* recommended that, prior to starting the Qubes installer
...
+Qubes 4.0.1 Warning
+-------------------
 
-4. This opens the gedit text editor in a window. Change all four instances of `http` to `https`.
-5. Click the "Save" button in the top-right corner of the window.
-6. Close the window.
-7. Check for updates normally.
-8. Shut down the TemplateVM.
+After installing Qubes 4.0.1, please install fresh APT-based (e.g., Debian and Whonix) TemplateVMs in order to be protected from the APT update mechanism vulnerability that was patched after the release of 4.0.1.
+For instructions and further details, please see [QSB #46].
```
marmarek: Since just after installation templates are definitely installed using rpm (not for example upgraded from previous template version or cloned), there is an easier option to reinstall them:
```
sudo qubes-dom0-update --action=upgrade qubes-template-debian-9
```
(and similar for other templates)

Do you think it worth noting it here? The full procedure described in the QSB of course works too, but is more complex (to handle the cases where template wasn't installed directly with rpm).

andrewdavidwong: Sure, let's make it easier on users.

andrewdavidwong: Let me know what you think of the updated recommendation, @marmarek.

installing/installation-guide.md
```diff
@@ -31,8 +31,16 @@ Furthermore, if you are installing Qubes on a potentially compromised system, we
 Qubes 4.0.1 Warning
 -------------------
 
-After installing Qubes 4.0.1, please install fresh APT-based (e.g., Debian and Whonix) TemplateVMs in order to be protected from the APT update mechanism vulnerability that was patched after the release of 4.0.1.
-For instructions and further details, please see [QSB #46].
+Immediately after installing Qubes 4.0.1, please upgrade all of your APT-based (e.g., Debian and Whonix) TemplateVMs by executing the following command in a dom0 terminal for each such TemplateVM:
+
+    $ sudo qubes-dom0-update --action=upgrade <template-package-name>
```
marmarek: Should it also mention relevant --enablerepo option for Whonix templates?

andrewdavidwong: How about how?

PR #791, opened 2019-02-17
Update instructions in vm-sudo.md (su.qubes)

starius:
UPD. I was wrong in my original version. Commenting out that line makes any action requiring approval (e.g. terminal open). It looks this line was moved to another file (su.qubes).

### DIFF
--- a/security/vm-sudo.md
+++ b/security/vm-sudo.md
@@ -142,7 +142,7 @@ this for extra security.**
           [root@debian-8]# rm /etc/polkit-1/rules.d/00-qubes-allow-all.rules
           [root@debian-8]# rm /etc/polkit-1/localauthority/50-local.d/qubes-allow-all.pkla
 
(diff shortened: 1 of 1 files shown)


security/vm-sudo.md
```diff
@@ -142,7 +142,7 @@ this for extra security.**
           [root@debian-8]# rm /etc/polkit-1/rules.d/00-qubes-allow-all.rules
           [root@debian-8]# rm /etc/polkit-1/localauthority/50-local.d/qubes-allow-all.pkla
 
-    - In /etc/pam.d/su, comment out this line near the bottom of the file:
+    - In /etc/pam.d/su.qubes, comment out this line near the bottom of the file:
```
marmarek: `/etc/pam.d/su` is a symlink to `su.qubes`, so technically editing either of them should work (as long as the editor do not break the symlink).
But since some editors may break the symlink, it may be a good idea to change it here.

PR #794, opened 2019-02-19
Add troubleshooting step to fedora-29-minimal.md

rowland007:
Adds steps on how to add `e2fsprogs` to the fedora-29-minimal template if a user is unable to interact with it.

Instructions pulled from issue #[4671](https://github.com/QubesOS/qubes-issues/issues/4671) that was described by @marmarek

### DIFF
--- a/managing-os/templates/fedora-minimal.md
+++ b/managing-os/templates/fedora-minimal.md
@@ -37,6 +37,18 @@ It is highly recommended to clone the original template, and make any changes in
 
 You must start the template in order to customize it.
 
+Troubleshooting fedora-29-minimal
+---------------------------------
+
+If you're unable to interact with the fedora-29-minimal template, you may need to install `e2fsprogs`
+
+Without `e2fsprogs` installed, /rw cannot be prepared and cause applications to fail
+
+Quick workaround in a dom0 terminal:
+~~~
+qvm-run -u root --no-gui -p fedora-29-minimal 'dnf install -y e2fsprogs'
+~~~
+
 Customization
 -------------
 


marmarek:
I'm not sure if we should include specific, already fixed issues in template documentation.
The fix is already included in newer template version: https://github.com/QubesOS/updates-status/issues/848
Just moved it to stable.

rowland007:
The reason I added it is because I installed the `fedora-29-minimal` template on 4.0 out of the stable repo last month when it was [announced](https://www.qubes-os.org/news/2019/01/07/fedora-29-template-available/).  When it didn't work, I uninstalled it and then reinstalled it just to see if something failed during installation. After the reinstall it still did not work. The template has just been sitting on my machine since.

Yesterday, I saw the announcement on my Reddit feed (even though it was the same one from last month) I saw someone describing the same [problem](https://www.reddit.com/r/Qubes/comments/adq9ox/fedora_29_templatevm_available_for_qubes_40/edrrvz2) that linked me to the issue #[4671](https://github.com/QubesOS/qubes-issues/issues/4671) and I did see that the issue was talking about the testing repo but I figured I give it a shot.  It worked and now my `fedora-29-minimal` works without having to spend the hours re-downloading the template.

I was debating on adding it to the packages listed towards the bottom but I can't even get the `fedora-29-minimal` template to install on 3.2.  I am not really concerned with that since I will be upgrading my last 3.2 machine to 4.0 shortly but thought this little bit of information may help out the small niche of us who may encounter the issue that installed 4.0 and the `fedora-29-minimal` template at release. I think it is easier to find solutions in documentation once it has been proven than digging through the issue tracker.

Hopefully, when I upgrade my 3.2 machine to 4.0.1, I will not see that issue.

marmarek:
Since fixed template is in stable repository already, it isn't needed anymore.

PR #803, opened 2019-03-27
Old key expired

Adrianzo:


andrewdavidwong:
I'm inclined to say that we shouldn't print the actual key in our documentation. What do you think, @marmarek?

marmarek:
Including a key may be a good idea - as another place for key verification. I confirm this key is the same as I have recorded.

andrewdavidwong:
> Including a key may be a good idea - as another place for key verification. I confirm this key is the same as I have recorded.

Ok, merging.

PR #807, opened 2019-04-06
Update qubes-iso-building.md et al

awokd:
Change document name from `-r3-` to `-iso-` and update links
Update `./setup` steps
Update build environment from Fedora 26 to 29 and dependencies


awokd:
I attempted to use the https://github.com/QubesOS/qubes-issues/issues/4788 option and see `USE_QUBES_REPO_VERSION = $(RELEASE)` in my `builder.conf`, but GCC has been compiling for the past couple hours on a `make qubes`. Commenting here in case there's something I should update in the documentation while I'm at it, but can take it to the appropriate issue if not.

building/qubes-iso-building.md
```diff
@@ -1,9 +1,10 @@
 ---
 layout: doc
 title: Qubes ISO Building
-permalink: /doc/qubes-r3-building/
+permalink: /doc/qubes-iso-building/
```
marmarek: When changing permalink, please add redirect from the old one.

awokd: Added. I think that last CI run failed due to unrelated links to some `/doc/dom0-tools/`.

awokd:
I see in https://github.com/QubesOS/qubes-issues/issues/4158#issuecomment-408709757 I'm supposed to edit `builder.conf` to remove components I don't want to compile, but I'm not seeing them. I'll open an issue and separate pull if needed.

PR #809, opened 2019-04-09
Create nitrokey.md

neowutran:
some things I didn't know when I first tried to use the nitrokey


marmarek:
I don't think this remark deserve a separate page. Maybe a FAQ entry instead?

andrewdavidwong:
Agreed. We should not have a page dedicated to a specific brand of product.

PR #810, opened 2019-04-09
Create mount-hvm-disk.md

neowutran:
How to mount a HVM disk. 
Not directly related to Qubes OS ( related to LVM only) 
But useful to know


marmarek:
Mounting a VM disk in dom0 is a bad idea. You should use a (Disposable)VM for that instead. Right now a manual step is needed for that:
https://github.com/QubesOS/qubes-issues/issues/4687#issuecomment-451626625

debugging/mount-hvm-disk.md
```diff
@@ -1,35 +1,43 @@
...
-Disk Flags:
-Number       Start          End            Size           Type     File system  Flags
-1            1048576B       105906175B     104857600B     primary  ntfs         boot
-2            105906176B     314571751423B  314465845248B  primary  ntfs
+```bash
+qvm-run -v --dispvm=YOUR_DVM_TEMPLATE --service qubes.StartApp+xterm &
 ```
 
-Then mount the partition you want:
+3: Mount the partition you want to, and do what you want with it (disp)
```
marmarek: Missing step before this one: connect the device. Either using devices widget, or from command line using `qvm-block`.

Also it isn't clear that the thing in parenthesis is where to do this action.

marmarek:
To make it discoverable, it should also be linked from the doc index (doc.md)

PR #811, opened 2019-04-09
Create windows-gaming-hvm.md

neowutran:
How to play game on Qubes OS 
( instead of dual boot ) 

I don't know if such kind of doc can be accepted on the official Qubes OS doc, but can be useful for people who want to avoid to have a dual boot to play some games

### DIFF
--- a/configuration/windows-gaming-hvm.md
+++ b/configuration/windows-gaming-hvm.md
@@ -1,0 +1,163 @@
+---
+layout: doc
+title: Windows gaming hvm
+permalink: /doc/windows-gaming-hvm/
+---
+
+# Windows gaming HVM
+
+Some information to configure a windows HVM for gaming. 
+This is not officially supported, just some community trial & errors
+
+## References
+
+Everythings needed is referenced here
+
+- [Usefull technical details](https://paste.debian.net/1043341/)
+- [Reddit thread of what is needed for GPU passthrough](https://www.reddit.com/r/Qubes/comments/9hp3e7/gpu_passthrough_howto/)
+- [Solution to have more than 3Go of RAM in the Windows HVM](https://github.com/QubesOS/qubes-issues/issues/4321#issuecomment-423011787)
+- [Some old references](https://www.reddit.com/r/Qubes/comments/66wk4q/gpu_passthrough/)
+
+## Prerequise
+
+You have a functional Windows 7 HVM.
+
+The "how to" for this part can be found on the Qubes OS documentation and here: [Usefull github comment](https://github.com/QubesOS/qubes-issues/issues/3585#issuecomment-453200971).
+
+However, few tips:
+
+- Do a backup (clone VM) of the Windows HVM BEFORE starting to install QWT
+- The Windows user MUST BE "user"
+- Windows 7 Only, do not use Windows 10 or others.
+
+## Hardware
+
+To have a Windows HVM for gaming, you must have:
+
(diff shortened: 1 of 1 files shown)


configuration/windows-gaming-hvm.md
```diff
@@ -0,0 +1,163 @@
...
+```
+
+Then find the devices id for your secondary gpu.
+
+In my case, it is "dom0:0a_00.0" and "dom0:0a_00.1".
+
+Edit /etc/default/grub, and add the PCI hiding 
+
+```
+GRUB_CMDLINE_LINUX="....rd.qubes.hide_pci=0a:00.0,0a:00.1 modprobe=xen-pciback.passthrough=1 xen-pciback.permissive"
```
marmarek: `modprobe=xen-pciback.passthrough=1` looks weird, I think it does nothing.
`xen-pciback.permissive` - better to enable it for a particular device. You can do that with qvm-pci:
```
qvm-pci attach windows-hvm dom0:0a_00.0 -o permissive=True -o no-strict-reset=True
```

marmarek:
This is definitely something useful. Ideally all those custom steps wouldn't be needed, but until then, having them documented is a good thing.
Since it is clearly marked as not officially supported thing, I think it can go in.

configuration/windows-gaming-hvm.md
```diff
@@ -0,0 +1,163 @@
+---
+layout: doc
+title: Windows gaming hvm
+permalink: /doc/windows-gaming-hvm/
+---
+
+# Windows gaming HVM
+
+Some information to configure a windows HVM for gaming. 
+This is not officially supported, just some community trial & errors
```
andrewdavidwong: I think this would be much more appropriate for https://github.com/Qubes-Community. @awokd and others can help you with this.

andrewdavidwong: Also, we need to be careful about taking on too many optional docs. The more we have, the harder they are to maintain, and the less we are able to ensure high quality. See https://github.com/QubesOS/qubes-issues/issues/4693 for more on this.

awokd: > @awokd and others can help you with this.

Be happy to. @neowutran, I'll flag you over there.

configuration/windows-gaming-hvm.md
neowutran: Hi, I just checked, 

```
modprobe=xen-pciback.passthrough=1
```

was indeed useless. 
And the ``qvm-pci ... -o permissive=True`` replace correctly the ``xen-pciback.permissive``. 

But in fact, the "permissive" doesn't seems to be necessary at all, still running few tests, but seems like it is not required

I will update my doc & send it to Qube Community docs. 

Also:
- There is a mistake in the official qubes doc https://www.qubes-os.org/doc/pci-devices/#permissive doesn't contain the "permissive=True" 
- Does a doc with the full list of all the options we can pass ``-o xxxx`` exist ? 

Thanks ! :)

awokd: Who broke my document, it used to have the right setting in there! :) Anyways, `man qvm-device` describes the available options.

awokd: Submitted https://github.com/QubesOS/qubes-doc/pull/812 to fix the example.

neowutran: @awokd thanks :) 

And I confirm, all the "permissive" flag are unnecessary for this usecase

PR #814, opened 2019-04-15
Apply <lower> filter to <color.hex>.

FireGrace:
As now, some hex values are in uppercase. To unify the page, and prevent this from happening, we can apply lower filter.
Pros:
> Easy of copy-paste

### DIFF
--- a/basics_dev/style-guide.md
+++ b/basics_dev/style-guide.md
@@ -36,7 +36,7 @@ The following **grayscale** colors are currently used on the Qubes website and d
   <div class="swatch more-bottom more-right">
     <div class="color add-bottom bg-{{color.class}}"></div>
     <strong class="add-bottom">{{color.name}}</strong>
-    <code>#{{color.hex | lower}}</code>
+    <code>#{{color.hex | downcase}}</code>
   </div>
   {% endif %}
 {% endfor %}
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
This doesn't seem to make any difference when I serve the site locally.

andrewdavidwong:
I think you were looking for `downcase` rather than `lower`.

FireGrace:
True, `downcase` solves it perfectly. Thank you.

unman:
On Tue, Apr 16, 2019 at 09:11:07AM +0000, FireGrace wrote:
> True, `downcase` solves it perfectly. Thank you.
> 
It looks monstrous to me.
I must be old fashioned.

FireGrace:
> On Tue, Apr 16, 2019 at 09:11:07AM +0000, FireGrace wrote: True, `downcase` solves it perfectly. Thank you.
> It looks monstrous to me. I must be old fashioned.

I Agree, that does not solve the problem at it's core. Data in database is still the same.
But the data from database may be used somewhere else, changing it is a risk. Changing the only
one page, is much safer choice. 

But it does changes the presentation, which is the only concern (in this commit), regarding web page looks. 
And if the core idea of `color.hex` variable stays the same, this will negate future typos, in sense of Upper/Down case thing.

andrewdavidwong:
> It looks monstrous to me.

Which part looks monstrous to you? Lowercase rather than uppercase? The presence of `| downcase` in the source?

> I Agree, that does not solve the problem at it's core. Data in database is still the same.
But the data from database may be used somewhere else, changing it is a risk. [...]

I don't see why it would be risky, since they're just hex color codes. Both uppercase and lowercase letters are valid in such codes. You can always submit a PR against the source codes, if you like, but the casing doesn't really matter. As you recognize, it's just a very minor cosmetic issue.

PR #822, opened 2019-05-12
Update optical-discs.md

dancingonadragon:
More verbose language on just how to use a USB optical drive.

### DIFF
--- a/common-tasks/optical-discs.md
+++ b/common-tasks/optical-discs.md
@@ -18,3 +18,5 @@ Currently, the only options for reading and recording optical discs (e.g., CDs, 
  3. Use a SATA optical drive attached to dom0.
     (**Caution:** This option is [potentially dangerous](/doc/security-guidelines/#dom0-precautions).)
 
+To access an optical disc via USB follow the [typical procedure for attaching a USB device](/doc/usb-devices/#with-the-command-line-tool) (using the device-manager-icon widget is buggy) then check with the device-manager-icon widget to see what device in the target qube the USB optical drive was attached to. Typically this would be "sr0". For example, if sys-usb has device 3-2 attached to work qube's sr0 you would mount it with "$ mount /dev/sr0 /mnt/removable". You could also write to a disc with "$ wodim -v dev=/dev/sr0 -eject /home/user/Qubes.iso"
+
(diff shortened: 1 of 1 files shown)


dancingonadragon:
Not clear why this failed. First time user of GIT. Anything else I should do?

andrewdavidwong:
> using the device-manager-icon widget is buggy

Instead of putting this in the documentation, we would rather fix the bugs in the widget and have that be the user-friendly solution for using a USB optical drive. If the bugs you're experiencing haven't already been reported, please [report them](https://www.qubes-os.org/doc/reporting-bugs/).

> Not clear why this failed. First time user of GIT. Anything else I should do?

We use a simple typo/misspelling check that fails whenever it encounters an unknown spelling, in this case `sr0` and `wodim`. Since these spellings aren't erroneous, there's nothing to worry about.

dancingonadragon:
My experience with getting an optical USB drive working was needlessly confusing about the preferred way to attach and mount and, in some cases, write to these devices. I spent many hours trying to understand which of the many ways offered was the preferred and/or correct method for my situation. I needed a simple procedure to follow and I found this seriously lacking. The section about optical disks is the best place to do this. I believe fixing the widget is a great idea but those who need to use the terminal or those who want to know better how to control attachment themselves or those who don't have a fixed widget would all benefit from having clear and concise instructions. IMHO.
Not sure if I actually encountered any bugs at this point. Thank you.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

andrewdavidwong:
> My experience with getting an optical USB drive working was needlessly confusing about the preferred way to attach and mount and, in some cases, write to these devices. I spent many hours trying to understand which of the many ways offered was the preferred and/or correct method for my situation. I needed a simple procedure to follow and I found this seriously lacking. The section about optical disks is the best place to do this. I believe fixing the widget is a great idea but those who need to use the terminal or those who want to know better how to control attachment themselves or those who don't have a fixed widget would all benefit from having clear and concise instructions. IMHO. Not sure if I actually encountered any bugs at this point. Thank you.

I agree that it is good and important to have the type of documentation you are adding. My comment was just about the sentence "using the device-manager-icon widget is buggy," which I removed for the reasons explained above. This was a minor parenthetical remark in your text, and its removal doesn't detract from the main point of your documentation. Nonetheless, I wanted to provide you with an explanation for the change.

dancingonadragon:
Ok. I misunderstood the thrust of your comment. Makes sense. Thank you for the commit.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

PR #829, opened 2019-06-14
Recommend maintaining one list of firewall rules and calling it from elsewhere…

MatrixManAtYrService:
Instead of maintaining one rule list in `/rw/config/qubes-firewall-user-script` and another in `/rw/rc.local` I think it makes more sense to keep one list of rules and then just call one script from the other.

Had to add a `sudo` to get it to work though.

### DIFF
--- a/user/security-in-qubes/firewall.md
+++ b/user/security-in-qubes/firewall.md
@@ -49,9 +49,9 @@ firewall rules by hand. The firewall rules for each VM are saved in an XML file
 in that VM's directory in dom0:
 
     /var/lib/qubes/appvms/<vm-name>/firewall.xml
-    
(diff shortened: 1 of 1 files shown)


marmarek:
What exactly is a case where you needed to add `sudo`? Both of those scripts normally are called as root.

andrewdavidwong:
> What exactly is a case where you needed to add `sudo`? Both of those scripts normally are called as root.

Pinging @MatrixManAtYrService

andrewdavidwong:
Closing this for now due to the lack of response. If you revisit the matter in the future, or if anyone would like to take over this PR, please leave a comment, and we'll be happy to reopen this. Thank you.

PR #830, opened 2019-07-01
minor formatting

adrelanos:


adrelanos:
There's no consistency in Qubes docs. It's done on this page part of the
time, so I wanted to get it to all the time for consistency.

PR #837, opened 2019-07-15
Fix Outdated paragraph in Split GPG limitations section

tapaswenipathak:
I have just added a line for R4.

Fix https://github.com/QubesOS/qubes-issues/issues/4901.


andrewdavidwong:
If this is true, we should probably just remove that whole paragraph. @marmarek?

marmarek:
> If this is true, we should probably just remove that whole paragraph. @marmarek?

Generally yes. It requires graphical pinentry installed (like pinentry-gtk), text based one (the only one in minimal template by default) isn't enough.

andrewdavidwong:
> > If this is true, we should probably just remove that whole paragraph. @marmarek?
> 
> Generally yes. It requires graphical pinentry installed (like pinentry-gtk), text based one (the only one in minimal template by default) isn't enough.

Ok. If I understand correctly, this means that the existing text is still accurate, and the new sentence added by this PR is not really accurate. Closing.

If this is incorrect, please leave a comment or reopen.

PR #841, opened 2019-07-19
Add few docs for xorg.conf.d

tapaswenipathak:
Fix https://github.com/QubesOS/qubes-issues/issues/4208.


tapaswenipathak:
> The actual config fragment is ok, but description how to add it sounds weird.
What do you mean by "See what is present in /var/log/Xorg.0.log. This file has logs, edit the configuration and systemctl restart lightdm."? Definitely it should not suggest adding config options into the log file...

Good catch, that is a false statement and will be variable. Removed. Looks ok now?

PR #842, opened 2019-07-29
Added Whonix 14->15 Upgrade instructions

0ss:
Hello. 
# What I've done

- I created a directory for information regarding Whonix in the `external` directory.
- I have also created instructions on upgrading the default Whonix 14 Templates / AppVM to Whonix 15.
    - These instructions are taken from the Whonix wiki. 

# Why?
So there is Qubes-Whonix specific information on the Qubes wiki. Makes it easier for end users to find documentation related to what they are running.  

### Question
I do realize that this may be redundant. Should the Qubes documentation have such pages if the Whonix wiki already does?


andrewdavidwong:
> I do realize that this may be redundant. Should the Qubes documentation have such pages if the Whonix wiki already does?

No, I don't believe it should. Common scenario: A mistake is corrected in the Whonix wiki instructions, but no one remembers to update the duplicate page in qubes-doc. The qubes-doc page is now a trap for unsuspecting users who stumble into it and unwittingly follow outdated instructions. Traps are worse than gaps.

Duplicating content is almost always a bad idea. It's almost always better to maintain the content in a single place and *link* there from other places.

PR #843, opened 2019-08-02
Create Hardware Testing page

andrewdavidwong:


user/hardware/hardware-testing.md
```diff
@@ -0,0 +1,28 @@
...
+
+These hardware models have been (and continue to be) tested and work well with Qubes OS:
+
+ - Lenovo ThinkPad X1 Carbon Gen 5
+
+Note: The Lenovo X and T series are similar enough to assume similar compatibility of the matching model from the other series.
+
+## Desired Models
+
+The Qubes devs would like to tese these models, but the hardware is not available to us.
```
mfc: typo here

user/hardware/hardware-testing.md
```diff
@@ -0,0 +1,28 @@
...
+permalink: /doc/hardware-testing/
+---
+
+# Hardware Testing
+
+The Qubes developers test Qubes OS on certain hardware models.
+
+## Tested Models
+
+These hardware models have been (and continue to be) tested and work well with Qubes OS:
```
marmarek: I'd explicitly mention this is in addition to certified hardware.

user/hardware/hardware-testing.md
andrewdavidwong: Thanks, fixed.

user/hardware/hardware-testing.md
andrewdavidwong: I have significantly expanded the introduction to explain the relationship between tested hardware, certified hardware, and the HCL. Please let me know what you think.

marmarek: First two points in this list reminds me hardware support levels here: https://www.qubes-os.org/news/2017/07/08/toward-a-reasonably-secure-laptop/

andrewdavidwong: Ah, indeed, but Level 0 still seems distinct from what we have here, because it comes with a guarantee of compatibility. I think it would still be good to allow vendors to submit hardware models for Level 0 certification, but I don't think we want to commit to our own testing to being Level 0.

PR #849, opened 2019-08-11
uefi-troubleshooting.md: Remove <R4.0 content

awokd:
and a couple other misc edits

### DIFF
--- a/user/advanced-configuration/uefi-troubleshooting.md
+++ b/user/advanced-configuration/uefi-troubleshooting.md
@@ -14,7 +14,7 @@ If you've installed successfully in legacy mode but had to change some kernel pa
 
 **Change the xen configuration on a USB media**
 01. Attach the usb disk, mount the EFI partition (second partition available on the disk) 
-02. Edit your xen config (`xen.cfg/BOOTX64.cfg`) changing the `kernel` key to add your kernel parameters on the boot entry of your choice
+02. With `sudo`, edit your xen config (`EFI/BOOT/BOOTX64.cfg`) changing the `kernel` key to add your kernel parameters on the boot entry of your choice
 03. Install using your modified boot entry
 
 **Change xen configuration directly in an iso image**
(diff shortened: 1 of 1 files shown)


user/advanced-configuration/uefi-troubleshooting.md
```diff
@@ -14,7 +14,7 @@ If you've installed successfully in legacy mode but had to change some kernel pa
 
 **Change the xen configuration on a USB media**
 01. Attach the usb disk, mount the EFI partition (second partition available on the disk) 
-02. Edit your xen config (`xen.cfg/BOOTX64.cfg`) changing the `kernel` key to add your kernel parameters on the boot entry of your choice
+02. With `sudo`, edit your xen config (`EFI/BOOT/BOOTX64.cfg`) changing the `kernel` key to add your kernel parameters on the boot entry of your choice
```
marmarek: I think I see what you meant, but the first impression is "how to use sudo as an editor".

user/advanced-configuration/uefi-troubleshooting.md
```diff
@@ -25,77 +25,6 @@ If you've installed successfully in legacy mode but had to change some kernel pa
...
-   or if you're installing 3.2 execute:
-   
-        sed -i -e 's/^options=.*/\0 efi=attr=uc/' /mnt/sysimage/boot/efi/EFI/qubes/xen.cfg
-
-6. Now you can reboot the system by issuing `reboot` command.
-
-* * *
-<b name="f1">1</b> If you use rEFInd, you can see 3 options regarding the USB installer. Choose "Fallback Boot Loader" to enter the GRUB menu. [↩](#a1-1) [↩](#a1-2)
-
-
```
marmarek: Those two entries do apply to R4.0 too. The only difference is how to add Xen parameters (there is no grub on efi) - the way described above those two points (by editing config file).

awokd: Aren't `mapbs=1`, `noexitboot=1`, and `efi=attr=uc` all enabled by default in the R4.0 installer?

marmarek: They are on the installer, but not on installed system. If your system require them, you can still boot the installer as kind of rescue image (or do that just after installation, if you know that beforehand) and add them to the target /boot/efi

awokd: I see. I'd like to move them after the installer boot troubleshooting sections then, and make them generic. I'll start over with a different pull.

PR #847, opened 2019-08-08
[WIP] Rewrite qrexec documentation (Phase 1)

pierwill:
Greetings! :wave:

I'm embarking on a project to address the qrexec docs in response to

- https://github.com/QubesOS/qubes-issues/issues/1392;
- Qubes Google Season of Docs project idea (https://www.qubes-os.org/gsod/#rewrite-qrexec-documentation); and
- the need for this work identified by the SecureDrop team (@eloquence, @redshiftzero).

I'm opening this PR as a way to track my progress, share my work, and discuss the technical specifics of qrexec with the Qubes and SD teams. Hopefully people find this work useful!

PR #852, opened 2019-08-21
Update qrexec keyword characters

pierwill:
Per [QSB #38](https://github.com/QubesOS/qubes-secpack/blob/master/QSBs/qsb-038-2018.txt), this updates several (but not all) qrexec keywords in the docs to use the `@` character. Updating keywords in the qrexec docs themselves is being handled in the ongoing rewrite (see https://github.com/QubesOS/qubes-doc/pull/847).


marmarek:
Since we still use `$` in multiple places IMO it's worth add a note about deprecated `$` syntax. At least in rpc-policy.md.

pierwill:
Added a commit with a note in rpc-policy.md.

PR #853, opened 2019-08-24
uefi-troubleshooting.md: re-org and update

awokd:


awokd:
Is the note about rEFInd still relevant?

awokd:
> Is the note about rEFInd still relevant?

Don't believe so, so am deleting.

PR #854, opened 2019-08-24
Make Note of Viable Option for Non-Experts Wanting to Install Windows Qubes

crazyqube:
Any novice computer user will struggle to implement the steps provided. Give them and people who don't have the time to do it manually a viable option.

@marmarek seems to approve:
https://github.com/crazyqube/qvm-create-windows-qube/stargazers

### DIFF
--- a/external/os-guides/windows/windows-vm.md
+++ b/external/os-guides/windows/windows-vm.md
@@ -85,6 +85,8 @@ Windows VM installation
 ### qvm-create-windows-qube ###
 
 An unofficial, third-party tool for automating this process is available [here](https://github.com/crazyqube/qvm-create-windows-qube).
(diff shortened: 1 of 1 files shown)


marmarek:
Generally good idea, to link there, but there should be a clear disclaimer the tool isn't reviewed nor controlled by Qubes OS project
@andrewdavidwong do you think the current "unofficial, third-party tool" is enough?

andrewdavidwong:
> Generally good idea, to link there, but there should be a clear disclaimer the tool isn't reviewed nor controlled by Qubes OS project
> @andrewdavidwong do you think the current "unofficial, third-party tool" is enough?

I've made it even more explicit now.

PR #862, opened 2019-09-05
--hvm not an option, use --property virt_mode=mvh

BlackMaria:
I may be mistaken, but I do not have an `--hvm` option on my `qvm-create`.

OT: I suggest we use qubes versions in he doc's url, most of the docs I follow have bugs just like this

### DIFF
--- a/external/os-guides/windows/windows-tools.md
+++ b/external/os-guides/windows/windows-tools.md
@@ -153,6 +153,6 @@ Then, periodically check for updates in the Template VM and the changes will be 
 Once the template has been created and installed it is easy to create AppVMs based on it:
 
 ~~~
-qvm-create --hvm <new windows appvm name> --template <name of template vm> --label <label color>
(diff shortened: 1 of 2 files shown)


BlackMaria:
As it turns out, if this isn't just a bug on my side, there are several things such as `--mem=4096` that are also wrong.  if you want me to fix those too before merging, let me know and I will `grep -rI` the repo for like.

andrewdavidwong:
> OT: I suggest we use qubes versions in he doc's url, most of the docs I follow have bugs just like this

https://www.qubes-os.org/doc/doc-guidelines/#version-specific-documentation

Nonetheless, I'm coming around to the idea of segregating the documentation by version, simply because so many people keep asking for it. (The complaint will then be that there is no documentation for the latest version, which will be true.)

marmarek:
Regardless of version-specific documentation, qvm-create could get aliases for more common options, like `--mem` or `--hvm`. There is for example `--label` as an alias for `--property label=` or `--standalone` as an alias for `--class StandaloneVM` already.

PR #868, opened 2019-09-21
uefi-troubleshooting.md copies instead of moves, use xen-*.efi instead

awokd:


marmarek:
Why exactly? `xen.efi` nor `xen.cfg` are not going to be used in `EFI/BOOT`, copying would only waste space in there.

awokd:
You're right, I misread that. I'll limit to just the `-*` change.

PR #877, opened 2019-10-17
adding x230

mfc:
I think we can add a couple more models here, right? dev team is using more than just x1.


marmarek:
FWIW here is the current full list of machines I use for testing: https://lists.xenproject.org/archives/html/xen-devel/2019-10/msg00971.html
Some more often some less. Indeed x1 and x230 gets most of attention.

mfc:
you want to add some of the others that you test with? Thinkpad W540, Thinkpad P52, Dell Latitude E6420, Librem 14 v1 (may be too complicated to include due to versioning)?

marmarek:
Not really, as I test there only individual components mostly. Especially for the two thinkpads "I test" does not mean "it works" - there are issues caused mostly by nvidia GPU.

PR #880, opened 2019-11-06
Append /var/spool/mail to binds variable ...

SvenSemmler:
... to make received email persist through reboots.

### DIFF
--- a/external/configuration-guides/fetchmail.md
+++ b/external/configuration-guides/fetchmail.md
@@ -90,4 +90,18 @@ for rc in /usr/local/etc/fetchmail/*.rc; do
 done
 ~~~
 
+Make sure the folder '/rw/config/qubes-bind-dirs.d' exists.
+
+~~~
+sudo mkdir -p /rw/config/qubes-bind-dirs.d
+~~~
+
+Create the file '/rw/config/qubes-bind-dirs.d/50_user.conf' with root rights.
+
+Now edit it to append the '/var/spool/mail/' directory to the binds variable.
+
+~~~
+binds+=( '/var/spool/mail' )
+~~~
+
 Now reboot your AppVM and you are done.


andrewdavidwong:
qubes-users thread for context:

https://groups.google.com/d/msgid/qubes-users/bf55f561-c2bb-44ed-9d3c-3875946df8d2%40www.fastmail.com

SvenSemmler:
Guys, hold off for a moment. I think I didn't do the .procmail step. I'll verify

SvenSemmler:
@andrewdavidwong I am pretty sure the fault was on my part, but to verify I have to repeat all the steps exactly as outlined in the doc with another VM (which I am planing to do anyway). However, I will be traveling the next two weeks without my Qubes laptop. So I think it is best you reject this change request and I verify after my return. In case a change to the documentation is really necessary, i will send a new change request. 
Alternatively you could leave this open and assigned to me until I give feedback. Whatever works better for you. Sorry about this, I should have been more thorough before submitting the change request.
/Sven

andrewdavidwong:
> @andrewdavidwong I am pretty sure the fault was on my part, but to verify I have to repeat all the steps exactly as outlined in the doc with another VM (which I am planing to do anyway). However, I will be traveling the next two weeks without my Qubes laptop. So I think it is best you reject this change request and I verify after my return. In case a change to the documentation is really necessary, i will send a new change request.
> Alternatively you could leave this open and assigned to me until I give feedback. Whatever works better for you. Sorry about this, I should have been more thorough before submitting the change request.
> /Sven

No problem! We can just leave it open for now. Thanks for letting us know. :slightly_smiling_face:

SvenSemmler:
Hi @andrewdavidwong we should go ahead with the pull request. Thank you for your patience. (There are more issues with the mutt, fetchmail and postfix pages when applied to the latest Fedora, but this pull request at the very least avoids data loss as in my case). I'll do more changes at a later time.

andrewdavidwong:
Sorry for the delay, @SvenSemmler.

PR #888, opened 2019-11-21
Add `qubes-usb-proxy` as requirement for sys-usb

imjustmatthew:
When using a minimal template based sys-usb, `qubes-usb-proxy` appears to be an actual requirement, otherwise USB devices fail to attach. I'm not sure if this is also a bug (i.e., package `qubes-input-proxy-sender` should depend on `qubes-usb-proxy`), but it's definitely a documentation issue. I will open a separate issue for the dependency to see if there's traction to change that, but I think the documentation update should happen regardless.


imjustmatthew:
Updated branch with further changes based on discussion with Marek on this issue: https://github.com/QubesOS/qubes-issues/issues/5476

PR #889, opened 2019-11-22
Additional step included in section 2. Get the Release Signing Key

frederikfaye:
This makes it explicit that the release key should be imported. This may be obvious to some, but as a first-time user of both QubesOS and gpg2, this caused quite a bit of frustration.


andrewdavidwong:
The command immediately before the one you added already is the command to import the key. It even says "Once you've downloaded your Release Signing Key, import it with GPG:"

andrewdavidwong:
Also, the command you added omits this important option: `--keyserver-options no-self-sigs-only,no-import-clean`

PR #892, opened 2019-11-24
Update installation-guide.md

TheodoreChu:
Latest versions of Rufus tool ask for "dd" image. It doesn't show on the menu as in the picture.


andrewdavidwong:
Would you be able to provide an updated screenshot?

TheodoreChu:
Here they are. One more related problem with the installation instructions regarding Rufus is the next sentence: "Warning: If you do that on Windows 10, you can only install Qubes without MediaTest, which isn’t recommended." There's no recommendation as to what to do instead of that (on Windows 10). 

Rufus-3 8 1580-screenshot
Rufus-3 8 1580-screenshot-write-dd-image-mode

PR #894, opened 2019-11-25
Update faq.md

TheodoreChu:
added details about whether we can install Qubes without VT-x and VT-d


marmarek:
Looks like duplicate of #985

PR #901, opened 2019-11-30
Distinguish between template availability and support status

andrewdavidwong:
QubesOS/qubes-issues#5489

### DIFF
--- a/user/downloading-installing-upgrading/supported-versions.md
+++ b/user/downloading-installing-upgrading/supported-versions.md
@@ -16,15 +16,15 @@ or minor release (see [Version Scheme]). The current release and past major
 releases are always available on the [Downloads] page, while all ISOs, including
 past minor releases, are available from our [download mirrors].
 
-| Qubes OS      | Start Date | End Date   | Status                  |
-| ------------- | ---------- | ---------- | ----------------------- |
-| Release 1     | 2012-09-03 | 2015-03-26 | Old, unsupported        |
-| Release 2     | 2014-09-26 | 2016-04-01 | Old, unsupported        |
-| Release 3.0   | 2015-10-01 | 2016-09-09 | Old, unsupported        |
-| Release 3.1   | 2016-03-09 | 2017-03-29 | Old, unsupported        |
-| Release 3.2   | 2016-09-29 | 2019-03-28 | Old, unsupported        |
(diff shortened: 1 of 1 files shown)


user/downloading-installing-upgrading/supported-versions.md
```diff
@@ -75,6 +73,45 @@ Currently, only [Fedora] and [Debian] TemplateVMs are officially supported by th
...
+
+
+### Debian
+
+Qubes support for each [Debian] TemplateVM ends when that Debian release reaches [end-of-life][debian-eol], unless otherwise noted.
+
+| TemplateVM    | Status                  |
+| ------------- | ----------------------- |
+| Debian 7      | Old, unsupported        |
+| Debian 8      | Old, unsupported        |
```
andrewdavidwong: @marmarek, IIRC, you started a discussion about dropping support for Debian 8 on qubes-devel. I don't recall seeing a resolution to that question. Did you decide to drop support? If so, how should we represent that here?

marmarek: > Did you decide to drop support?

Yes.

> If so, how should we represent that here?

Something like this. TBH I don't understand what "old"/"current" means in this table - is it just another word for "unsupported"/"supported"?

marmarek:
I think there needs to be also information about Qubes release. For example currently Qubes Fedora 31 packages are available ("supported", if that applies in this case) for R4.1, but not R4.0.

user/downloading-installing-upgrading/supported-versions.md
andrewdavidwong: >  TBH I don't understand what "old"/"current" means in this table - is it just another word for "unsupported"/"supported"?

"Old" means that it is not the newest release. An old release can still be supported. For example, when Qubes 4.0 came out, 3.2 was still supported, but it was the old version.

However, we can drop "old"/"new" in order to simplify the table.

andrewdavidwong:
> I think there needs to be also information about Qubes release. For example currently Qubes Fedora 31 packages are available ("supported", if that applies in this case) for R4.1, but not R4.0.

That's what the existing table is for. It already reflects this kind of information.

However, it has not yet been updated with info about Fedora 31 and Qubes 4.1. I can update that now, as a separate matter from this PR.

andrewdavidwong:
Done.

PR #902, opened 2019-11-30
Add appvms-networkmanager.md

jdorel:


andrewdavidwong:
I'm skeptical that this merits its own dedicated page. Perhaps it should instead be added to one of the System docs in the Developer section. What do you think, @marmarek?

jdorel:
Maybe I could split this in two ?
- In UserDocumentation/AdvancedConfiguration/Networking
- In DevelopersDocumentation/System/Networking#NetworkManager_in_AppVMs

marmarek:
Yes, it doesn't warrant separate page. I'd add this all to the existing "VM network in Qubes" page. Besides the fact that created connections are persistent, I don't think user really should care about being it symlink or anything else.

andrewdavidwong:
Closing this for now due to the prolonged lack of activity. If you would like to take up this PR again, please leave a comment, and we'll be happy to reopen this. Thank you.

PR #903, opened 2019-12-03
Add guide on how to install qmenu

sine3o14nnae:
Add a short guide on how to install [qmenu](https://github.com/sine3o14nnae/qmenu) in dom0 to use dmenu for managing Qubes specific tasks.

The installation itself has serious security implications but users with issues like [#3910](https://github.com/QubesOS/qubes-issues/issues/3910), [#4294](https://github.com/QubesOS/qubes-issues/issues/4294), [#5037](https://github.com/QubesOS/qubes-issues/issues/5037), [#5386](https://github.com/QubesOS/qubes-issues/issues/5386) & [#5447](https://github.com/QubesOS/qubes-issues/issues/5447) could benifit from these tools.

### DIFF
--- a/external/configuration-guides/qmenu.md
+++ b/external/configuration-guides/qmenu.md
@@ -1,0 +1,92 @@
+---
+layout: doc
+title: Manage Qubes via dmenu
+permalink: /doc/qmenu/
+---
+
+
+Manage Qubes via dmenu
+==============================
+
+qmenu
+-----
+
+[qmenu](https://github.com/sine3o14nnae/qmenu/) is a collection of tools that utilize
+[dmenu](https://tools.suckless.org/dmenu/), a dynamic menu for X, to provide you with a drop down menu for Qubes specific tasks.
+
+When configured to execute these tools via hotkeys, you are able to list, start and stop your qubes, attach and detach your connected devices,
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
Since this is a guide for an unofficial, third-party tool, it should instead be added to the [Qubes Community docs](https://github.com/Qubes-Community).

sine3o14nnae:
Understood.

So this means that the [external documentation](https://www.qubes-os.org/doc/#external-documentation) is for third party documentation _that only includes trusted software_? Atleast when it comes to dom0 i suppose.

andrewdavidwong:
> Understood.
> 
> So this means that the [external documentation](https://www.qubes-os.org/doc/#external-documentation) is for third party documentation _that only includes trusted software_? Atleast when it comes to dom0 i suppose.

An understandable question. No, the external documentation section contains two types of entries:
* External links to documentation not hosted from `qubes-doc`
* External documentation that is to be migrated out of `qubes-doc` (but has not yet migrated)

In other words, yes, there are currently docs in the external section that are like this one, but we're trying to move them out of there. :slightly_smiling_face:

See https://github.com/QubesOS/qubes-issues/issues/4693 for more on this.

PR #905, opened 2019-12-05
fix instructions for Debian minimal templates

adrelanos:
by adding missing initramfs-tools package

https://github.com/QubesOS/qubes-issues/issues/5490#issuecomment-562298886


marmarek:
This shouldn't be needed after https://github.com/QubesOS/qubes-linux-utils/pull/48, but I'll keep it open until it's verified.

adrelanos:
I would suggest to merge this. Just extends an already long line that needs to be copied/pasted anyhow. It is easily removed later when VM kernel gets simplified or default or after https://github.com/QubesOS/qubes-linux-utils/pull/48 was deployed in Qubes stable.

marmarek:
This is already in R4.0 stable for some time.

PR #913, opened 2020-01-10
Update anonymizing-your-mac-address.md

3hhh:
add hostname randomization

PR #914, opened 2020-01-11
Update anonymizing-your-mac-address.md

3hhh:
Adds random hostname section.

Improved code of PR #913 to remove sleep.

### DIFF
--- a/external/privacy-guides/anonymizing-your-mac-address.md
+++ b/external/privacy-guides/anonymizing-your-mac-address.md
@@ -90,7 +90,7 @@ set -e -o pipefail
 #  network VM.
 # 2. Execute `sudo touch /etc/hosts.lock` inside the template VM of your network VM.
 # 3. Execute inside your network VM:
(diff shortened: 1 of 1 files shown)


3hhh:
It depends...

The chain is as follows:
1. systemd sets the hostname based on /etc/hostname or on machine ID (if e.g. /etc/hostname is empty)
2. Qubes OS sets it (or maybe later... cf. https://github.com/QubesOS/qubes-issues/issues/5570)
3. NetworkManager starts and executes pre-up --> the proposed script writes to /etc/hostname
4. NetworkManager sets the hostname based on the hostname-mode setting: This by default means based on /etc/hostname (it seems for Qubes OS). If NetworkManager deems the hostname _invalid_, it may even set a totally different one.

However you're right that I should set it with `hostname` to be sure it is set, e.g. for `hostname-mode=none`. I'll push that change.

Btw `hostnamectl` doesn't work during boot because dbus isn't started yet.

PR #915, opened 2020-01-13
Adds a current build results matrix.

mfp20:
The table is added to "spare a lot of building time to new users, smooth the learning curve, and save the world!". By looking at the table one can quickly understand the working/not-working build combos, and focus on his own needs instead of reinventing the wheel (and wasting his own time or, worse, add duplicate issues in the tracker). A bit of guidance for new enthusiasts try to solve their own issues by building new templates or testing the upcoming releases...

### DIFF
--- a/developer/building/qubes-builder.md
+++ b/developer/building/qubes-builder.md
@@ -187,3 +187,28 @@ By default Qubes developers' keys are imported automatically, but if you need so
 All Qubes developers' keys are signed by the Qubes Master Signing Key (which is set as ultimately trusted key), so are trusted automatically.
 
 If you are the owner of Master key and want to revoke such signature, use the `revsig` gpg key edit command and update the key in qubes-developers-keys.asc - now the key will be no longer trusted (unless manually set as such).
+
+Build Recipies Status Matrix
+----------------------------
+
+Community contributed results in order to keep track of current real capabilites. Details might be not accurate, your mileage may vary. Please read [Documentation Guidelines](https://www.qubes-os.org/doc/doc-guidelines/), then post your results.
+
+Editing help: most of the info is taken from the rpm file names found in \[builder root directory\]/qubes-src/linux-template-builder/rpm/noarch/ . Once you completed the build process you can easily add a row to the table by cutting and pasting the different parts of the file names. If the build process failed, creating the entry in this table needs more work: have a look at the logs and search the issue tracker for previous art, then report here.
+Please, don't post the same result twice, ie: have a look at the entry date and the builder commit, if those two are similar, it may not be needed to add another result row. Or might be useful to substitute the old entry with your more recent one. Let's try to keep the table tidy: it can spare a lot of building time to new users, smooth the learning curve, and save the world!
+
+|Entry Date|Component|Result|Used Qubes Release|Builder commit|Source Repo|Bin Package Repo|Notes|
+|:----------:|:---------:|:------:|:-------------------:|:--------------:|:------------:|:----------------:|:----------------------|
+|202001111016|fedora-30|Ok|4.0.1|?|marmarek/qubes- (Stable)|current|-|
+|202001111105|fedora-30-minimal|Ok|4.0.1|?|marmarek/qubes- (Stable)|current|-|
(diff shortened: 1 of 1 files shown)


marmarek:
I think such table will be outdated very quickly. @andrewdavidwong any opinion on including such dynamic content in qubes-doc repo?

As for the table itself:
 - builder commit is pretty meaningless in itself - most of the issues causing build failures are actually in other repos; more meaningful would be a snapshot of commit ids from all repos - see `make build-id`
 - the content already looks inaccurate - for example you listed `xenial` build successful with "Bin Package Repo" set to "current", while binary repo with qubes package does not exist for ubuntu (and only recently the setting is ignored, instead of failing the build)

andrewdavidwong:
> I think such table will be outdated very quickly. @andrewdavidwong any opinion on including such dynamic content in qubes-doc repo?

Yes, I'm also concerned that it will quickly go out of date. "Please, don't post the same result twice" and "Let's try to keep the table tidy" suggest that this would be more appropriate for a wiki or shared spreadsheet that multiple people can easily update whenever they want without having to pass qubes-doc review. Perhaps https://github.com/Qubes-Community might be a place for this.

mfp20:
Probably the best change is in the ncurses setup script. Example: hide options that can't build with current selection, ie: if I select R4.0, the fc31 and kali templates should be hidden, and then the builders for rpm and deb should be hidden too (unless I selected buster and fc30). Then an entry "Show experimental builds", that un-hide all.

I don't know what tool are you using for the setup; in the past I've been using the same of the kernel config and it was a nightmare. Some things simply can't be done without complex and cumbersome procedures, mostly relying on external tools, because the kconfig syntax is very limited and error prone. But if you can work in there instead of adding this table, is way better.

About the table itself: I made it too detailed. It has to be a fixed table {QubesRelases, WorkingTemplates}, to be updated once per year or on new releases only. By looking at the table I must be able to figure out that in R4.0 I can use fc30, but not fc31 and kali.

@marmarek 

I placed it there as an example, a starting point, to be modified on (eventual) merge. That's why there are inaccuracies. I just listed all the templates I've been able to produce in the 4 days before, using multiple different installs. The only accurate ones are the ones with builder commit id, because it's the builder I still have, and currently using for building.

@andrewdavidwong 

I'd prefer something less detailed and more reliable in the docs, rather than a messy report in the community pages. As I wrote up here, something like:

|Releases/Templates|Fedora29|Fedora30|Fedora31|Buster|Kali*|Xenial*|Bionic*|
|---|:-------------:|:-------------:|:-------------:|:--------:|:-----:|:---:|:---:|
|R3.2|ok|x|x|x|x|ok|ok|
|R4.0|x|ok|ok|x|x|ok|ok|
|R4.1|x|x|ok|ok|ok|x|ok|

[*] Not supported by QubesOS team, rely on community contribs.

marmarek:
Some of this info is already here: https://www.qubes-os.org/doc/supported-versions/

mfp20:
@marmarek I see. It's not the exact piece of information. That page tells the templates available IN the distribution, here we are talking of the scripts in the builder (to build the templates).

Ok, I've got enough feedback to do The Right Thing. Close this ticket, I'll push another merge request to edit 2 pages:
* this builder page, adding just a row of text with a link to the "supported versions" page you just linked here.
* the "supported versions" page, adding a new chapter after the whonix one: "Unsupported TemplateVMs" (available in Qubes Builder or external repos), where to place a couple line of text to describe the table, and the table for both experimental builds (ex: fc31 on R4.0) and community templates (ex: kali).

Looks good?

marmarek:
Yes, I think that would be much better.

mfp20:
It looks much better now. 

* the first patch points the users searching for QubesBuilder info to QubesBuilderDetails page.
* the second patch includes a new table for "extra TemplateVMs" in SupportedVersion, and tells the users they can go beyond QubesOS defaults, using the QubesBuilder.
* the third patch help the users to find the table, in order to not waste their time trying release/template combos known to fail on build.

andrewdavidwong:
> Ok, I've got enough feedback to do The Right Thing. Close this ticket, I'll push another merge request to edit 2 pages

I see that you've opened the other pull requests. I think you intended to close this one, so I'll close it now. If that's not what you intended, let me know.

PR #918, opened 2020-01-16
Adds "Note on custom TemplateVMs" chapter

mfp20:
In order to point advanced users to the right direction...

Set of patches, 2/3. See #915 discussion.

### DIFF
--- a/user/downloading-installing-upgrading/supported-versions.md
+++ b/user/downloading-installing-upgrading/supported-versions.md
@@ -127,6 +127,15 @@ To be precise:
 
 We aim to announce both types of events one month in advance in order to remind users to upgrade.
 
+### Note on custom TemplateVMs
+
+[QubesBuilder](/doc/qubes-builder/) includes both experimental TemplateVMs (ex: fc31 backported to R4.0) and community-contributed builders for custom TemplateVMs (ex: kali). They are not supported by QubesOS itself and not included in QubesOS ISOs, but can be built by experienced users.
+
+| Qubes OS      | TemplateVM    | Status       |
+| ------------- | ------------- | ------------ |
+| Release 4.0   | Fedora 31 | Fail |
+| Release 4.0   | Kali      | Fail |
+| Release 4.1   | Kali      | Builds |
 
 [Version Scheme]: /doc/version-scheme/
 [Downloads]: /downloads/


user/downloading-installing-upgrading/supported-versions.md
```diff
@@ -127,6 +127,15 @@ To be precise:
...
 
+### Note on custom TemplateVMs
+
+[QubesBuilder](/doc/qubes-builder/) includes both experimental TemplateVMs (ex: fc31 backported to R4.0) and community-contributed builders for custom TemplateVMs (ex: kali). They are not supported by QubesOS itself and not included in QubesOS ISOs, but can be built by experienced users.
+
+| Qubes OS      | TemplateVM    | Status       |
+| ------------- | ------------- | ------------ |
+| Release 4.0   | Fedora 31 | Fail |
+| Release 4.0   | Kali      | Fail |
+| Release 4.1   | Kali      | Builds |
```
marmarek: Is that really true? AFAIR the problem was missing `salt` package in Debian bullseye, which shouldn't depend on Qubes version. See https://tracker.debian.org/pkg/salt

mfp20: No idea, I didn't investigate the fail. I just reported what I've got. Feel free to adapt the table itself.

mfp20:
@andrewdavidwong do I have to do something else to clear this one?

andrewdavidwong:
> @andrewdavidwong do I have to do something else to clear this one?

I think we were waiting to see if you had any further response to @marmarek's review. Since you don't, I'll return to @marmarek for a decision on how to proceed from a technical content perspective.

user/downloading-installing-upgrading/supported-versions.md
marmarek: Kali template fails to build regardless of qubes version.

marmarek:
In any case, it may be good to add a column with some reference (either link to report of failed a build or successful build). In some cases such reports include some extra steps needed, but in all the cases it allows to reason about freshness of the information.

BTW this table is already outdated, as changes required for Fedora 31 are now backported to R4.0: https://github.com/QubesOS/qubes-issues/issues/5289#issuecomment-582442319

mfp20:
> In any case, it may be good to add a column with some reference (either link to report of failed a build or successful build). In some cases such reports include some extra steps needed, but in all the cases it allows to reason about freshness of the information.

I can do it but ... where do I get those reports? You mean the build log? 

> BTW this table is already outdated, as changes required for Fedora 31 are now backported to R4.0: [QubesOS/qubes-issues#5289 (comment)](https://github.com/QubesOS/qubes-issues/issues/5289#issuecomment-582442319)

Ok, let's do like this: 
* please, accept this patch so I don't need to remember what I did and why. There's more text changed in the page.
* I pull the changes in my local repo, then edit the table to add the reports column and refresh the table content (ie: f31 success, kali fail). I don't have time to re-install the builder VM and try f31 myself, I give for good that you have successfully built it.

marmarek:
> I can do it but ... where do I get those reports? You mean the build log?

You put information in the table based on something, right? Such table should be based on someone trying it (and saying that publicly), not just a guess.

mfp20:
> > I can do it but ... where do I get those reports? You mean the build log?
> 
> You put information in the table based on something, right? Such table should be based on someone trying it (and saying that publicly), not just a guess.

I gave for granted that we (users) report on pull, and you (developers) confirm on commit. As we are doing here, I reported "f31 fail, kali success"; then you asked "f31 NOW works", and I replied "I got a fail, I report a fail". Then you said "kali fail". So I replied "Ok, so I modify the table contents according to your reports: f31 success, kali fail".

If you move from this setup to a trust-based one, you must trust whoever submit a success/fail. Linking a report doesn't change much, there's anyway some margin of error on any trust-based approach.
An option can be to make the Builder create automatically a "build report", so that it can be attached to any Builder related activity. Then accept only entries with that report attached, whether it is a fail or a success.

andrewdavidwong:
I think this exchange shows that it would be too much maintenance for the Qubes team to keep such a table reliably up-to-date. Perhaps it might be better to host this over at https://github.com/Qubes-Community or even https://qubes-os.discourse.group/, where more users could potentially contribute and achieve a Wikipedia-like reliability via frequent, open contributions.

CC: @awokd, @deeplow

deeplow:
> or even https://qubes-os.discourse.group/

@mfp20 We can make a wiki post over at https://qubes-os.discourse.group/. Then users can easily update it just like a regular wiki.

PR #924, opened 2020-01-21
Fix some links in the FAQ

strugee:
Not all of them are fixed, but I found these changes sitting on my disk from a while ago and figured I might as well send what I have so far. Note that I haven't previewed this yet; I can do that soon to double-check that the changes worked.

Ref QubesOS/qubes-issues#5537

### DIFF
--- a/introduction/faq.md
+++ b/introduction/faq.md
@@ -96,6 +96,12 @@ in a few clicks. Users concerned about privacy will appreciate the
 [Tor] securely, while those concerned about physical hardware attacks will
 benefit from [Anti Evil Maid].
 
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
These empty brackets shouldn't be necessary to create working links.

strugee:
Hm, funky. Is that a Jekyll extension?

In any case, I pushed an amended commit.

andrewdavidwong:
> Hm, funky. Is that a Jekyll extension?
> 

Pretty sure it's just standard Markdown syntax.

> In any case, I pushed an amended commit.

Thanks!

PR #925, opened 2020-01-21
Improve Firefox notes for the U2F proxy

strugee:
* Remove notes about pre-Firefox Quantum versions, as these are *long* outdated
* Be clearer about U2F being deprecated and how that affects Firefox support

Maybe we should also just say outright "the real fix for this problem is for the proxy to be updated to FIDO2/Webauthn"?

### DIFF
--- a/user/security-in-qubes/u2f-proxy.md
+++ b/user/security-in-qubes/u2f-proxy.md
@@ -114,16 +114,16 @@ For example, if your `banking` qube enrolls your banking key, and your `twitter`
 
 ## TemplateVM and browser support
 
-The large number of possible combinations of TemplateVM (Fedora 27, 28; Debian 8, 9) and browser (multiple Google Chrome versions, multiple Chromium versions, multiple Firefox versions) made it impractical for us to test every combination that users are likely to attempt with the Qubes U2F Proxy.
+The large number of possible combinations of TemplateVM (Fedora 27, 28; Debian 8, 9) and browser (multiple Google Chrome versions, multiple Chromium versions) made it impractical for us to test every combination that users are likely to attempt with the Qubes U2F Proxy.
 In some cases, you may be the first person to try a particular combination.
 Consequently (and as with any new feature), users will inevitably encounter bugs.
 We ask for your patience and understanding in this regard.
 As always, please [report any bugs you encounter].
 
-Please note that, in Firefox before Quantum (e.g. Firefox 52 in Debian 9), you have to install the [U2F Support Add-on][ff-u2f-addon].
-In Firefox post-Quantum you may have to enable the `security.webauth.u2f` flag in `about:config`.
-Chrome and Chromium do not require any special browser extensions.
-
+Unfortunately, the U2F protocol that Qubes U2F Proxy supports has been superseded by a newer, better-standardized protocol: Web Authentication, also known as Webauthn.
+Chrome and Chromium support U2F out of the box, but Firefox does not.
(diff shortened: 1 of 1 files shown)


user/security-in-qubes/u2f-proxy.md
```diff
@@ -114,16 +114,16 @@ For example, if your `banking` qube enrolls your banking key, and your `twitter`
...
 As always, please [report any bugs you encounter].
 
-Please note that, in Firefox before Quantum (e.g. Firefox 52 in Debian 9), you have to install the [U2F Support Add-on][ff-u2f-addon].
-In Firefox post-Quantum you may have to enable the `security.webauth.u2f` flag in `about:config`.
-Chrome and Chromium do not require any special browser extensions.
-
+Unfortunately, the U2F protocol that Qubes U2F Proxy supports has been superseded by a newer, better-standardized protocol: Web Authentication, also known as Webauthn.
+Chrome and Chromium support U2F out of the box, but Firefox does not.
+Therefore if you want to use Qubes U2F Proxy with Firefox you need to enable `security.webauthn.u2f` in `about:config`, but this is not recommended because U2F is deprecated technology.
+Mozilla has no plans to enable this flag by default in the future, and while there are no plans to remove it, Mozilla does not guarantee that it will always be available.
```
woju: This is all wrong.

1. `security.webauth.u2f` (not `webauthn`!) is enabled by default on non-Android builds: https://github.com/mozilla/gecko-dev/blob/e878e5b81bb319c141900ce9cfcde732df5c8449/modules/libpref/init/StaticPrefList.yaml#L7719-L7722
2. WebAuthn standard provides explicit way to support legacy U2F-only tokens. If we're to believe [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Web_Authentication_API#Browser_compatibility) this is the only way Firefox supports WebAuthn.
3. This works with Firefox from Debian 10 and the proxy. Just tested with https://webauthn.bin.coffee/.

So the conclusion that "this is not recommended, because U2F is deprecated technology" is incorrect. I'd just remove the reference to both addon and `about:config`.

strugee: Hey @woju, thanks for the review and sorry it took me so long to reply! dev-platform confirms they turned it on on Firefox desktop as a compat thing: https://groups.google.com/forum/#!searchin/mozilla.dev.platform/security.webauth.u2f%7Csort:date/mozilla.dev.platform/q5cj38hGTEA/p8mH-wS2AwAJ

So I was misinformed - thank you for the correction. I think the implementation notes on MDN are ambiguous; I read them as "Firefox does not support U2F over e.g. NFC" but I suppose that's irrelevant for us :P

In any case, I'll push a fix.

strugee:
Ok, just pushed a fix removing the incorrect claims about U2F in Firefox. One concern - what if a user registers a FIDO2 device on e.g. another computer, or with USB passthrough, and then tries to use this device through the U2F Proxy? Did you test that scenario? If not I can. I think that might not work and we might need to add a note about it.

woju:
LGTM. Thanks!

> One concern - what if a user registers a FIDO2 device on e.g. another computer, or with USB passthrough, and then tries to use this device through the U2F Proxy? Did you test that scenario? If not I can. I think that might not work and we might need to add a note about it.

On default setup, this will work, as there is no reason (no protocol possibility) to break. I think I tested this once, but if you'd like to test it yourself, please do! There is no automatic test for that, because it would be hard to set up.

user/security-in-qubes/u2f-proxy.md
woju: The fix is OK.

strugee:
Interesting, you're right! Just tested it, including with direct attestation. Feel free to merge, this LGTM.

PR #929, opened 2020-01-26
LibreOffice was almost unusable slow in Debian 10

kramse:
I dont know if this is the right place, but LibreOffice 6.1.5-3 on Debian 10 Buster was definitely different than before on Debian 9. LibreOffice is also an application that I expect lots of users use.


demokir:
This is not right place. you have to reach Qubes mailing list.

andrewdavidwong:
@kramse, the issue you filed (https://github.com/QubesOS/qubes-issues/issues/5600) is sufficient for now. It is not appropriate to document this here, but thank you for the effort to help!

PR #934, opened 2020-02-02
Update centos.md

mkarasoff:
The repo argument is wrong.  Updated per [qubes-users] Is CentOS in Qubes Broken?

### DIFF
--- a/external/os-guides/centos.md
+++ b/external/os-guides/centos.md
@@ -15,7 +15,7 @@ For the minimal version, please see [Minimal TemplateVMs](/doc/templates/minimal
 
 CentOS-7 can be installed with the following command:
 
-    [user@dom0 ~]$ sudo qubes-dom0-update --enablerepo=qubes-templates-community qubes-template-centos-7
+    [user@dom0 ~]$ sudo qubes-dom0-update --enablerepo=qubes-templates-community-testing qubes-template-centos-7
 
 To switch, reinstall and uninstall a CentOS TemplateVM that is already installed in your system, see *How to [switch], [reinstall] and [uninstall]*.
 


andrewdavidwong:
Wouldn't this change have to be reverted when CentOS-7 migrates to stable soon? I think it would be better to say something like: "Use the latest that's available in stable, unless you're willing to test something unstable, in which case pick what you want from testing."

CC: @marmarek @fepitre

fepitre:
Yes indeed. It's applicable to all community templates. BTW, the template should be in stable very soon.

andrewdavidwong:
Assigning to @mkarasoff to make the suggested change.

mkarasoff:
I don't feel I know or understand enough about differences between testing and stable to make that suggested change.  I understand why the Qubes team may want to make the change, but from my perspective that really doesn't address the underlying issue, which is the directions to install CentOS that exist today are just plain wrong with the latest updates of Qubes.  Presumably the CentOS image broke at some point in the past, but I don't know when.  I just know the instructions today are wrong.

A naive user such as myself, following the documentation as it exists now, can easily walk away from this situation thinking Qubes is buggy or incomplete.  The change I proposed makes it capable to install CentOS without frustration, and the "testing" repo is a clear indication to me of the testing state of the template.  The idea of stable vs. testing in Qubes is well documented in other places by people who deeply understand the difference with respect to Qubes.  I use that guidance.

The suggestion "Use the latest that's available in stable..." implies that a stable version of the template exists, which it doesn't.  The suggestion also points to a bigger policy within Qubes of testing/stable labels, which I get as a user, but understand only superficially, and am in no position to document.  If calling out the install from testing process that works today is the wrong thing to do, then maybe I should be logging a bug rather than trying to change the documentation?  Either way, the fact that instructions as they exist don't currently work should get captured somewhere.

marmarek:
Anyway, CentOS templates are already in stable repository.

mkarasoff:
OK, then I'm quite confused:

```
[mike@dom0 Pictures]$ sudo qubes-dom0-update --enablerepo=qubes-templates-community qubes-template-centos-7
Using sys-firewall as UpdateVM to download updates for Dom0; this may take some time...
Qubes Community Templates repository            1.6 kB/s | 5.4 kB     00:03    
No match for argument: qubes-template-centos-7
Error: Unable to find a match
```

marmarek:
Maybe you have cached old metadata, try adding `--clean` option.

mkarasoff:
```
[user@dom0 ~]$ sudo qubes-dom0-update --clean --enablerepo=qubes-templates-community qubes-template-centos-7
Using sys-firewall as UpdateVM to download updates for Dom0; this may take some time...
2 files removed
Fedora 25 - x86_64 - Updates                    3.4 MB/s |  24 MB     00:07    
Fedora 25 - x86_64                              3.4 MB/s |  50 MB     00:14    
Qubes Dom0 Repository (updates)                 433 kB/s |  16 MB     00:38    
determining the fastest mirror (15 hosts).. done..0 kB/s | 2.7 kB     00:00 ETA
Qubes Templates repository                      4.2 kB/s |  14 kB     00:03    
Qubes Community Templates repository            2.5 kB/s | 5.4 kB     00:02    
No match for argument: qubes-template-centos-7
```
Error: Unable to find a match

marmarek:
Hmm, interesting, I've just tried:
```
$ sudo qubes-dom0-update --enablerepo=qubes-templates-community qubes-template-centos-7
Using sys-firewall as UpdateVM to download updates for Dom0; this may take some time...
Qubes Community Templates repository             12 kB/s | 5.4 kB     00:00    
Last metadata expiration check: 0:00:00 ago on Thu Feb  6 21:57:46 2020.
Dependencies resolved.
================================================================================
 Package              Arch   Version            Repository                 Size
================================================================================
Installing:
 qubes-template-centos-7
                      noarch 4.0.1-201909141946 qubes-templates-community 1.1 G

Transaction Summary
================================================================================
Install  1 Package

Total download size: 1.1 G
Installed size: 4.0 G
DNF will only download packages for the transaction.
Downloading Packages:
```
Note which repository is next to the template name. I also don't see a typo in your command.
To be absolutely sure you are getting fresh metadata, lets try disabling alternative mirrors: in `/etc/yum.repos.d/qubes-templates.repo` commend out `metalink` line for this repository and uncomment `baseurl` one.
The template was uploaded just few hours ago, so it might not propagate to all the mirrors yet.

mkarasoff:
Thanks!  I will wait some hours for the propagation, try again, and post the result.

mkarasoff:
It is working for me now.  Thanks!

PR #936, opened 2020-02-07
Clarify procedure in CLI step 2, more.

tasket:
Change 'Edit and add' to better phrase.

Change nano to gedit.

Update 'Usage' to reflect R4.0 proxy capabilities.

### DIFF
--- a/external/configuration-guides/vpn.md
+++ b/external/configuration-guides/vpn.md
@@ -95,9 +95,11 @@ Set up a ProxyVM as a VPN gateway using iptables and CLI scripts
 ----------------------------------------------------------------
 
 This method is more involved than the one above, but has anti-leak features that also make the connection _fail closed_ should it be interrupted.
-It has been tested with Fedora 23 and Debian 8 templates.
+It has been tested with Fedora 30 and Debian 10 templates.
 
-1. Create a new VM, name it, click the ProxyVM radio button, and choose a color and template.
+Before proceeding, you will need to download a copy of your VPN provider's configuration file(s) and have your VPN login information handy.
+
+1. Create a new VM, name it, choose "provides network", and choose a color and template.
 
    ![Create\_New\_VM.png](/attachment/wiki/VPN/Create_New_VM.png)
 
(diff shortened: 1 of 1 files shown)


external/configuration-guides/vpn.md
```diff
@@ -304,16 +305,11 @@ Configure your AppVMs to use the VPN VM as a NetVM...
 
 ![Settings-NetVM.png](/attachment/wiki/VPN/Settings-NetVM.png)
 
-If you want to be able to use the [Qubes firewall](/doc/firewall), create a new FirewallVM (as a ProxyVM) and set it to use the VPN VM as its NetVM.
-Then, configure AppVMs to use your new FirewallVM as their NetVM.
-
-If you want to update your TemplateVMs through the VPN, enable the `qubes-updates-proxy` service in your new FirewallVM.
+If you want to update your TemplateVMs through the VPN, it will be necessary to create a separate "sys-update" or "sys-firewall-vpn" VM with the "provides network" option set and with its netvm set to use the VPN VM. Next, use `qubes-global-settings` to change the update VM to the one you created. Then enable the `qubes-updates-proxy` service in your new update VM.
```
marmarek: Why is separate VM necessary? Just isolation, or there is something more? Since openvpn uses routing table (compared to transparent Tor, which use iptables redirect), it should work for local processes out of the box too.

In any case, one step is missing: making templates to use this VM as updates proxy. You can just link to https://www.qubes-os.org/doc/software-update-domu/#updates-proxy

tasket: Good question. The firewall is configured to not allow any other OUTPUT besides the VPN client. For the proxy to work, I think it would need to be run under the `qvpn` group. Can you suggest an alternative?

marmarek: That's a good point about the current stat of the script. On the other hand, it should be safe to impose this restriction (requiring qvpn group) only on eth0 interface, but allow any process to talk through tun0 (or other vpn interface). Then, running updates proxy as it is will just work - and will be properly filtered out if vpn disconnects.

tasket: I'll defer to your judgement about "should be safe". But IIRC when I wrote the script I was thinking that VPN VM programs have access to both the clearnet identity and tunnel identity of the user; probably why I chose a blanket policy.

marmarek: If we consider actively malicious applications in VPN VM, then such application can probably leak info regardless of the rule (especially it can try to remove those rules). But if we consider only accidental leaks, application don't really have (accidental) access to the clearnet identity, as it is cut of clearnet right from the start (blocking rules applied at VM startup). One thing that still could possibly link those two is VPN configuration (certificate, user) - but this again looks like requiring rather active exfiltrating instead of accidental application doing its thing.
In summary: I think it is safe to modify OUTPUT chain to allow traffic through VPN for any user.
BTW this is how Whonix Gateway is configured too - any user from there directly can connect through tor. And there is special user that is allowed clearnet access.

tasket: The restriction was definitely intended to protect against accidental misuse, i.e. occasional wrong command in VPN VM that leaks info. But I think you're right: blocking tun+ isn't really necessary, so I'll change it.

tasket:
OK. In this commit I also changed `ping` to use the more neutral `1.1.1.1` address.

PR #940, opened 2020-02-15
Create gui-not-working.md

jdorel:


andrewdavidwong:
We might consider adding this to https://www.qubes-os.org/doc/gui-configuration/ rather than creating a new, separate page.

jdorel:
> We might consider adding this to https://www.qubes-os.org/doc/gui-configuration/ rather than creating a new, separate page.

I only saw the title `GUI Configuration` in the page list. Just moved it to the correct page, also renamed it.

PR #945, opened 2020-02-17
make image a hyperlink to the PrivacyBeast website

mfc:
I think this makes more sense from a usability standpoint


andrewdavidwong:
The idea was to be able to click the responsively-resized (hence possibly small) image on the page to obtain the full-size image by itself, i.e., a standard "click for full size" image. But this is fine too.

PR #941, opened 2020-02-15
Reformat packages into table, remove unneeded info

deathgrippin:
- Reformat the package listing within the Fedora section into a table, consistent with the Debian and CentOS sections
- Reformat a package name to use code style
- Remove packages mentions that are unneeded with the new table

PR #947, opened 2020-02-22
Move formatting from tables to lists

deathgrippin:
Change formatting , as requested in https://github.com/QubesOS/qubes-doc/pull/941


andrewdavidwong:
Thanks. The formatting looks fine to me. However, it looks like there are also some content changes included in this patch, so let's see if @marmarek has any comments.

PR #949, opened 2020-03-02
Remove broken image

deathgrippin:


deathgrippin:
Added another small fix.

PR #957, opened 2020-03-07
clarify language around openvpn config files

WestleyArgentum:
While trying to follow this doc I was quite confused by the line `Files referenced in openvpn-client.ovpn should not use absolute paths such as /etc/...`

Many openvpn config files will include absolute paths like `up /etc/openvpn/update-resolve-conf`. I don't think replacing those with relative paths would make much sense.

I began following the git history back and here: https://github.com/QubesOS/qubes-doc/blame/2778fbccc8f6ecc71091223a4fa339148774c344/configuration/vpn.md#L79 it becomes clear that originally the warning about absolute paths was meant to be specifically about accompanying *.crt and *.pem files.

### DIFF
--- a/external/configuration-guides/vpn.md
+++ b/external/configuration-guides/vpn.md
@@ -124,13 +124,11 @@ Before proceeding, you will need to download a copy of your VPN provider's confi
        sudo mkdir /rw/config/vpn
 
    Copy your VPN configuration files to `/rw/config/vpn`.
-   Your VPN config file should be named `openvpn-client.ovpn` so you can use the scripts below as is without modification. Otherwise you would have to replace the file name. Files accompanying the main config such as `*.crt` and `*.pem` should also be placed in the `/rw/config/vpn` folder.
+   Your VPN config file should be named `openvpn-client.ovpn` so you can use the scripts below as is without modification. Otherwise you would have to replace the file name. Files accompanying the main config, such as `*.crt` and `*.pem`, should also be placed in the `/rw/config/vpn` folder. Make sure to check that these accompanying files are not referenced by absolute paths such as `/etc/...` inside `openvpn-client.ovpn`.
 
    Check or modify configuration file contents using a text editor:
 
        sudo gedit /rw/config/vpn/openvpn-client.ovpn
-
-   Files referenced in `openvpn-client.ovpn` should not use absolute paths such as `/etc/...`.
 
    The config should route all traffic through your VPN's interface after a connection is created; For OpenVPN the directive for this is `redirect-gateway def1`.
 


tasket:
There shouldn't be any absolute path references in the ovpn. The example you gave for `up` isn't correct and the doc actually shows how you are supposed to replace `up` and `down` with relative paths in step 4. So I don't agree with this edit.

WestleyArgentum:
Oh! You're totally right, apologies for taking your time before reading through the whole doc.

If you can spare a little more time, however, I think there are a couple possible changes that might help people who are working through this doc in the future...

```
Files referenced in openvpn-client.ovpn should not use absolute paths such as /etc/...
```

1) Is there some sort of practical or security related concern about using absolute paths? I did a bit of research online but wasn't able to come up with anything. Looking back at the original wording here: https://github.com/QubesOS/qubes-doc/blame/2778fbccc8f6ecc71091223a4fa339148774c344/configuration/vpn.md#L79 it is not clear this was meant as important security advice.

    If there is a reason, I would love to help capture it in the docs. If there isn't, maybe that line doesn't actually need to be there?

2) Assuming there is a reason that we shouldn't use absolute paths, it would still be nice to make it clear to people working through the guide that they don't have to remove all of the absolute paths before moving forward. I was hesitant to move forward because the next steps involved testing the connection and I hadn't figured out how to remove all the absolute paths yet.

    Would you be open to a different PR that changed that line to say: 

```
Files referenced in openvpn-client.ovpn should not use absolute paths such as "/etc/...".
Make sure to update the paths to accompanying files (such as "*.crt" and "*.pem") now, and
we will change other paths (like for "up" and "down") in later steps.
```

tasket:
1. There's no security concern wrt that statement. Its meant as a cue to the reader so that if they notice absolute paths then its likely not pointing where it should (i.e. the script they're currently creating).  I've seen a few examples of ovpn that have absolute paths for their cert files bc the configs came with instructions to use /etc/openvpn... so the statement covers those as well, and I think its helpful.

2. I think if there was a situation where absolute paths were useful (instead of a probable indicator of misconfiguration in this use case) it would already be stated. I don't recall the phrase being questioned before this, however. And the doc already tells the user .crt and .pem files should reside in the ../vpn dir with the ovpn, so these statements work together.

Also the vpn doc isn't considered to be the ideal way to setup a vpn, just a way to demonstrate how it can be done on Qubes by addressing the necessary areas. There is a process underway to incorporate a version of [Qubes-vpn-support](https://github.com/tasket/Qubes-vpn-support) in Qubes in the future to simplify the process bc no file editing is needed.

WestleyArgentum:
Ok, obviously we think about things differently, feel free to close. 

One last question if you have time though. Looking into Qubes-vpn-support, I see a lot of overlap with the scripts in the vpn doc but also some differences. Which set of instructions do you think is better for people to follow right now? In the end are there any usability or security differences?

andrewdavidwong:
After reading the discussion, it sounds like there is mutual agreement that this PR can be closed. (If not, please let me know.)

PR #969, opened 2020-03-27
Document MSG_CURSOR

pwmarcz:
See QubesOS/qubes-issues#1551.

### DIFF
--- a/developer/system/gui.md
+++ b/developer/system/gui.md
@@ -255,11 +255,20 @@ struct msg_window_hints {
   <td>MSG_WINDOW_FLAGS</td>
   <td><pre>
 struct msg_window_flags { 
-      uint32_t flags_set; 
+     uint32_t flags_set;
      uint32_t flags_unset;
 };
 </pre> </td>
  <td>Change window state request; fields contains bitmask which flags request to be set and which unset</td>
+</tr>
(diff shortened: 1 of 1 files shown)


pwmarcz:
Travis build is only: `The following new unknown spellings were introduced: 0x100`

developer/system/gui.md
```diff
@@ -9,6 +9,11 @@ redirect_from:
 - /wiki/GUIdocs/
 ---
 
+<style>
+/* Override Bootstrap's nowrap (breaks table layout) */
+article table { white-space: normal; }
+</style>
+
```
andrewdavidwong: Instead of this, shouldn't we add to or edit the CSS files?

pwmarcz: Sounds good. Do we have a custom CSS file? I couldn't find it.

marmarek: see main repo: https://github.com/QubesOS/qubesos.github.io/

pwmarcz: Okay, I see now - the `white-space: nowrap` hack was added because of the admin API table (https://github.com/QubesOS/qubes-issues/issues/853). I'll try to do something about it, and for now I'm removing this fix from here.

PR #976, opened 2020-04-13
Updated "Building archlinux template"

scrouthtv:
When we make the individual components the doc was missing `make app-linux-split-gpg-vm` which would, if chosen this way, not be built and thus resulting in an error.

Please further refer to https://github.com/QubesOS/qubes-issues/issues/5769#issuecomment-612684970 as well as the makefiles.


andrewdavidwong:
Since `building-archlinux-template.md` is in External Documentation, I'm going to go ahead and merge this.

PR #981, opened 2020-04-26
Edited

clayt0nk:
I have stubbed my toe on this stuff twice in the last month or so, so I think a little formal documentation is in order.

Note that I was not clear on what to do with redirect_from list, so I just edited them like they were a template.


andrewdavidwong:
> Note that I was not clear on what to do with redirect_from list, so I just edited them like they were a template.

`redirect_from` is generally used when pages are moved/renamed in order not to break existing links. With a brand new page, there are no existing links pointing to it, so `redirect_from` is not necessary (but doesn't hurt either).

PR #988, opened 2020-05-09
Update templates.md

vseshu:
Adding text based on knowledge I learned from the user forum about how to apply a new default templateVM to sys-usb for devices that are dependant upon sys-usb running for keyboards and mice, etc. I found this to be a very useful set of scripts to run.

### DIFF
--- a/user/managing-os/templates.md
+++ b/user/managing-os/templates.md
@@ -137,6 +137,16 @@ When you install a new template or upgrade a clone of a template, it is recommen
         [user@dom0 ~]$ qvm-features new-template-dvm appmenus-dispvm 1
         [user@dom0 ~]$ qubes-prefs default-dispvm new-template-dvm
 
+4. Updating the template for sys-usb if peripheral devices are dependent upon the VM
+    
+    If you are running Qubes on a Desktop or other device where the peripheral devices such as keyboard / mouse / trackpad are dependant upon the sys-usb appVM then updating the template is a challenge. In this situation, you can use the following commands in a dom0 terminal window to update the templateVM.
+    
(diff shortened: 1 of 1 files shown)


user/managing-os/templates.md
```diff
@@ -137,6 +137,16 @@ When you install a new template or upgrade a clone of a template, it is recommen
         [user@dom0 ~]$ qvm-features new-template-dvm appmenus-dispvm 1
         [user@dom0 ~]$ qubes-prefs default-dispvm new-template-dvm
 
+4. Updating the template for sys-usb if peripheral devices are dependent upon the VM
+    
+    If you are running Qubes on a Desktop or other device where the peripheral devices such as keyboard / mouse / trackpad are dependant upon the sys-usb appVM then updating the template is a challenge. In this situation, you can use the following commands in a dom0 terminal window to update the templateVM.
+    
+        $ qvm-shutdown --wait sys-usb; \
+        $ qvm-prefs sys-usb template fedora-31; \
+        $ qvm-start sys-usb
```
marmarek, line 146: It is important for this to be on one command - putting `$` at each line will break it. But also adding space after `\` (which is easily done by mistake). Better list all three in one line.

vseshu: aahh, good point. ok so the preferred command should be
$ qvm-shutdown --wait sys-usb | qvm-prefs sys-usb template fedora-31 | qvm-start sys-usb 

I'm a bit of a newbie to this, so did I get the command line correct?

marmarek: Semicolons like you did initially are fine.

vseshu: Thanks!

PR #994, opened 2020-05-17
Update gsod.md

sarru1291:
[gsod.md] href link of `improved getting started page` redirect to completed project idea which is removed from the list. So, redirects to nowhere.

Hence, I've replaced it with a real getting started page so contributors can look at it.


andrewdavidwong:
Reference already removed in a separate commit, but thank you!

PR #998, opened 2020-05-23
USB Devices: Make requirement for USB qube more visible & remove redundant sentence

WillyPillow:
I was playing with `qvm-usb` a few days ago, only to find, after some failures and head-scratching, that a USB qube is required.

Issues such as QubesOS/qubes-issues#4345 also show the confusion of other users, so it should make sense to move the requirement to the beginning where it is more visible.

Also, I find that the sentence "`qubes-usb-proxy` should be installed..." to only reiterate what was being said in the previous paragraph, so I opt to remove it.

### DIFF
--- a/user/common-tasks/usb-devices.md
+++ b/user/common-tasks/usb-devices.md
@@ -100,7 +100,6 @@ This section exists for reference or in case something broke and you need to rei
 Under normal conditions, `qubes-usb-proxy` should already be installed and good to go.
 
 If you receive this error: `ERROR: qubes-usb-proxy not installed in the VM`, you can install the `qubes-usb-proxy` with the package manager in the VM you want to attach the USB device to.
-`qubes-usb-proxy` should be installed by default in the template VM.
 
 - Fedora: `sudo dnf install qubes-usb-proxy`
 - Debian/Ubuntu: `sudo apt-get install qubes-usb-proxy`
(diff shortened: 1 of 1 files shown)


user/common-tasks/usb-devices.md
```diff
@@ -12,6 +12,8 @@ redirect_from:
 
 If you are looking to handle USB *storage* devices (thumbdrives or USB-drives), please have a look at the [block device] page.
 
+**Note:** you cannot pass through devices from dom0 (in other words: a [USB qube][USB-qube howto] is required).
```
andrewdavidwong, line 15: I think this sentence is very confusing when removed from its previous context (and perhaps it was already confusing before). Seeing it near the top of the page like this, I'm not even sure how to interpret it.

andrewdavidwong: If I were a user reading this page for the first time, after reading this sentence, I would probably wonder:
 - Pass through to where?
 - Why can't I pass through devices from dom0 if that's where they are? How else am I supposed to get them out?
 - A USB qube is required for what?

WillyPillow: Perhaps it is less confusing to rephrase the sentence to something like the following?

> Attaching USB devices to VMs requires a [USB qube](USB-qube howto).

One can also move the sentence to the "Creating And Using a USB qube" section, though again the visibility may not be as high.

(For the record, my original justification for placing the sentence there untouched was that, as the term "USB passthrough" was used quite frequently in the first section and throughout the page, I ended up assuming the concept to be prior knowledge. I understand the concerns though.)

andrewdavidwong: I think your new proposed sentence is much clearer and simpler.

WillyPillow: The updated version is pushed.

PR #1002, opened 2020-05-30
Fix wrong numbering

sarru1291:


andrewdavidwong:
The numbering is not exactly "wrong," since the existing numbering is allowed in Markdown and will render correctly. However, using actual numbering is preferred for source readability, so thank you for that. However, I'm confused about why you also removed some of the correct punctuation.

sarru1291:
I found that redundant. So, it's cool if you think that's correct.

andrewdavidwong:
Ah, I see now. It's actually the indentation that's causing the numbering not to render correctly.

PR #1010, opened 2020-06-10
modified:   vpn.md Resolving lack of current hook for /rw/config/qubes-firewall-user-script

Lemlay:


Lemlay:
Nah, that was just me making a mistake in the title.

I can't find the issues that I had previously seen regarding the lack of a hook, but in practice that script has never run on any of my systems (except for adding a personal hook in /rw/config/rc.local) and I recall others having the same issue. 

Maybe I should clarify removing any dynamically added rules due to downstream VMs before running the iptables-save, but the point is to create a standard iptables configuration, removing the need for adding the individual rules in a script that doesn't seem to get called, and since rc.local is run when the VM comes up this method sets that standard iptables configuration (which, obviously could be modified by a user if they so desire) after the clean set of iptables rules are applied, any dynamic configuration that occurs from connecting a downstream VM, or from the VPN having DHCP configured DNS rules would be added after this initial preparation.

andrewdavidwong:
@tasket, would you be willing to give this a quick review?

tasket:
Review:

The addition of `iptables-save` & `iptables-restore` commands is understandable because its closer to Linux admin conventions. However, its incorrect in a Qubes proxyVM environment where iptables rules are (very) dynamic; the edit is saving/restoring ephemeral state from a different runtime session creating the potential for a GIGO failure. Beyond that, the attempt to make the script more conventional doesn't add anything functional.

The correct way to setup iptables is to use the qubes-firewall features (i.e. 'qubes-firewall-user-script'), and having it issue `iptables` commands reduces the script complexity (no need to create/restore a data file). So I'd have to recommend against using the edits here.

andrewdavidwong:
Ok, thanks for the review, @tasket.

Based on this review, I'm going to close this PR, but I greatly appreciate your desire to contribute and invite you to do so again in the future, @Lemlay.

P.S. -- This document is in the "External" category, which will probably be moving out of `qubes-doc` soon (see https://github.com/QubesOS/qubes-issues/issues/4693).

PR #1011, opened 2020-06-18
GSOC: Admin API Fuzzer idea

3hhh:
Just an idea. If you don't like it, ditch it.


3hhh:
> I have no idea how to practically do this - especially checking if nothing unexpected happened - because we don't only care about crashes, but also unexpected side effects. But I like the idea.

For dom0 the journal error entries would be a good bet. No error is likely suspicious. I guess it heavily depends on the RPC call in question though.

For VM <--> VM calls it should be even more straightforward:
Somehow measure a VMs state (processes running, snapshot the FS, whatever, ...), execute an invalid RPC, measure again & check diff. If there's any diff, it's suspicious.

PR #1024, opened 2020-07-29
Make template upgrade documentation user-friendly

qtpies:
Upgrading TemplateVM's by installing a new TemplateVM, then redo changes is way more user friendly, and technically there are benefits too. This option is currently behind a link at the bottom of this page , but it should be at the top. 

- User friendlyness: The commands involved are less in number, less error prone, and far less complex.
- The technical benefit is that you are getting a fresh OS, which can not have issues that are possibly inherited from older versions. The fact that this is possible without reinstalling your laptop is actually one of the benefits of Qubes.
- Of course there are users who are lacking diskspace or who are on a slow or metered internet connection. They can still use the other method.

I want to explicitly include the part 'There is currently ... dependencies.' (line 33). Installing extra packages is the main thing users do in a TemplateVM, so including this saves users time from searching for such a method themselves, and it will prompt somebody who in fact does know a way to report this, because it would be a great addition. To be clear, a list of packages to be installed, should not include dependencies, because when the user manually installs packages formerly installed as dependencies, the packagemanager will not remove them automatically anymore when they are no longer needed.

My addition deviates a little bit from the succinct and technical style of the rest of the page, but given that upgrading templates is a frequent operation in Qubes that no user can avoid, I think this is fair.

### DIFF
--- a/user/managing-os/debian/debian-upgrade.md
+++ b/user/managing-os/debian/debian-upgrade.md
@@ -12,7 +12,7 @@ redirect_from:
 
 # Upgrading Debian TemplateVMs
 
-[Should I upgrade using a fresh installation or an in-place upgrade?][Upgrading Fedora TemplateVMs]
+[Should I upgrade using a fresh installation or an in-place upgrade?][Upgrading Debian TemplateVMs]
 
 This page provides instructions for performing an in-place upgrade of an installed [Debian TemplateVM].
 If you wish to install a new, unmodified Debian TemplateVM instead of upgrading a template that is already installed in your system, please see the [Debian TemplateVM] page instead.
--- a/user/managing-os/debian/debian-upgrade.md
+++ b/user/managing-os/debian/debian-upgrade.md
@@ -1,6 +1,6 @@
 ---
 layout: doc
-title: Upgrading Debian TemplateVMs
+title: In-place upgrade of Debian TemplateVMs
 permalink: /doc/template/debian/upgrade/
 redirect_from:
 - /doc/template/debian/upgrade-8-to-9/
@@ -11,6 +11,8 @@ redirect_from:
 ---
 
 # Upgrading Debian TemplateVMs
+
+[Should I upgrade using a fresh installation or an in-place upgrade?][Upgrading Fedora TemplateVMs]
 
 This page provides instructions for performing an in-place upgrade of an installed [Debian TemplateVM].
 If you wish to install a new, unmodified Debian TemplateVM instead of upgrading a template that is already installed in your system, please see the [Debian TemplateVM] page instead.
--- a/user/managing-os/debian/debian.md
+++ b/user/managing-os/debian/debian.md
@@ -46,7 +46,9 @@ Please see [Updating software in TemplateVMs].
 
 ## Upgrading
 
-Please see [Upgrading Debian TemplateVMs].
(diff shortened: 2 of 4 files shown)


qtpies:
The build fails because it doesn't know the word 'textfile'? I guess this is a false positive?

```
+_utils/spellcheck.sh /home/travis/old_site _site

    +textfile

The introduced suspects in context are:

./_doc/user/managing-os/fedora/fedora.md:43:There are two ways to upgrade a TemplateVM. The easiest way is to [install] the new Fedora TemplateVM next to the Fedora TemplateVM you are currently using. Then redo all desired template modifications, and switch everything that was set to the old template to the new template.To make this process as efficient as possible, document modifications to your TemplateVMs in a textfile. If you do not have this documentation yet, open a terminal in the old Fedora TemplateVM, and use the `history` command. (There is currently no other way to gain a list of explicitly installed packages. Methods like `dnf repoquery --userinstalled` and `rpm -qa` all include packages that have been installed as dependencies.)

./_doc/user/managing-os/debian/debian.md:51:There are two ways to upgrade a TemplateVM. The easiest way is to [install] the new Debian TemplateVM next to the Debian TemplateVM you are currently using. Then redo all desired template modifications, and switch everything that was set to the old template to the new template.To make this process as efficient as possible, document modifications to your TemplateVMs in a textfile. If you do not have this documentation yet, open a terminal in the old Debian TemplateVM, and use the `history` command. (There is currently no other way to gain a list of explicitly installed packages. Methods like `apt list --installed`, `dpkg -l` and `aptitude search '~i!~M'` all include packages that have been installed as dependencies.)

If these are false-positives, simply commit as is and they will

be ignored in the future.

+all_ok=false
```

andrewdavidwong:
> The build fails because it doesn't know the word 'textfile'? I guess this is a false positive?

The way the spell checker works is that it flags every word it doesn't recognize. If it's a word that we consider correct, we simply ignore that flag and proceed. Then, it won't flag that word in the future. This is what it means when it says, "If these are false-positives, simply commit as is and they will be ignored in the future."

The preferred spelling seems to be "text file" (with a space), but "textfile" is sometimes used ([Wikipedia](https://en.wikipedia.org/wiki/Text_file)).

user/managing-os/debian/debian-upgrade.md
```diff
@@ -12,6 +12,8 @@ redirect_from:
 
 # Upgrading Debian TemplateVMs
 
+[Should I upgrade using a fresh installation or an in-place upgrade?][Upgrading Fedora TemplateVMs]
```
andrewdavidwong, line 15: The link says Fedora, but I think you meant Debian.

andrewdavidwong:
Thanks for your contribution!

By the way, the `history` command just gives you your terminal history, so it's not a very good way to get the information you're looking for. I've gone through and made some improvements before pushing this live.

qtpies:
Nice to have this merged! Just to clarify, I actually do think though that `history` is the best way to find out which packages have been manually installed. I didn't litteraly mean just `history`, but something like `history | grep 'dnf install'`. I just thought that explaining that literally would be more than necessary for the intended audience. This gives me a list of packages that I can copy paste into a text document or terminal without second thought. In terms of userfriendlyness that is what I think is necessary.

`dnf repoquery --userinstalled --queryformat "%{name}"` and `apt-mark showmanual` are I think the commands that come closest,  but both still include dependencies in my local test. For Debian I think few people who are on the technical level that they need to read the manual, can really make sense of the dpkg or apt logs. 

Anyway, I don't mean to get academic on this so let's leave it like this, thanks for merging.

PR #1028, opened 2020-08-21
Update software-update-domu.md

ghijg:
When adding rpmfusion repositories it's important to add the related update repositories, too.

Without enabling update repos it leads to problems like this:
https://ask.fedoraproject.org/t/dnf-not-updating-correctly/8416/12
https://www.reddit.com/r/Qubes/comments/847x8e/cant_download_package_from_rpm_fusion/

PR #1035, opened 2020-09-07
Add RPM Fusion 'Updates' repositories

de-souza:
Closes QubesOS/qubes-issues#4667 (see last comment)


de-souza:
This problem was also addressed in #1028

andrewdavidwong:
> This problem was also addressed in #1028

Unless I'm missing something, it doesn't make sense to have both of these PRs open, since they try to do basically the same thing and will conflict. Closing this one since the other was open first.

PR #1036, opened 2020-09-10
Update uefi-troubleshooting.md

Germayn:
A clean solution to https://github.com/QubesOS/qubes-issues/issues/5383

Instead of changing:
default=qubes-verbose to default=qubes 

which helped for me as well, I tracked down the problem to be the "options=console=vga" parameter.
Changing to "options=console=none" allowed a proper qubes os installation.


marmarek:
This is included in https://github.com/QubesOS/qubes-doc/pull/1048 and merged already.

PR #1039, opened 2020-09-16
Update disposablevm-customization.md

92VV3M42d3v8:
More consistent with qubes philosphy.


andrewdavidwong:
> More consistent with qubes philosphy.

I would say "defaults" rather than "philosophy," since the Qubes philosophy is that the meanings of the labels are entirely up to the user. :slightly_smiling_face:

PR #989, opened 2020-05-10
Rework Kali Template documentation

neowutran:
My issues with the current Kali template documentation
-------------------------------------------------------------------------
- Contains and start with obsolete QubesOS version (r3.X)
- The "Katoolin" tools haven't been updated for quite some times, never succeeded to create a Kali template with it.
- The package "kali-linux-full" doesn't exist anymore

Main modification I am trying to make
--------------------------------------------------
* Delete all the obsolete part, and keep only documentation for 4.0, and only the method "from the iso" and the method "from a debian template"
* Try to explain the advantages and disadvantages of both way of creating a Kali VM, and recommend one

PR #1040, opened 2020-09-21
Describe additional Windows installation options

GWeck:
Added instructions for installing QWT on Windows 10, using the Xen version 9.0.0 drivers and added instructions on creating Windows 7 and 10 as TemplateVMs and AppVMs instead of StandaloneVMs.

### DIFF
--- a/external/os-guides/windows/windows-vm.md
+++ b/external/os-guides/windows/windows-vm.md
@@ -117,7 +117,7 @@ To install Qubes Windows Tools, follow instructions [below](#xen-pv-drivers-and-
 
 MS Windows versions considerations:
 
-- The instructions *may* work on other versions than Windows 7 x64 but haven't been tested. 
+- The instructions *may* work on other versions than Windows 7 x64 but haven't been tested.
 - Qubes Windows Tools (QWT) only supports Windows 7 x64. Note that there are [known issues](https://github.com/QubesOS/qubes-issues/issues/3585) with QWT on Qubes 4.x
 - For Windows 10 under Qubes 4.0, a way to install QWT 4.0.1.3, which has worked in several instances, is described below.
 
@@ -227,7 +227,7 @@ So to get a working Windows 10 system (Standalone or Template VM) under Qubes R4
 **end of replaced text**
 
 **new text**
-- In the Windows 10 VM, download the installation kits of `xenvbd` and `xenbus` Version 9.0.0 (two files`xenvbd.zip`and `xenbus.zip`) from the Xen web site and the file `qubes-tools-4.0.1.3.exe` from https://www.qubes-os.org/doc/windows-tools/ **enter the final url** and store them on the Windows system drive (normally `C:\`.)
+- In the Windows 10 VM, download the installation kits of `xenvbd` and `xenbus` Version 9.0.0 (two files`xenvbd.tar`and `xenbus.tar`) from the Xen web site and the file `qubes-tools-4.0.1.3.exe` from https://www.qubes-os.org/doc/windows-tools/ **enter the final url** and store them on the Windows system drive (normally `C:\`.) In order to extract the contents from the tar-archives, you will need an external utility like 7zip.
 **end of new text**
 
 - Check the integrity of the file `qubes-tools-4.0.1.3.exe`by comparing its hash checksum. This can be done using the Windows command `certutil` specifying an appropriate hash algorithm like:
@@ -236,9 +236,9 @@ certutil --hashfile qubes-tools-4.0.1.3.exe SHA256
 ~~~
 This utility supports the algorithms MD5, SHA1, SHA256 and SHA512 (to be entered in uppercase!). The correct hash values can be retrieved from the Qubes website: https://www.qubes-os.org/doc/windows-tools/ **enter the final url**
 
-- Install `xenvbd` and `xenbus` version 9.0.0 by starting the file `dpinst.exe` from the `x64` directories of the Zip-file.
+- Install `xenvbd` and `xenbus` version 9.0.0 by starting the file `dpinst.exe` from the `x64` directories of the extracted tar-files. If during installation, the Xen driver requests a reboot, select "No" and let the installation continue.
 - After installation, reboot.
-- Install Qubes Windows Tools 4.0.1.3 by starting `qubes-tools-4.0.1.3.exe`, not selecting the Xen storage driver and the `Move of user data to Drive D` (which would probably lead to problems in Windows, anyhow).
+- Install Qubes Windows Tools 4.0.1.3 by starting `qubes-tools-4.0.1.3.exe`, not selecting the `Xen PV disk drivers` and the `Move user profiles` (which would probably lead to problems in Windows, anyhow). If during installation, the Xen driver requests a reboot, select "No" and let the installation continue - the system will be rebooted later.
 - Shut down Windows.
 - Set `qvm-features win10new gui 1`
 - Reboot Windows. The VM starts, but does not show any window.
@@ -269,7 +269,7 @@ when creating the VM. To have the user data stored in AppVMs depending on this t
 
 -  For Windows 7, the option to move the user directories from drive `C` to drive `D` works and causes any user data to be stored in the AppVMs based on this template, and not in the template itself.
 
--  After installation of Windows 10 as a TemplateVM, the Windows disk manager may be used to add the private volume as disk `D:`, and you may, using the documented Windows operations, move the user directories `C:\users\<username>\Documents` to this new disk, allowing depending AppVMs to have their own private volumes. Moving the hidden application directories `AppData`, however, is likely to invite trouble - the same trouble that occurs if, during QWT installation, the option `Move of user data to drive D` is selected.
+-  After installation of Windows 10 as a TemplateVM, the Windows disk manager may be used to add the private volume as disk `D:`, and you may, using the documented Windows operations, move the user directories `C:\users\<username>\Documents` to this new disk, allowing depending AppVMs to have their own private volumes. Moving the hidden application directories `AppData`, however, is likely to invite trouble - the same trouble that occurs if, during QWT installation, the option `Move user profiles` is selected.
 
 For Windows 10, configuration data like those stored in directories like `AppData` still remain in the TemplateVM, such that their changes are lost each time the AppVM shuts down. In order to make permanent changes to these configuration data, they have to be changed in the TemplateVM, meaning that applications have to be started there, which violates and perhaps even endangers the security of the TemplateVM. Such changes should be done only if absolutely necessary and with great care. It is a good idea to test them first in a cloned TemplateVM before applying them in the production VM.
 
--- a/external/os-guides/windows/windows-vm.md
+++ b/external/os-guides/windows/windows-vm.md
@@ -213,7 +213,11 @@ At that point you should have a functional and stable Windows VM, although witho
 
 ### Installing Qubes Windows Tools on Windows 10
 
-If the Xen bus and storage drivers version 9.0.0 are installed in a Windows 10 system without Qubes Windows Tools, and QWT 4.0.1.3 are installed after the Xen installation has finished, the Qubes interface works correctly. Files can be exchanged with other VMs, and the common clipboard works in both directions. So to get a working Windows 10 system (Standalone or Template VM) under Qubes R4.0, the following steps should be performed:
+If the Xen bus and storage drivers version 9.0.0 are installed in a Windows 10 system without Qubes Windows Tools, and QWT 4.0.1.3 are installed after the Xen installation has finished, the Qubes interface works correctly. Files can be exchanged with other VMs, and the common clipboard works in both directions.
+
+The installation of Qubes Windows Tools should **not** be done by using the parameter `--install-windows-tools`or by directly specifying `--cdrom=...`when starting the Windows VM, as this is bound to crash the VM on booting, showing the error `INACCESSIBLE BOOT DEVICE`- which makes no sense, but does happen.
+
+So to get a working Windows 10 system (Standalone or Template VM) under Qubes R4.0, the following steps should be performed:
 
 - Install Qubes Windows Tools in dom0: `sudo qubes-dom0-update qubes-windows-tools`. The iso will be the file `/usr/lib/qubes/qubes-windows-tools-4.0.1.3.iso`.
 - Copy this file to some AppVM: `qvm-copy-to-vm VMname /usr/lib/qubes/qubes-windows-tools-4.0.1.3.iso`.
--- a/external/os-guides/windows/windows-vm.md
+++ b/external/os-guides/windows/windows-vm.md
@@ -117,8 +117,9 @@ To install Qubes Windows Tools, follow instructions [below](#xen-pv-drivers-and-
 
 MS Windows versions considerations:
 
-- The instructions *may* work on other versions than Windows 7 x64 but haven't been tested.
+- The instructions *may* work on other versions than Windows 7 x64 but haven't been tested. 
 - Qubes Windows Tools (QWT) only supports Windows 7 x64. Note that there are [known issues](https://github.com/QubesOS/qubes-issues/issues/3585) with QWT on Qubes 4.x
+- For Windows 10 under Qubes 4.0, a way to install QWT 4.0.1.3, which has worked in serverel instances, is described below.
 
 Create a VM named win7new in [HVM](/doc/hvm/) mode (Xen's current PVH limitations precludes from using PVH):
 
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
@deeplow, would you be willing to review this? Mainly looking for someone to verify factual accuracy by testing the instructions.

deeplow:
> @deeplow, would you be willing to review this? Mainly looking for someone to verify factual accuracy by testing the instructions.

:+1: @andrewdavidwong I can try to reproduce this over this or the next week. Just a small disclaimer: I'm not terribly familiar with windows. But the Qubes integration is something I've been interested in trying.

Also thanks @GWeck for making the instructions. Do you know if I can use the official windows ISOs for testing purposes like these?

GWeck:
@deeplow I used the German W10 1909 ISO which I built using the media creation tool. So I think that it's just the standard iso. As product key I used the one installed in the firmware and read out from Linux. When the system was running there were no problems activating the license.

If there should be any problems during your tests, I'll gladly try to help! As always with windows, anything can happen ...

That the checks of the Travis pull request fail seems to be caused by the word xenbus, which is unknown to the spell checker. I'll try to fix that, but I'm not sure if I succeed.

deeplow:
@GWeck I followed the guide and in the end I managed to get it working (As a StandaloneVM). But there was one part in particular where I had to find my way around a problem (see bellow).

I will try installing it as an TemplateVM in the next days. 

I have some further feedback that I'd like to give which I'll also do later.

## Blue Screen on Booting with Qubes Tools ISO
So, I did get the famous blue screen when running `qvm-start win10 --install-windows-tools`

win10_qubes_tools_blue_screen_of_death

However, I managed to go around this by:
* downloading the RPM pacakge
* extracting `qubes-windows-tools-4.0.1.3.iso` out of the package
* mouting the `iso` to a windows drive
* and finally the executing the `.exe`

andrewdavidwong:
> +1 @andrewdavidwong I can try to reproduce this over this or the next week. Just a small disclaimer: I'm not terribly familiar with windows. But the Qubes integration is something I've been interested in trying.

Thanks! Please feel free to take your time.

> That the checks of the Travis pull request fail seems to be caused by the word xenbus, which is unknown to the spell checker. I'll try to fix that, but I'm not sure if I succeed.

The spell checker works by flagging any words it doesn't recognize. So, if the spelling is actually correct, you can leave it as is. After this is merged, the spell checker will recognize that word and no longer flag future occurrences.

GWeck:
@deeplow I had something similar. There seems to be a problem attaching the QWT iso as a CD drive. That's why we have to get the exe into the VM somehow and start it there. As far as I remeber I copied the exe via SMB from a separate Windows system.

GWeck:
@deeplow I tried to start Windows AppVMs with the parameter `--install-winddows-tools`. For Windows 7, the VM started, but no CDROM showed up. For Windows 10, I got the same crash as you did. If I use the parameter `--cdrom=...,` the result is the same one.

andrewdavidwong:
I'm about to move external documents like this one out of `qubes-doc` (see https://github.com/QubesOS/qubes-issues/issues/4693). However, I don't want to move this out from under you just as you're working on this PR, so I'll wait until this is merged this first. (Still no need to rush; just letting you know.)

deeplow:
> @deeplow I had something similar. There seems to be a problem attaching the QWT iso as a CD drive. That's why we have to get the exe into the VM somehow and start it there. As far as I remember I copied the exe via SMB from a separate Windows system.

@GWeck that crash should be mentioned. Or at least provide the alternative path while that is not resolved.

@andrewdavidwong the qubes-windows-tools.exe being distributed directly somewhere? or only thought the RPM?

andrewdavidwong:
> the qubes-windows-tools.exe being distributed directly somewhere? or only thought the RPM?

I'm not sure. Probably only through the RPM. QWT has been unmaintained for quite a long time now.

GWeck:
You could get the QWT iso from dom0, if QWT was installed there. I'l try to describe it.

GWeck:
There's still the question of how to get the three installation files into the Windows 10 VM, since file copy does not yet work to this VM.

I did it via an external Windows system where I stored these files and then copied them via SMB, which works from Windows 10.

Any better idea???

deeplow:
> I did it via an external Windows system where I stored these files and then copied them via SMB, which works from Windows 10.

But in that case you alrady had the `exe` file in the first place. Normal users will only get the RPM (not even the ISO directly).


> Any better idea???

Hum. Not really.

marmarek:
I can upload it somewhere to the ftp.qubes-os.org. How do you verity its integrity (something previously covered by the rpm signature it was shipped inside)? A detached gpg signature? Or maybe somebody knows how to properly sign .exe file using embedded signature?

deeplow:
> I can upload it somewhere to the ftp.qubes-os.org. How do you verity its integrity (something previously covered by the rpm signature it was shipped inside)? A detached gpg signature? Or maybe somebody knows how to properly sign .exe file using embedded signature?

Thanks, @marmarek. That would make it much easier. Unfortunately I don't know how to verify the integrity on windows (and from the little I've seen it's bit of a [convoluted process](https://tails.boum.org/install/win/usb-download/index.en.html). @GWeck how this can be done the proper way (if there is one even)?

andrewdavidwong:
One simpler option is to publish signed hash values, then hash the EXE from within Windows and compare the values.

GWeck:
You could get the exe from the rpm by first installing the rpm via qubes-update-dom0, then getting the iso from /usr/lib/qubes in dom0, and then extracting the exe from this iso. So the integrity of the exe should be covered by the rpm signature. The trustworthiness of the exe is still somewaht lower than that auf the rpm, because the archive manager needed to extract it from the iso shoukld not run in dom0, so you have to transfer the iso to some VM. But that could perhaps be the VM "vault" or so, which should be fairly trustworthy.

The critiacl part is then the transfer to windows, so the integrity should be checked there once again. For me, the checksums are:
MD5: 0C1732F537FEFC457189D860E25525EC
SHA1: BBD83EFD90FA909DA53130D5D8F1F086CBCEAA3D
(calculated with two tools on windows). The file properties in Windows also contain a certificate, which I attach in cer- and pcks#7 format:

[qubes-tools-4.0.1.3.zip](https://github.com/QubesOS/qubes-doc/files/5281662/qubes-tools-4.0.1.3.zip)

Perhaps you could - after checking - just publich these values / files on the website?

GWeck:
Publishing the exe on the website could solve the problem of getting it into the Windows VM. This VM just could download the exe from the web, which should be secure if the checksums and the certificate are also published.

The certificate could be checked directly from the file's properties - the Windows explorer supports that. For checking the checksums, there are several tools available, e.g. the MD5 & SHA Checksum Utility
[https://raylin.wordpress.com/downloads/md5-sha-1-checksum-utility/]

deeplow:
> MD5: 0C1732F537FEFC457189D860E25525EC
> SHA1: BBD83EFD90FA909DA53130D5D8F1F086CBCEAA3D

Aren't these considered broken cryptographic hashes?

There is a built-in "certUtil" which can be used to verify the hashes.

windows

andrewdavidwong:
> > MD5: 0C1732F537FEFC457189D860E25525EC
> > SHA1: BBD83EFD90FA909DA53130D5D8F1F086CBCEAA3D
> 
> Aren't these considered broken cryptographic hashes?

It depends on what you mean by "broken" and for what purpose. They're vulnerable to _collision_ attacks, but that doesn't seem relevant here. Rather what matters here is _preimage_ resistance. As far as I know regarding these two, a preimage attack has been published against MD5, but it's only theoretical.

Having said that, there's not much reason not to use SHA-256 or SHA-512 nowadays, except when backward compatibility with older systems is needed.

GWeck:
I checked now with the tool, which also provides SHA-256 and SHA-512:
SHA-256: `148A2A993F0C746B48FA6C5C9A5D1B504E09A7CFBA3FB931A4DCF86FDA4EC9B1`
SHA-512: `9C461EFA272C65CE66A2114D55C4CD90BED64123D40BED3160418662D41811B8C3080BCD122E0BBE4620AF63CBC7745545EBDAD265F8DD126DF55010D6FE52A4`

@deeplow SHA-256 is the same as that provided by certutil, so there is some hope.

I didn't know that certutil exists, but now I tried it, and, fortunately, it gets the same values. Using the utility requires to specify the hash algorithm in uppercase, otherwise you get the (unexplained) error code `0xd00000bb`.

I changed the QWT installation instructions accordingly, saying that the Xen drivers and `qubes-tools-4.0.1.3.exe` should be downloaded and its checksum verifeid with the certutil utility. Still need the download location and the location for the checksums.

GWeck:
I checked my instructions now again and had some minor changes. The text should now be rather close to reality - I hope so, at least.

external/os-guides/windows/windows-vm.md
```diff
@@ -210,6 +211,82 @@ qvm-prefs win7new qrexec_timeout 300
 
 At that point you should have a functional and stable Windows VM, although without updates, Xen's PV drivers nor Qubes integration (see sections [Windows Update](#windows-update) and [Xen PV drivers and Qubes Windows Tools](#xen-pv-drivers-and-qubes-windows-tools) below). It is a good time to clone the VM again.
 
+### Installing Qubes Windows Tools on Windows 10
```
deeplow, line 214: I think this should be moved to `external/os-guides/windows/windows-tools.md`

external/os-guides/windows/windows-vm.md
```diff
@@ -229,10 +306,14 @@ Installing Xen's PV drivers in the VM will lower its resources usage when using
...
 - the option to install the storage PV driver is disabled by default in Qubes Windows Tools 
 - in case you already had QWT installed without the storage PV driver and you then updated the VM, you may then install the driver from Xen's site (xenvbd.tar).
 
+**Caution:** Installing the version 9.0.0 Xen drivers on Windows 7 (a system without QWT - QWT uninstalled) leads to an unbootable system. The drivers install without error, but after reboot, the system aborts the reboot saying ´Missing driver xenbus.sys´.
+
+- **Windows 10:** The version 9.0.0 Xen drivers have to be installed before installing Qubes Windows Tools. Installing them on a system with QWT installed is likely to produce a system which crashes or has the tools in a non-functional state. Even if the tools were installed and then removed before installing the Xen drivers, they probably will not work as expected.
```
deeplow, line 315: These and the above explanations should also be moved and integrated with `external/os-guides/windows/windows-tools.md`

deeplow:
One key improvement I think would be making this as step-by-step as I can imagine that some users might get lost. This also removes the need for comments about the need for installing the PV drivers before the QWT.

The feedback I had to give would be much more work writing it than actually implementing it so I went ahead and forked you branch. I have no Idea how to PR onto your PR, though. I hope that's ok @GWeck. Feel free to dismiss it if you think otherwise.

I also added the following table on the feature availability breakdown which I borrowed from [here](https://github.com/tabit-pro/qwt-crossbuild/edit/master/README.md)

|             Feature                  |  Windows 7 x64 | Windows 10 x64 |
| ------------------------------------ | :------------: | :------------: |
| Qubes Video Driver                   |        +       |       -        |
| Qubes Network Setup                  |        +       |       +        |
| Private Volume Setup (move profiles) |        +       |       +        |
| File sender/receiver                 |        +       |       +        |
| Clipboard Copy/Paste                 |        +       |       +        |
| Application shortcuts                |        +       |       +        |
| Copy/Edit in Disposible VM           |        +       |       +        |
| Block device                         |        +       |       +        |
| USB device                           |        -       |       -        |
| Audio                                |        -       |       -        |

external/os-guides/windows/windows-vm.md
deeplow: I've addressed this in https://github.com/GWeck/qubes-doc/pull/1

external/os-guides/windows/windows-vm.md
deeplow: I have addressed this in https://github.com/GWeck/qubes-doc/pull/1

GWeck:
As I concentrated on the Windows installation text so far, neglecting the Windows-Tools installation text, the two were getting out of sync. I already had the feeling that there was some work to be done to get them consistent again -and you did it: Thank you!

I checked your texts and think they'll help the user better to get through the maze of Windows and QWT installation.

So I made some minor corrections in your text, hoping I didn't break anything. The main change is that I copied the text on moving user directories to drive `D:` to the QWT installation text, where it may fit better. Perhaps we could remove this text from the Windows installation module - what do you think about that?

Also I am not too sure whether it still makes sense to say that QWT only works for W7. Marketing-wise, it would help to say that W10 is supported, too. Or is that over-optimistic?

deeplow:
> I already had the feeling that there was some work to be done to get them consistent again -and you did it: Thank you!

You're welcome! :smile:

> I checked your texts and think they'll help the user better to get through the maze of Windows and QWT installation.

Thanks, that's the objective :+1: 

> The main change is that I copied the text on moving user directories to drive `D:` to the QWT installation text, where it may fit better. Perhaps we could remove this text from the Windows installation module - what do you think about that?

I agree with with your changes overall. There is just one thing that I didn't fully get from your guide and it has to do with the "move user directories". Does that correspond to the "move user profiles" option in the installer? If so, then the language should be changed to that.

w10_qubes_windows_tools1


> Also I am not too sure whether it still makes sense to say that QWT only works for W7. Marketing-wise, it would help to say that W10 is supported, too. Or is that over-optimistic?

Well, you guide proves it is supported. Not fully, but supported indeed. So I think it's best to say it is. (along with all the disclaimers we added).

---

Also, I took screenshots of the installation process (20 in total). But I'm not sure if I should add them or even what workflow I need to add them. @andrewdavidwong 

Should we add pictures? Here are a few examples


w10_qubes_windows_tools

win10_4

deeplow:
So do we agree on having the download available through the Qubes FTP and simple verification via SHA256? @GWeck

marmarek:
I'll just paste a link :)
https://ftp.qubes-os.org/qubes-windows-tools/

andrewdavidwong:
> Also, I took screenshots of the installation process (20 in total). But I'm not sure if I should add them or even what workflow I need to add them. @andrewdavidwong
> 

The procedure is documented here:

https://www.qubes-os.org/doc/doc-guidelines/#how-to-add-images

GWeck:
@deeplow Fine - we seem to be converging!

Move user directories, which is described for Windows 10, means just to copy/move the `C:\users\<username\Documents` directories manually, not the whole profiles. W10 still expects the profiles (I suppose the `NTUSER.DAT` files and the AppData directories, possibly more) to be on drive `C:`. Especially the Windows updater seems run astray if it does not find the profiles there; it is known to sometimes just delete all user data in this case.

I just entered the ftp url provided by @marmarek into the QWT installation text and changed the cross reference from on of the texts to the other one from absolute to relative.

One more point: View/Edit in dispVM appears in Windows as `Send to --> Edit in DispVM`. This command has no effect in Windows 7, and in Windows 10, a dispVM is created, but does nothing and seems to be waiting for data. There is also a command `Send to --> Edit in VM` which also does nothing in Windows 7; in Windows 10, it lets you select a target VM, starts that if it is not running, but then also does nothing. Could you check that?

deeplow:
> One more point: View/Edit in dispVM appears in Windows as `Send to --> Edit in DispVM`. This command has no effect in Windows 7, and in Windows 10, a dispVM is created, but does nothing and seems to be waiting for data. There is also a command `Send to --> Edit in VM` which also does nothing in Windows 7; in Windows 10, it lets you select a target VM, starts that if it is not running, but then also does nothing. When the W10 VM is shut down, the dispVM is also shut down, and in the target VM the intended application starts, but is empty. Could you check that?

None of these started a vm for me.



> Move user directories, which is described for Windows 10, means just to copy/move the `C:\users\<username\Documents` directories manually, not the whole profiles. W10 still expects the profiles (I suppose the `NTUSER.DAT` files and the AppData directories, possibly more) to be on drive `C:`. Especially the Windows updater seems run astray if it does not find the profiles there; it is known to sometimes just delete all user data in this case.

ok. I didn't do that. But on the installer I did confuse that with the move user profiles and deactivated that on the installer. Don't know if that affects my conclusions.

Also, I installed it as a standalone VM instead of a template



> I just entered the ftp url provided by @marmarek into the QWT installation text and changed the cross reference from on of the texts to the other one from absolute to relative.

thanks

GWeck:
>  None of these started a vm for me.

That just shows how consistent Windows is. :-(

>  ok. I didn't do that. But on the installer I did confuse that with the move user profiles and deactivated that on the installer. Don't know if that affects my conclusions.

Deactivating this option in the installer is good: We shouldn't tempt users to select something which has no use and is dangerous.

GWeck:
@adw @deeplow Now that the text seems to be more or less consistent, what is going to happen now? Has the pull request to be merged, and who will do that? Could you please help me, and tell me if I have to do something?

GWeck:
I have tried what happens in Qubes R4.1 (Version 20200914) with Windows and QWT. So far, the following results are obtained:

- For Windows 7 with QWT installed, the VM from R4.0 can be imported, but not started - error in `libxenlight`.
- For Windows 10 with QWT installed, the VM from R4.0 can be imported and started, but somewhere in the boot process, the VM crashes with the error message `INACCESSIBLE BOOT DEVICE,` an error I already have seen in my earlier tests on Qubes R4.0.
- For Windows 10 without QWT, the VM from R4.0 can be imported and started. The Xen drivers and QWT can be installed, like documented above. However, the VM stays invisible, until `qvm-features gui-emulated` is set to 1. Then, the VM seems to behave quite normally, but is cheating on that behalf: The Qube Manager shows the VM as running (green dot). The Qubes RPC agent is not started, although its start type is set to Automatic, but it can be started manually. Communication with the Qubes does not seem to work, since file copy to/from other VMs does not work, and the Windows VM is killed after its `qrexec_timeout `value, meaning the host does not really see the VM.

It seems, there is still a lot to do for R4.1. :-(

deeplow:
> @adw @deeplow Now that the text seems to be more or less consistent, what is going to happen now? Has the pull request to be merged, and who will do that? Could you please help me, and tell me if I have to do something?

I think you can just merge [this pull request](https://github.com/GWeck/qubes-doc/pull/1) which will merge onto this PR.

I think I may end up adding the images since my time is relatively limited now. But one thing that we're missing and that I don't know how to do is to explain to people can verify the integrity of the DIGESTS download  from the FTP server. I'm  thinking on windows it will be very long and convoluted. @GWeck 

Other than this, I think it can be merged.

andrewdavidwong:
Let me know when you're ready for me to actually merge this PR. I don't want to do it prematurely and mess you up, so I will wait for a clear signal.

GWeck:
@adw I think it can be merged, too.

@deeplow The .DIGEST file just contains the 4 hash values (for MD5, SHA1, SHA256 and SHA512) and a digital signature, created by the key `Qubes OS Release 4 Signing Key`, thus verifying the hash values. If GPG is installed, it can be checked via
`gpg --verify qubes-tools-4.0.1.3.exe.DIGESTS`, if the Qubes keys are signed and set to trustworthy.

For Windows on my test system, GPG does not allow me to sign any keys, throwing the error `no_pubkey`. If the Qubes master key and the release 4 key are signed on a Linux system and then imported to Windows GPG, they are accepted as valid and trustworthy, and the signature verification succeeds, although it writes the error message. On Linux, the verification works without this trouble. I'll try and check what is going wrong here. Until I find out if it is a general problem or just my installation of GPG, I think it is best not to mention the .DIGEST file in the documentation. We can always amend it later.

GWeck:
I checked now GPG on a different Windows maschine and still get the same error. It could perhaps be caused by GPG dropping support for SHA-1??? But then, why does it work on Linux? I did not find this error mentioned anywhere.

andrewdavidwong:
Guys, the whole point of publishing the hash values is so that you don't have to do any PGP verification inside of Windows, remember? All you have to do inside of Windows is hash the file, then compare (using eyeballs) the value with the trusted value outside of the Windows VM. If you were wiling to do PGP verification inside of the Windows VM, then you could have simply verified the EXE directly with a detached PGP signature. The hash values were the solution to avoiding that.

Of course, the clearsigned hash values can (and should) be PGP-verified outside of the Windows VM, but that's already thoroughly covered in our Verifying Signatures doc.

andrewdavidwong:
By the way, if the reason you were inclined to verify the digest file inside of Windows is that you're worried that the file might have been altered by the (potentially compromised) Windows VM itself , then you should understand that a malicious Windows VM could also lie to you about the result of any PGP verification operation performed inside of it. If you don't trust that VM, then you can't trust any outputs generated inside of that VM.

deeplow:
> Guys, the whole point of publishing the hash values is so that you don't have to do any PGP verification inside of Windows, remember?

Yup. Makes sense. Then I guess from my side it's ready to be merged. 

Also, another user made a guide on this as well and posted [on the forum](https://qubes-os.discourse.group/t/documentation-windows-10-installation-on-qubes-r4-0/990). Maybe they have something to contribute, but since this is almost ready I'm not sure if there's any extra contribution if that should go on this PR or another one.

@andrewdavidwong is this PR the only blocker for moving the external documentations onto the community docs (https://github.com/QubesOS/qubes-issues/issues/4693)?

andrewdavidwong:
> is this PR the only blocker for moving the external documentations onto the community docs (QubesOS/qubes-issues#4693)?

No, there's also https://github.com/QubesOS/qubes-doc/pull/989.

GWeck:
That's o.k. from me, too - ready to be merged.

The funny behaviour of gpg on my Windows machine had two causes: First, there were some rather old keys in my keyrings, probably from GPG version 1, and then and then the fact that SHA-1 is no longer supported on Windows, so that trying to sign something with a key using this algorithm does no longer function. On Fedora-32, it still works.

Now I looked into the Qubes documentation text about verifying ISOs and signatures, and there is everything explained as needed, so we shouldn't repeat that here. No need to add anything else. (Should have read the documentation before - RTFM - shame on me...)

andrewdavidwong:
No worries. Thank you both for the fantastic work here! I'm sure many users will benefit from this. :slightly_smiling_face:

GWeck:
Thanks a lot! Butt ...

The final version was not was is merged no, but the 2 files under [https://github.com/GWeck/qubes-doc/pull/1](https://github.com/GWeck/qubes-doc/pull/1)

Could you change that?

deeplow:
> Thanks a lot! But ...
> 
> The final version was not was is merged no, but the 2 files under [GWeck#1](https://github.com/GWeck/qubes-doc/pull/1)
> 
> Could you change that?

Damn, I through those had been merged already. I think the best thing to do now is merge that https://github.com/GWeck/qubes-doc/pull/1 into your branch. And then open a new PR on qubes-doc and then merge it. Should be quick

andrewdavidwong:
> The final version was not was is merged no, but the 2 files under GWeck#1

Ah, that exactly why I wanted to wait until you both were completely ready...

> I think the best thing to do now @GWeck is merge that GWeck#1 into your branch. And then open a new PR on qubes-doc and then merge it. Should be quick.

Yes, please do this.

It's not clear to me exactly what changes you still want, and me trying to guess will only compound the confusion.

GWeck:
I merged the changes and created a new pull request (#1054) which contains the current (hopefully final) state. I hope I did not break anything - I am not so used to github.

GWeck:
I checked now R4.1 with Windows 7, and got the following results:

- For Windows 7 without QWT, the VM from R4.0 can be imported and started. The QWT can be installed, like documented for R4.0, even with the option to move the user profiles. The VM seems to behave quite normally, but is cheating on that behalf: The Qube Manager shows the VM as running (green dot). Communication with Qubes does seem to work only partially, since file copy to/from other VMs does not work, but the VM stays running. - The screen resolution stays fixed at 1920 x 1024 (my screen is 1920 x 1200), and the only other option shown is 800 x 600. Trying to chnage the resolution renders the VM inoperative - it cannot even be killed, and Qubes has to be restarted to get access to it again.

I am going to repeat these tests on the Alpha version of R4.1 published lately by @marmarek and publish the results on the new Testing R4.1 category.

PR #1047, opened 2020-10-02
[GSoD] Add Installation Troubleshooting, move files

PROTechThor:
This is the first section of the consolidated troubleshooting guide. 
I've added a new file for the Installation troubleshoots. 
In addition, I moved the Troubleshooting files from "External documentation" to "User documentation"

PR #1048, opened 2020-10-02
[GSoD] UEFI Troubleshooting

PROTechThor:
I've added more information and modified some heading. 
Some content was gotten from https://github.com/QubesOS/qubes-doc/pull/1036
I also moved the UEFI Troubleshooting guide from the "Advanced Configuration" directory to "Troubleshooting".


marmarek (approved):

andrewdavidwong:
We should really discuss such reorganization of the docs first. What is the rationale for moving these back out of external to the core docs? The commitment to maintain them as core docs goes on in perpetuity (unless we move them out again someday), so is not to be taken lightly.

PROTechThor:
> We should really discuss such reorganization of the docs first. What is the rationale for moving these back out of external to the core docs? The commitment to maintain them as core docs goes on in perpetuity (unless we move them out again someday), so is not to be taken lightly.

@andrewdavidwong I moved them just so they can look more official. @marmarek agrees that some core things like booting, networking, updating should be kept in the User Documentation. But then, it may be confusing to have the Troubleshooting guide split in two sections - User and External. So maybe it's best to keep it at External. I'm torn about it.

andrewdavidwong:
I don't think that simply *looking* more official is a good reason. There is a very real difference between official and unofficial docs. PRs against official docs have to go through a review process to ensure accuracy before being merged, which takes up valuable time from the team. I'd also like to have higher quality standards for the core docs (style and mechanics, formatting), which takes up a lot of time. It's also important to periodically review and update this documentation, especially when a new Qubes release comes out. The more documentation there is relative to maintainers, the harder all of this will be.

We have to think about who is going to spend the time to maintain these docs in the years ahead. It can be very tempting for doc contributors to add something to the pile when they get to simply walk away afterward and not have to worry about what happens to the pile in the future. Don't get me wrong; I love doc contributions. But the quality and relevance of contributions is highly variable, and most contributions seem to be for adding new content rather than improving or maintaining existing content. Given the very real and significant maintenance burden, I think there should be a presumption *against* any docs being official unless strong reasons can be demonstrated for why they should be official.

For example, docs that cover code produced by the Qubes OS Project should generally be official. We should document our own code. However, docs that cover code produced by other projects or that cover hardware generally should not. We should not get into the business of documenting other people's stuff when we can barely document our own stuff. Let's get our own house in order (and make sure we can keep it that way) first. Another consideration is whether the documentation is security-sensitive. In other words, if this documentation were malicious, how likely is it that users could compromise themselves, and how severely? For example, if the documentation doesn't involve doing anything in dom0, then it's less sensitive. If it also doesn't involve doing anything in templates, it's even less sensitive, and so forth.

I would also like to avoid a situation where docs get moved back and between official and external without a principled rationale as to why. Before we reverse a decision that has been made, we should understand why the original decision was made, and we should be able to explain what has changed since then such that reversing the decision now makes sense (see [Chesterton's fence](https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence)).

marmarek:
Troubleshooting of common issues should be part of the official documentation. This should include topics like networking, starting VMs, but also known hardware-related issues. While for the online version it's just a matter of going to a different place, it makes a huge difference for offline documentation version (like - local copy of qubes-doc repo). If the thing you have issue with prevents you from accessing network (for example sys-net doesn't start), then having troubleshooting guide locally is critically helpful.

andrewdavidwong:
> Troubleshooting of common issues should be part of the official documentation. This should include topics like networking, starting VMs, but also known hardware-related issues.

Ok, then I suppose the remaining question is where we draw the line between common and uncommon. Most likely, this will have to be decided on a case-by-case basis. Do you think that everything being moved into the official docs in this PR qualifies as "common"?

> While for the online version it's just a matter of going to a different place, it makes a huge difference for offline documentation version (like - local copy of qubes-doc repo). If the thing you have issue with prevents you from accessing network (for example sys-net doesn't start), then having troubleshooting guide locally is critically helpful.

The unofficial docs can also be downloaded as a repo for offline use. They're just Markdown files in a git repo, same as our docs.

-----

It would be nice if we could move these docs in a way that preserves their git history (as viewable within GitHub). I know we didn't do that when they were originally moved to external (because we didn't know how), but would any of [these](https://github.com/QubesOS/qubes-issues/issues/4693#issuecomment-697315837) methods work?

PROTechThor:
@andrewdavidwong @marmarek 
This is an insightful discussion! But I'm stuck in limbo until we have a final verdict on what to do. Both options have their pros and cons, and I can't decide if any outweighs the others. 

The easier option is to keep everything in "External", but this may be at the detriment of users not finding official troubleshoots when they need them. 

If we want some core troubleshooting in the Official documentation, then some will have to be in the External community docs. This will call for two separate troubleshooting sections, right? Won't this be confusing? In fact, one of the issues raised in my [GSoD project explanation](https://www.qubes-os.org/gsod/#consolidate-troubleshooting-guides) was the fact that "Troubleshooting guides are scattered across many pages". I've been really trying to ensure that everything in one place. 

In case we decide to separate the troubleshooting guide into Core and External, would the following categorization work?:

**Core**

- Installation troubleshooting
- UEFI
- Verifying Signatures
- Suspend/resume
- VM
- Disk
- PCI
- USB
- GUI
- Multiboot

**External**

- Performance
- Media
- VPN
- Firewall
- Wireless
- Nvidia
- AMD
- Intel Integrated Graphics
- Lenovo
- Apple Macbook
- Sony Vaio
- Other hardware issues
- Application
- Update
- Qubes-Whonix
- Tails

marmarek:
I'm tempted to say all the troubleshooting guides applicable to the standard installation (so - no custom, non-default applications) with exception to very specific to uncommon/unsupported hardware should go into core part. What does it mean "uncommon/unsupported hardware" is still an open question though. I'd say installing nvidia proprietary driver is out of scope, and also Sony Vaio (that's because this page is extremely old and probably doesn't work on any machine meeting current qubes requirements), but others should go in. Alternative could be moving all docs about specific hardware to external documentation. @andrewdavidwong what do you think?

andrewdavidwong:
> If we want some core troubleshooting in the Official documentation, then some will have to be in the External community docs. This will call for two separate troubleshooting sections, right? Won't this be confusing? In fact, one of the issues raised in my GSoD project explanation was the fact that "Troubleshooting guides are scattered across many pages". I've been really trying to ensure that everything is in one place.

When you say "ensure that everything is in one place," you have in mind putting everything on one _page_ (as opposed to the current situation where it's "scattered across many pages"), right? But this PR keeps all of the pages separate and moves/renames them. This seems somewhat contradictory and leaves me confused about what's happening here.

> I'm tempted to say all the troubleshooting guides applicable to the standard installation (so - no custom, non-default applications) with exception to very specific to uncommon/unsupported hardware should go into core part. What does it mean "uncommon/unsupported hardware" is still an open question though. I'd say installing nvidia proprietary driver is out of scope, and also Sony Vaio (that's because this page is extremely old and probably doesn't work on any machine meeting current qubes requirements), but others should go in.

I agree.

> Alternative could be moving all docs about specific hardware to external documentation. 

One consideration is that there is a *lot* of different hardware out there, so a lot of "specific hardware" documentation could accumulate over time. On the other hand, certain models are definitely more common for Qubes, e.g., Thinkpads, especially X230. Qubes Hardware Certification should probably have some weight here. So, I'm more inclined toward your initial suggestion (e.g., Thinkpad docs in core, Sony Vaio in external). Perhaps even more importantly, it can be difficult or impossible for us to properly review hardware troubleshooting docs for "uncommon" hardware when we have no access to that hardware ourselves.

PROTechThor:
@andrewdavidwong @marmarek Like this?

**Core**

- Installation troubleshooting
- UEFI
- Verifying Signatures
- Suspend/resume
-  VM
- Disk
- PCI
- USB
- GUI
- Multiboot
- Performance
- Media
- VPN
- Firewall
- Wireless
- Update
- Lenovo
- Qubes-Whonix
- Nvidia

**External**

- Tails
- AMD
- Intel Integrated Graphics
- Apple Macbook
- Sony Vaio
- Other hardware issues
- Application

andrewdavidwong:
> Verifying Signatures

I think the "Verifying Signatures" FAQ makes sense where it is right now (i.e., on that same page). It's not part of using Qubes itself. It's mainly about basic GPG usage errors.

> Qubes-Whonix

This is on the Whonix wiki. I do not recommend trying to duplicate it into the Qubes documentation. (That has been attempted before and never works out well, not to mention that doc duplication is a bad idea in general.)

> AMD
> Intel Integrated Graphics

These might be considered common enough. I leave it up to @marmarek's judgment.

> Application

What does this refer to?

andrewdavidwong:
> When you say "ensure that everything is in one place," you have in mind putting everything on one page (as opposed to the current situation where it's "scattered across many pages"), right? But this PR keeps all of the pages separate and moves/renames them. This seems somewhat contradictory and leaves me confused about what's happening here.

I'm still unclear on this point, @PROTechThor.

PROTechThor:
> > Application
> 
> What does this refer to?

Issues relating to apps installed on Qubes, such as Firefox and Libreoffice:

- Can't launch any applications after starting VM
- Fullscreen firefox is frozen
- "qubes-usb-proxy not installed in the VM" error
- LibreOffice opens as a tiny window
- Constant firefox crashes



> When you say "ensure that everything is in one place," you have in mind putting everything on one page (as opposed to the current situation where it's "scattered across many pages"), right? But this PR keeps all of the pages separate and moves/renames them. This seems somewhat contradictory and leaves me confused about what's happening here.

By "everything in one place", I meant there should be one troubleshooting directory which contains every page. There will still be different pages for different topics. Now that you've asked me to clarify this, I'm not even sure if I'm making any sense @andrewdavidwong

andrewdavidwong:
> By "everything in one place", I meant there should be one troubleshooting directory which contains every page. There will still be different pages for different topics. Now that you've asked me to clarify this, I'm not even sure if I'm making any sense

Well, the directories are mainly just a way to organize the source files. Having directories that mirror the categories displayed on the website makes them easier to find, but in theory it could be different. For example, every `.md` could be in a single flat directory. This is because the permalinks are declared in the YAML headers at the top of each page. (We do have some custom stuff that checks the parent directory, but that's just our choice. Jekyll doesn't care.)

I fear that this may be getting into bikeshed territory. For the purposes of GSoC, I think it makes more sense to focus first on improving the content within documents rather than reorganizing the documents.

PROTechThor:
> Well, the directories are mainly just a way to organize the source files. Having directories that mirror the categories displayed on the website makes them easier to find, but in theory it could be different. For example, every `.md` could be in a single flat directory. This is because the permalinks are declared in the YAML headers at the top of each page. (We do have some custom stuff that checks the parent directory, but that's just our choice. Jekyll doesn't care.)

I see. Thanks for clarifying :)

andrewdavidwong:
Looks like we're almost ready to merge this, but I never received a response to [this question](https://github.com/QubesOS/qubes-doc/pull/1048#issuecomment-703163560), so I'd like to repeat it:

It would be nice if we could move these docs in a way that preserves their git history (as viewable within GitHub). I know we didn't do that when they were originally moved to external (because we didn't know how), but would any of [these](https://github.com/QubesOS/qubes-issues/issues/4693#issuecomment-697315837) methods work?

PROTechThor:
@andrewdavidwong 

> It would be nice if we could move these docs in a way that preserves their git history (as viewable within GitHub). I know we didn't do that when they were originally moved to external (because we didn't know how), but would any of these methods work?

I can't answer this question, because I don't really understand the procedure and results expected. Apart from the basic clone-commit-push commands, I struggle with Git. 
If you can confirm that the methods work, I'll try to implement it to the best of my ability.

andrewdavidwong:
My mistake. I forgot that we're only renaming files within the same repo, not moving files across repos here. So, the steps I linked earlier aren't necessary. However, in order to preserve history (as viewable in GitHub), it's still necessary to make sure that each rename commit is separate from a commit that changes the contents of the file too much. You've already done that with almost all of the files here except for the UEFI Troubleshooting doc. If it's not too much trouble, making the file rename a separate commit from your content changes to that doc would be good.

andrewdavidwong:
@PROTechThor, I have just merged #1047, so this has some merge conflicts now (as we expected and discussed earlier).

PROTechThor:
@andrewdavidwong All good now?

andrewdavidwong:
> @andrewdavidwong All good now?

I think so.

PR #1051, opened 2020-10-04
Convert Introduction from HTML to Markdown

andrewdavidwong:
See discussion on https://github.com/QubesOS/qubes-issues/issues/1019.


andrewdavidwong:
@marmarek, is this what you had in mind?

marmarek:
I think that isn't all, there is also 'intro' layout. Anyway @maiska is preparing updated PR with her work right now, I'd simply wait instead of picking pieces that will make conflict resolution even harder.

andrewdavidwong:
> I think that isn't all, there is also 'intro' layout.

Oh, that layout can simply be deleted if it's no longer going to be used, and this intro page can be switched to the standard layout.

> Anyway @maiska is preparing updated PR with her work right now, I'd simply wait instead of picking pieces that will make conflict resolution even harder.

Oh, your [comment](https://github.com/QubesOS/qubes-issues/issues/1019#issuecomment-703314807) sounded like you wanted this change to be made.

PR #1053, opened 2020-10-10
[GSoD] Suspend/resume troubleshooting

PROTechThor:
I renamed "Wireless Troubleshooting" to "Suspend/Resume Troubleshooting" and added more content to it.

PR #1055, opened 2020-10-10
[GSoD] VM Troubleshooting

PROTechThor:


user/troubleshooting/vm-troubleshooting.md
```diff
@@ -0,0 +1,70 @@
...
+layout: doc
+title: Suspend/Resume Troubleshooting
+permalink: /doc/vm-troubleshooting/
+redirect_from:
+- /doc/remove-vm-manually/
+---
+
+# VM troubleshooting #
+
+## VM Kernel troubleshooting ##
```
marmarek, line 11: I think this needs clearer indication it's about non-default kernel choice described in https://www.qubes-os.org/doc/managing-vm-kernel/#using-kernel-installed-in-the-vm. Otherwise it sounds like it applies to any issues with a kernel in VM, which is not true.

user/troubleshooting/vm-troubleshooting.md
```diff
@@ -0,0 +1,70 @@
...
+You need to call it just after starting the VM (until `GRUB_TIMEOUT` expires); for example, in a separate dom0 terminal window.
+
+In any case you can later access the VM's logs (especially the VM console log `/var/log/xen/console/guest-VMNAME.log`).
+
+You can always set the kernel back to some dom0-provided value to fix a VM kernel installation.
+
+## Qubes starts, but no VMs load ##
+
+This issue may occur if a dom0 update is interrupted halfway through and/or a hard power off is done without shutting down Qubes, which results in files getting corrupted. 
+In this case, the best fix is to reinstall Qubes and restore your files from a backup.
```
marmarek, line 23: This is nuclear option, in many cases there are less intrusive methods, depending on why exactly VMs do not start. Better option is to recommend trying to start some specific VM to see exact failure message and direct further steps based on that.

user/troubleshooting/vm-troubleshooting.md
```diff
@@ -0,0 +1,70 @@
...
+One solution is to add a dummy interface to allow the package to install correctly:
+
+    ip link add d0 type dummy
+    ip addr add 192.168.0.1/24 dev d0
+    ip link set d0 up
+
+## "Cannot connect to qrexec agent" error ##
+
+If you face this error when starting a VM, it may be due to too little initial memory. 
+A solution is to increase the initial memory from 200MB to 400MB by navigating to VM settings » Advanced » Initial memory.
```
marmarek, line 69: Good idea is also to look into `/var/log/xen/console/guest-VMNAME.log`. There may be other reasons than too low memory - like VM kernel crash on startup, or some corrupted files.

marmarek (changes requested):
Besides comments about the content itself, I see you combined this with changes already included in https://github.com/QubesOS/qubes-doc/pull/1053. Either split them, or make a single PR about both things.

doc.md
```diff
@@ -135,13 +135,17 @@ Core documentation for Qubes users.
  * [Making Any File Persistent Using `bind-dirs`](/doc/bind-dirs/)
  * [GUI Configuration](/doc/gui-configuration/)
  * [Resizing Disk Images](/doc/resize-disk-image/)
- * [Troubleshooting UEFI](/doc/uefi-troubleshooting/)
+ * [Troubleshooting UEFI](/doc/uefi-troubleshooting/)
```
marmarek, line 138: Spurious space at the end.

marmarek (changes requested):

marmarek (approved):
Content seems fine now.

PROTechThor:
@andrewdavidwong I've fixed the merge conflict here :)

PR #1058, opened 2020-10-13
add warning about where public keys live in TB 78

mfc:


andrewdavidwong:
Yeah, it's worth repeating at the start, since a lot of people seem to be missing the similar statement a few paragraphs down.

PR #1071, opened 2020-11-01
next steps: adding an icon

deeplow:
The next step after installing an application is generally adding an icon to the AppVM, so it may be good to refer this as a next step here. 

Possibly what originated confusion with [this user on the forum](https://qubes-os.discourse.group/t/issues-installing-transmission/1315)

### DIFF
--- a/user/common-tasks/software-update-domu.md
+++ b/user/common-tasks/software-update-domu.md
@@ -26,6 +26,7 @@ To permanently install new software in a TemplateVM:
  4. Shut down the TemplateVM.
  5. Restart all [TemplateBasedVMs] based on the TemplateVM.
 
+After this you may want to add an icon of this application to the start menu of an AppVM based on this template (like your `personal` qube). See how to do this on [Managing shortcuts to applications in AppVMs](https://www.qubes-os.org/doc/managing-appvm-shortcuts/)
 
 ## Updating software in TemplateVMs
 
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
This should happen automatically. No manual step should be necessary. If there is some bug preventing this from working correctly, please file a bug report instead.

marmarek:
I think @deeplow meant going to appropriate VM settings "Applications" tab and actually selecting the application to be included in the menu. That doesn't happen automatically when you install new application,

deeplow:
> I think @deeplow meant going to appropriate VM settings "Applications" tab and actually selecting the application to be included in the menu. That doesn't happen automatically when you install new application,

Thanks. This is it @marmarek. Sorry for the confusion @andrewdavidwong

andrewdavidwong:
I think that this way of writing it is a bit too confusing. It makes it sound like some more complex manual step than what it is. Will close from separate commit.

andrewdavidwong:
Also, the "Managing shortcuts to applications in AppVMs" page is really about troubleshooting this process. It normally shouldn't be required, so it doesn't really belong under "Common Tasks."

deeplow:
> I think that this way of writing it is a bit too confusing. It makes it sound like some more complex manual step than what it is. Will close from separate commit.

Ok. No problem

PR #1073, opened 2020-11-03
Amend installation guide and troubleshooting

PROTechThor:
As discussed here: https://github.com/QubesOS/qubes-issues/issues/6179


ethanchinaski (approved):
as a new user this would help a lot, it points out the options a user have. understandable and precise.

PR #1076, opened 2020-11-06
Windows users can't verify ISO integrity. Warn them of this

deeplow:
The [installation instructions](https://www.qubes-os.org/doc/installation-guide/) warn people to verify the signatures and also include instructions

Some windows users might get really lost if they are reading this. 

The text is just a draft and probably needs something else telling them how to proceed if they want to verify the signature.

----

Instead of this PR another solution could be breaking down the ISO download and USB burning into a Windows and Linux Section. Then there could be a message there telling people they cannot verify the integrity on windows, currently.

### DIFF
--- a/project-security/verifying-signatures.md
+++ b/project-security/verifying-signatures.md
@@ -50,7 +50,9 @@ There are three basic steps in this process:
  2. [Get the Release Signing Key][RSK]
  3. [Verify your Qubes ISO][signature file]
 
-If you run into any problems, please consult the [Troubleshooting FAQ] below.
+If you run into any problems, please consult the [Troubleshooting FAQ] below. 
+
+If you are using windows these instructions do not apply.
 
 ### 1. Get the Qubes Master Signing Key and verify its authenticity
 


deeplow:
Motivation for this comes from here: https://qubes-os.discourse.group/t/iso-verification-instructions-cli-commands-arent-recognized-by-my-non-qubes-cli/1373

andrewdavidwong:
I don't think it's true to say that there's no way to verify PGP signatures on Windows. Example:

https://support.torproject.org/tbb/how-to-verify-signature/

andrewdavidwong:
Actually, I think I can do this today.

andrewdavidwong:
Since I can't test anything on Windows or Mac, I can only add so much. I'll open an issue for further improvement.

deeplow:
> I don't think it's true to say that there's no way to verify PGP signatures on Windows. Example:
> 
> https://support.torproject.org/tbb/how-to-verify-signature/
> 
> It would be great if someone could help us document the methods for Windows and Mac.

For sure! It's possible. Upon rereading my comment that's totally the impression I've passed. What I meant that it isn't currently documented. Let's see if some other folks can help on that issue :)

PR #1086, opened 2020-11-14
[GSoD] Multiboot Troubleshooting

PROTechThor:
There isn't enough content here to justify moving it to a new page.


andrewdavidwong:
Puzzling that the link is only being added, not moved. I thought it was already in the ToC before.
I'm guessing you removed this ToC link in a previous PR and are adding it back now as part of the reorganization.

PR #1088, opened 2020-11-16
Improve "copying" documentation

andrewdavidwong:


fepitre:
@andrewdavidwong if you experiment weird issues please don't hesitate to ping me or Marek. As we are currently moving to our new CI infra there is few new build issues that we have/are fixed/fixing.

andrewdavidwong:
@fepitre: Thanks! Only weird thing I've noticed so far is [here](https://gitlab.com/QubesOS/qubes-doc/-/jobs/851437615), where it says that the HTML failures are on the Research page, but they were actually on the Split GPG page.

fepitre:
Thank you for your feedback. I cc @marmarek too for having it mind.

PR #1084, opened 2020-11-14
Create split-ssh-keepassxc

invalid-error:
https://qubes-os.discourse.group/t/best-split-ssh-setup-guide/385/19?u=whoami

PR #1089, opened 2020-11-17
Split SSH Guide

santorihelix:


deeplow:
Hi @andrewdavidwong! We'd like to have your opinion on whether or not this belongs in official docs. This is a follow-up of https://github.com/QubesOS/qubes-doc/pull/1084 and a (private) discussion we've been having on the forum.

It was originally aimed at being just about qubes-ssh for the specific use-case of doing it with KeePassXC, but we (@invalid-error @santorihelix  and [Phil](https://qubes-os.discourse.group/u/phl) are now aiming at making this a more general split-ssh guide.

Considering we were thinking of following a similar structure [Split GPG](https://www.qubes-os.org/doc/split-gpg/) (which is on the official docs), would you consider this contribution for the official docs? Or do you think it still belongs on the community?

**Note:** I think the reason why split-SSH was on the external documentation was because it was a link to a blog post by @kushaldas.

andrewdavidwong:
IMO, the deciding factor is whether the documentation is documenting software that is officially written and maintained by the Qubes OS Project or not. The goal is to keep the core docs maintainable and high-quality by limiting them to the software output by the Qubes OS Project. (In other words, we want to document all of the software we put out into the world, but we don't want to commit to documenting or maintaining documentation for anything else, because we simply can't afford to do that.) Since the Split GPG software is officially written and maintained by the Qubes OS Project, its documentation belongs in the core docs. Since that does not appear to be the case here, I think this belongs in the community documentation. In addition, just because the Qubes OS Project has officially written and maintains some flexible framework, such as `qrexec`, I don't think it makes sense to include every tutorial that says "here's how to do something cool with `qrexec`" in the core docs. Such tutorials generally also belong in the community documentation.

deeplow:
Thanks for taking the time for the detailed explanation. Community docs it is!

PR #1096, opened 2020-11-23
fix indentation

deeplow:
I think the `[ref]:` part should be under the ref style bullet point


andrewdavidwong:
Well, an actual reference-style link would should not be indented at all. This is just a way to demonstrating how it would look in the source document.

However, I've recently changed my mind about this. Non-reference-style links can be more appropriate in certain contexts, and they tend to be easier to write and edit. Reference-style links are especially useful when links get repeated multiple times. The main thing is that it's nice to consistently use the same link style throughout a single document, but this is not required for things to work.

I'm going to remove this guideline now.

PR #1099, opened 2020-11-29
Explicit commands for installing software

deeplow:
After coming across [this thread](https://qubes-os.discourse.group/t/untrusted-application-launcher/1692) and taking a look at the this doc page, I realized there were no explicit instructions to install software on fedora or debian.

It is arguable, Qubes is a generic system, but Fedora and Debian are the default and installing software is a key thing. So I think it could make a service to new users to tell them explicitly the install commands.

### DIFF
--- a/user/common-tasks/software-update-domu.md
+++ b/user/common-tasks/software-update-domu.md
@@ -22,9 +22,7 @@ To permanently install new software in a TemplateVM:
 
  1. Start the TemplateVM.
  2. Start either a terminal (e.g. `gnome-terminal`) or a dedicated software management application, such as `gpk-application`.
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
> After coming across [this thread](https://qubes-os.discourse.group/t/untrusted-application-launcher/1692) and taking a look at the this doc page, I realized there were no explicit instructions to install software on fedora or debian.

I'm not convinced that these instructions would have made any difference in this particular case. If you look at the instructions that the user claims to have followed, they are exactly these (`sudo dnf install libreoffice`).

> It is arguable, Qubes is a generic system, but Fedora and Debian are the default and installing software is a key thing. So I think it could make a service to new users to tell them explicitly the install commands.

It's generally not a good idea to try to document other projects' software, but this is minor enough that it should be ok.

PR #1103, opened 2020-12-14
Mention OpenGL's software fallback in FAQ

AlxHnr:
Closes https://github.com/QubesOS/qubes-issues/issues/6241

### DIFF
--- a/introduction/faq.md
+++ b/introduction/faq.md
@@ -322,12 +322,13 @@ If it seems like a feature that we can and should enable, please [let us know][r
 
 Absolutely.
 
-### Can I run applications, like games, which require 3D support?
+### Can I run applications, like games, which require hardware acceleration?
 
 Those won’t fly. 
-We do not provide OpenGL virtualization for Qubes. 
+We do not provide GPU virtualization for Qubes. 
 This is mostly a security decision, as implementing such a feature would most likely introduce a great deal of complexity into the GUI virtualization infrastructure. 
-However, Qubes does allow for the use of accelerated graphics (OpenGL) in Dom0’s Window Manager, so all the fancy desktop effects should still work.
+However, Qubes does allow for the use of accelerated graphics (e.g. OpenGL) in Dom0’s Window Manager, so all the fancy desktop effects should still work.
+AppVMs use a software-only (CPU-based) implementation of OpenGL, which may be good enough for basic games and applications.
 
 For further discussion about the potential for GPU passthrough on Xen/Qubes, please see the following threads:
 
--- a/introduction/faq.md
+++ b/introduction/faq.md
@@ -327,7 +327,7 @@ Absolutely.
 Those won’t fly. 
 We do not provide GPU virtualization for Qubes. 
 This is mostly a security decision, as implementing such a feature would most likely introduce a great deal of complexity into the GUI virtualization infrastructure. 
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
The added note seems to contradict the text immediately preceding it. At least, I as a user am not sure how to understand this. Does OpenGL work or not? The answer (after this change) seems to be that the non-software version of OpenGL doesn't work, while the software version of OpenGL does. As a mere user, I don't understand exactly what OpenGL is, but I thought it *was* software, so what exactly is the non-software version that doesn't work? More importantly, what is the practical upshot of all of this for me? Which applications can I expect to work? Maybe some examples would help.

AlxHnr:
> The added note seems to contradict the text immediately preceding it.

Because the preceding text is wrong and implies there is no OpenGL support at all. I reworded the preceding to get rid of this contradiction.

OpenGL is indeed available and programs build with OpenGL will start and run in AppVMs. Software like [Krita (painting program) ](https://krita.org/en) uses OpenGL by default for its canvas. You can even run `glxgears` in your terminal (basic OpenGL demo).

Of course there is still a lack of _GPU virtualization_, causing some games to run at 9 fps. But the OpenGL API itself is available and works.

andrewdavidwong:
Looks much better, but now some users will think it means that no games or similar software can run at all, but your point is that actually some can. Perhaps it's worth integrating that information?

marmarek (approved):
```
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA256

Approved as of 2fa0fad4bccbcd0f6ffa4d085a11bc1290e70f2d
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCAAdFiEEnYWYU8Ll1fIX9iss8yiUvpaEk4oFAl/00pcACgkQ8yiUvpaE
k4ptZBAAu58feONRzpRNBKojjs6ltgypaDBTN974oyBYM26QlILiV2XySaoFWo3u
3PQug8YDRv1Kopx40GQ52z0XcyZHc+Yb8CjagnDhrCrpJ5Ozmzhh4VIthrGy0/ig
TIK22woQO/VhXzcqpuUPFAmdKLUKc7T0cnLCYOS1zF6qDrcADZ9OPYSAJpyOya/C
m+2PjmnMR4i3pNYs/eONCa/Lv2mqYBEvuxgbQz49/4jblYLjixo+AeXbBqmHFsx6
2ZvO61u2BWB0sTtbLAT4uxRIwqi0X76cGAVhoX9JhNBUD9wFq4hE3L1nTInaqVhM
WJkNR5sxdm8kAobXfb3GPDD6ENeYsvZ7fFb/ljtfv/tGmTzbI5hG/3AJRrYkLBzH
pLTMlKp3UWSSra824ZCUgtco+wdJ5cknDBOifMthO9vGxmA+fYwo8w27TFL8CL1Z
+F9+cttJ2+NsvEmAE6Ti3uu1qFeAS8iHcM2Ec7ilEEVyJA74N+SN3ojcPDEjfxQk
IN296jqHC+ymm814zs+eAB+V9H/BtbtSKQJ8oEeWI0yOCvl8OwVJp6eglgd9Nph9
qnUDbDaAMxOxTyNZ9SI7sskoSL58XYTRzziV7rJseSNcm67Lgeb8UQf80YOop75I
IYiLe4WO8yPy6a4UwzaiyROtR6+x/xZmC0KtNMwnq+C4FjghDIc=
=PF7z
-----END PGP SIGNATURE-----
```

PR #1105, opened 2020-12-19
Fix redirects to 404.

ccdunder:
Fix redirect to 404.

### DIFF
--- a/external/os-guides/windows/windows-tools.md
+++ b/external/os-guides/windows/windows-tools.md
@@ -1,7 +1,7 @@
 ---
 layout: doc
 title: Qubes Windows Tools
-redirect_to: https://github.com/Qubes-Community/Contents/blob/master/docs/os/windows-tools.md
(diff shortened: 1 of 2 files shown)


ccdunder:
Hi, It seems like there might have been a batch rename that broke some links, but this is just speculation since I'm new to the project. I'd be happy to look further into this if a maintainer think it's worth it. Tagging @andrewdavidwong since they seem to have the most expertise in this area. Thanks in advance for the review!
- Chris

andrewdavidwong:
> Hi, It seems like there might have been a batch rename that broke some links

Yes. Unfortunately, it was a very complicated sequences of `sed` commands, due to the different natures of the pages in both repos and different path structures.

> I'd be happy to look further into this if a maintainer think it's worth it.

That would be helpful. Thank you!

There should be no 404s anywhere on the Qubes website, so if you find any more, they are unintentional and will be fixed.

PR #1108, opened 2021-01-06
fix broken links to github

deeplow:


deeplow:
And it seems like the `qrexec-policy-agent.rst` no longer exists. Fixing this may be outside of the scope of this PR.

PR #1119, opened 2021-03-08
mention qubes updater

deeplow:
Although mentioned in the [getting started](https://www.qubes-os.org/doc/getting-started/#gui-and-command-line-tools), I feel like a mention of Qubes Update should absolutely also be in this page. The previous explanation was too technical / dense for people just simply wanting to update the system.

@andrewdavidwong I hope you don't mind the low-effort contribution (already spent too much time one the forum today).

This needs to be further developed (at least adding an image). If you agree with the change, could you complete it?

### DIFF
--- a/user/common-tasks/updating-qubes-os.md
+++ b/user/common-tasks/updating-qubes-os.md
@@ -11,6 +11,14 @@ Updating Qubes OS
 If you're instead looking to upgrade from your current version of Qubes OS to a newer version, see the [Upgrade Guides].*
 
 It is very important to keep your Qubes OS system up-to-date to ensure you have the latest [security] updates, as well as the latest non-security enhancements and bug fixes.
+
+Qubes update (on tray menu) » Launch updater » Next (leave all selected).
+
+[add image]
+
+Updating the system for advanced users
+--------------------------------------
+
 Fully updating your Qubes OS system means updating:
 
  - [Dom0]


andrewdavidwong:
I don't have any way to take a good screenshot for this right now, but maybe @hackerncoder does.

I disagree with moving the existing content into a "Updating the system for advanced users" subsection. The Qubes Update tool is the recommended way to accomplish these steps, not something separate.

deeplow:
> I disagree with moving the existing content into an "Updating the system for advanced users" subsection. The Qubes Update tool is the recommended way to accomplish these steps, not something separate.

My reasoning for the proposal was to make the "Qubes updater" very visibly mentioned on this page (which I think currently is not mentioned). They way, new users only really need to look at a small section at the top of this docs page. Curious users can look at what goes on behind the scenes & more advanced stuff. That was just my goal with the move. But it may be called something else.

andrewdavidwong:
> My reasoning for the proposal was to make the "Qubes Update" very visibly mentioned on this page (which I think currently isn't, if I'm not mistaken). They way, new users only really need to look at a small section at the top of this docs page.

This part is fine.

> When going down on the page they would see "advanced" and think "ok that is not for me". 

Sort of. You actually do have to update all three parts in order to fully update Qubes, so it's still "for everyone." The Qubes Update tool just happens to be able to do all three parts for you at the same time (not sure about StandaloneVMs, though). This was probably just written before the Qubes Update tool became the main update method.

> Curious users can always look at what goes on behind the scenes & more advanced stuff.

That sounds more like developer documentation, whereas this page is actually simple end-user documentation (even in its current state).

deeplow:
@andrewdavidwong did you merge this by mistake? It was still missing at least an image (I should have marked it as "draft")

andrewdavidwong:
> @andrewdavidwong did you merge this by mistake? It was still missing at least an image (I should have marked it as "draft")
> 

No, it was intentional. See the revised text I committed. It's fine to add an image, as well, but I think it's optional.

deeplow:
I see! I was just confused because you had made a commit outside of this branch and so I thought it had gone to the docs with the `[ add image here ]`. But I see you made a commit after outside of this: https://github.com/QubesOS/qubes-doc/commit/b8d1b4ceea9a14ab52d675c0ceb9ce9abc80fbb5.

Thanks for the clarification!

hackerncoder:
There is a actually a screenshot containing it: https://github.com/QubesOS/qubes-attachment/blob/master/wiki/QubesScreenshots/r4.0-software-update.png

andrewdavidwong:
> I see! I was just confused because you had made a commit outside of this branch and so I thought it had gone to the docs with the `[ add image here ]`. But I see you made a commit after outside of this: [b8d1b4c](https://github.com/QubesOS/qubes-doc/commit/b8d1b4ceea9a14ab52d675c0ceb9ce9abc80fbb5).
> 
> Thanks for the clarification!

No, I made that commit inside of your branch; I just didn't push it to this PR. (I made the commit in your branch, then merged your branch.)

-----

> There is a actually a screenshot containing it: https://github.com/QubesOS/qubes-attachment/blob/master/wiki/QubesScreenshots/r4.0-software-update.png

Thanks! I've added this image to the page.

deeplow:
@andrewdavidwong I was thinking of more along the lines of:

updates

andrewdavidwong:
> @andrewdavidwong I was thinking of more along the lines of:
> 
> updates

Added.

PR #1131, opened 2021-04-08
Add Markdown convention rule about full reference links

tokideveloper:


andrewdavidwong:
In that case, we might as well just prescribe non-reference-style links instead, right?

tokideveloper:
You mean something like `[our website](/https://example.com/)`?

Yes, we could but I didn't want to be too restrictive. That's why I chose that rule.

Shall I add a commit that replaces my rule with a rule that prescribes non-reference-style links?

andrewdavidwong:
Yeah, let's just recommend only non-reference-style links.

tokideveloper:
I can adapt all reference links to our new rule. Is one big commit (in a new PR) okay?

andrewdavidwong:
> I can adapt all reference links to our new rule. Is one big commit (in a new PR) okay?

Should be ok, I think. @marmarek, do you foresee any problems with that?

In any case, we can review the PR once it's made and see what CI says about it.

marmarek:
> I can adapt all reference links to our new rule. Is one big commit (in a new PR) okay?

Yes, should be ok.

PR #1136, opened 2021-05-03
add link to instructions on forum via email

deeplow:


deeplow:
The other alternative could be to copy the text on that post into a subheading of "# Forum"

andrewdavidwong:
> The other alternative could be to copy the text on that post into a subheading of "# Forum"

No, this is good. Linking is better than duplicating. 🙂

PR #1144, opened 2021-05-19
Show recommended requirements first

deeplow:
@andrewdavidwong what do you think about showing the recommended first? I feel like it should be what's emphasized. As discussed in the thread ["Is 4GB of RAM enough for running Qubes OS?"](https://qubes-os.discourse.group/t/is-4gb-of-ram-enought-for-running-qubes-os/4078/) 4GB is not enough for regular usage.

### DIFF
--- a/user/hardware/system-requirements.md
+++ b/user/hardware/system-requirements.md
@@ -21,14 +21,6 @@ title: System Requirements
   We strongly recommend consulting the <a href="/hcl/">Hardware Compatibility List</a> to verify that Qubes can install and run on your specific model in the ways you need it to.
 </div>
 
-## Minimum
-
-- **CPU:** 64-bit Intel or AMD processor (also known as `x86_64`, `x64`, and `AMD64`)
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
It's standard practice to list minimal first, followed by recommended. This is the order I see used everywhere.

andrewdavidwong:
> As discussed in the thread ["Is 4GB of RAM enough for running Qubes OS?"](https://qubes-os.discourse.group/t/is-4gb-of-ram-enought-for-running-qubes-os/4078/) 4GB is not enough for regular usage.

This is the wrong solution to the problem. The correct solution is simply to increase the minimum required memory to the true minimum amount required for regular usage.

andrewdavidwong:
(I could have sworn I had this exact discussion with someone at some point in the past, but I can't remember the details or why no change to the minimum reqs came of it.)

marmarek:
Indeed, 4GB is not enough in practice. It allows to start all the system VMs, and one other VM. It is possible to configure the system to be a bit better at this (like, combining sys-net and sys-usb into one), but that comes at the cost of reduced isolation.
I do have some test systems with 4GB of RAM and it is enough to run some of the tests, but I wouldn't call it normal usage.

We run most of the tests (including some starting several VMs for setups like split-gpg) with 6GB - I think it makes sense to bump minimum to that value.

andrewdavidwong:
> We run most of the tests (including some starting several VMs for setups like split-gpg) with 6GB - I think it makes sense to bump minimum to that value.

Done: https://github.com/QubesOS/qubes-doc/commit/58f9a3fc12de191a68636c14d546e50df927599b

> (I could have sworn I had this exact discussion with someone at some point in the past, but I can't remember the details or why no change to the minimum reqs came of it.)

Ah, I remember now. I was thinking of when we did this for the _recommended_ memory requirement.

PR #1147, opened 2021-05-27
add advice on choosing hardware

deeplow:
Stab at fixing https://github.com/QubesOS/qubes-issues/issues/6638

@SvenSemmler your input here is welcome if you have the bandwidth.

### DIFF
--- a/user/hardware/system-requirements.md
+++ b/user/hardware/system-requirements.md
@@ -46,7 +46,7 @@ title: System Requirements
 
 ## Choosing Hardware
 
-Before acquiring a computer you should be aware that not all computers are compatible with Qubes. Satifying the above requirements does not guarantee Qubes will run on it. So it is generally a good idea to choose a laptop which has been released 2+ years ago and checking the resources below.
+Beyond the advice at the top of the page, it is generally a good idea to choose a laptop which has been released 2+ years ago.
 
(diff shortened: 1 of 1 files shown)


SvenSemmler (changes requested):
"does not guarantee qubes will on it" ... there is a "run" or "function" missing.

SvenSemmler (changes requested):
bellow:  hollow, loud, animal cry, as a bull or cow
below: lower in place, rank or value

;-) ... I used to make that mistake for years (German native speaker)

SvenSemmler:
Other than the typos, I think this is a good change. Thanks for doing this @deeplow !

deeplow:
@SvenSemmler thanks for taking a look. Oh my gawd these typos :sweat_smile:

I've also noticed basically the exact same text is in the notice at the top of the page. And it has been there for 4+ years... How did I not notice?!

So I've changed it to add the only missing thing. But I'm not sure this change still makes sense if it's just to add that tiny bit of info.

SvenSemmler (approved):
Very nice!

andrewdavidwong:
- "The advice at the top of the page" is a bit vague and unnecessary
 - This is going to come across to non-Linux-familiar people as weird advice without further explanation, and I'm not sure that further explanation is warranted here. Linux people generally already know that compatibility is better on devices that are at least a couple years old.

Might be better just to add your advice to this doc and link to it: https://github.com/Qubes-Community/Contents/blob/master/docs/hardware/hardware-selection.md

PR #1149, opened 2021-06-02
Update test-bench machine setup

deeplow:
Updates the developer documentation for remote test bench.

Spent some hours removing some dust off of this 7-year-old documentation.


developer/debugging/test-bench.md
```diff
@@ -16,57 +16,72 @@ This guide shows how to set up simple test bench that automatically test your co
...
+            echo "${prog}: Driver ${2} not found"
+            return 1
+        fi
+        echo ${1} >${path}
+    }
+
+    pciunbind ${BDF}
+    pcibind ${BDF} ${DRIVER}
+
+    sudo dhclient
```
deeplow, line 73: Surprisingly to me without `sudo` here, it gives me the error `Cannot find device "eth0"`, even though the script itself is running with `sudo`.

developer/debugging/test-bench.md
```diff
@@ -16,57 +16,72 @@ This guide shows how to set up simple test bench that automatically test your co
 
 We will set up a spare machine (bare metal, not a virtual) that will be hosting our experimental Dom0. We will communicate with it via Ethernet and SSH. This tutorial assumes you are familiar with [QubesBuilder](/doc/qubes-builder/) and you have it set up and running flawlessly.
 
-## Setting up the machine
+<div class="alert alert-warning" role="alert">
+  <i class="fa fa-exclamation-triangle"></i>
+  <b>Notice:</b>
+  This setup intentionally weakens some security properties in the testing system. So make sure you understand the risks and use exclusively for testing.
+</div>
```
marmarek, line 23: We should avoid using HTML in markdown files directly, it makes it harder to convert to other formats (like PDF) and also to translate. @andrewdavidwong do you remember an alternative for notes like this?

developer/debugging/test-bench.md
marmarek: It just takes some time for the driver to pickup the device. `sleep 1` would be sufficient too.

marmarek (changes requested):

developer/debugging/test-bench.md
andrewdavidwong: > We should avoid using HTML in markdown files directly, it makes it harder to convert to other formats (like PDF) and also to translate. @andrewdavidwong do you remember an alternative for notes like this?

I also remember pointing out this problem and asking for a solution, but I don't recall anyone coming up with a solution.

andrewdavidwong: FWIW, I have also been using these alerts throughout the documentation, even though I also hate having HTML in Markdown files. My reasoning is that the value of clarity and preventing users from missing critical information outweighs the cost of having HTML in Markdown files.

developer/debugging/test-bench.md
deeplow: That makes more sense. Added that instead.

developer/debugging/test-bench.md
deeplow: Yes, I [am aware](https://github.com/QubesOS/qubes-issues/issues/1019#issuecomment-752507119). In double I decided to copy what was already on another page.

Is the following more acceptable? (block quote included)

> **Notice**:
> This setup intentionally weakens some security properties in the testing system. So make sure you understand the risks and use exclusively for testing.

deeplow:
After enabling internet access in dom0 `dnf install <PACKAGE>` stops being able to find packages. It just says the don't exist.

Is there any command parameter that can make it find the packages or is the best way to install packages just re-adding the network interface back to `sys-net`?

marmarek:
> After enabling internet access in dom0 `dnf install <PACKAGE>` stops being able to find packages in dom0. It just says they don't exist. 

Use `dnf --setopt=reposdir=/etc/yum.repos.d install <PAKCAGE>`

> And `qubes-dom0-update` is not usefull because `sys-net` does not have the network card.

In practice I keep one network interface (wifi one) in `sys-net` for this. But it's only a viable option if you have both wired and wireless network in the system.
There is another way: connect network to dom0 via `sys-net`. It isn't as reliable as direct connection (you can't for example restart `sys-net`), but it works (save as `dom0-network-via-netvm` and call via `sudo`):
```
#!/bin/sh

set -x
set -e
xl network-attach 0 ip=10.137.99.1 script=/etc/xen/scripts/vif-route-qubes backend=sys-net
sleep 2
ip a a 10.137.99.1/24 dev eth0
ip l s eth0 up
ip r a default dev eth0
echo -e 'nameserver 10.139.1.1\nnameserver 10.139.1.2' > /etc/resolv.conf
qvm-run -p --no-gui -u root sys-net 'systemctl stop qubes-firewall'
sleep 2
qvm-run -p --no-gui -u root sys-net 'iptables -t nat -I PREROUTING -p tcp --dport 22 -j DNAT --to 10.137.99.1'
qvm-run -p --no-gui -u root sys-net 'iptables -I FORWARD -p tcp --dport 22 -d 10.137.99.1 -j ACCEPT'
systemctl start sshd
```

As the other method, this is something suitable on a test system only.

deeplow:
@marmarek makes sense. I've added the `--setopt` method to the doc. I've also provided advice in the case the developer has two network cards, one can be left in `sys-net`.

The dom0 networking through `sys-net` also looks neat. And that script is quite didactic in itself. But for the sake of brevity, I added the first method.

developer/debugging/test-bench.md
deeplow: I believe I'm done with the proposed changes.

deeplow:
I believe I am done with the changes. Ready for final review.

deeplow:
Rebased from master

developer/debugging/test-bench.md
```diff
@@ -15,57 +15,71 @@ This guide shows how to set up simple test bench that automatically test your co
...
+
+4. Configure your DHCP server so your testbench gets static IP and connect your machine to your local network. You should ensure that your testbench can reach the Internet.
+
+5. Install `openssh-server` on your testbench (since dom0 now has internet access the command `qubes-dom0-update` is not needed).
+ 
+    ~~~
+    sudo dnf install openssh-server
+    ~~~
+
+> **Note:** If you want to install additional software in dom0 and your only network card was assigned to dom0, then instead of the usual `sudo qubes-dom0-update install <PACKAGE>` you run `sudo dnf --setopt=reposdir=/etc/yum.repos.d`.
```
marmarek, line 82: The "install" word should be with dnf command, not qubes-dom0-update one.

developer/debugging/test-bench.md
```diff
@@ -15,57 +15,71 @@ This guide shows how to set up simple test bench that automatically test your co
...
+    sleep 1
+    dhclient
+    ```
+
+4. Configure your DHCP server so your testbench gets static IP and connect your machine to your local network. You should ensure that your testbench can reach the Internet.
+
+5. Install `openssh-server` on your testbench (since dom0 now has internet access the command `qubes-dom0-update` is not needed).
+ 
+    ~~~
+    sudo dnf install openssh-server
```
marmarek, line 79: This should also get `--setopt` option.

marmarek (changes requested):

developer/debugging/test-bench.md
```diff
@@ -15,57 +15,71 @@ This guide shows how to set up simple test bench that automatically test your co
...
+
+4. Configure your DHCP server so your testbench gets static IP and connect your machine to your local network. You should ensure that your testbench can reach the Internet.
+
+5. Install `openssh-server` on your testbench.
+ 
+    ~~~
+    sudo dnf --setopt=reposdir=/etc/yum.repos.d install openssh-server
+    ~~~
+
+> **Note:** If you want to install additional software in dom0 and your only network card was assigned to dom0, then _instead_ of the usual `sudo qubes-dom0-update install <PACKAGE>` now you run `sudo dnf --setopt=reposdir=/etc/yum.repos.d install <PACKAGE>`.
```
marmarek, line 82: Almost. `qubes-dom0-update` has no `install` argument.

deeplow: Ah. You're right! Sorry about that.

marmarek (approved):
```
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA256

Approved as of 99083295b0a39cd03cd688bb85c363f381d3b934
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCAAdFiEEnYWYU8Ll1fIX9iss8yiUvpaEk4oFAmDWZSwACgkQ8yiUvpaE
k4qrOQ/+NMe0hrvqnRuKm4kEiO/CS/8QquDY/OzRdefMfGU/Xcz+RmLm1NDzYr7p
hqbFMs7yPJr5nIby5A/rX9+xAFQm4Zv/OEDdyVjoVUU50IztxGmoMrFCKO0Fh0In
gIK+o4HU1bZm6mWD0Fr1MbuB7LXZ1a8BUsd9ssRm65EQDXLmNxBTFe+/DD3JLxD1
LboXv8MJh7Aj2LUw3SxEAygvA02v5SExsQvnWa1tuItBUzFlFmTlDMTGgH10l28P
7FBXvaoMuGS113OysWt1GhDt8TQXnbbQ91MTcCeDgUklgabVyODrBt5cCZ0CEFSK
msU7tVI5i+Z8xO6ufJu4X7VVoxZegVp8+Uqju8JK7H0X9w3sQNL+Ki0gj3y0rp0m
bU23kCUlcigUz3m3jTnkCYVtd8z5IXTHzG6JaPHoilr0k7YFn+d30w2XDxBxNYR1
GH0FbvA5LEvMmkabGV2ruPXQGsk+c92+YOYnMR88grlLFMIitVk1x4LlaMoPUW+Q
qrUDnVeqKDr7HXg6kREQxxJalHawp/C89/DTEhj/kPNs9DLCUt8E19ME65PYyke8
zCzyibe6WQ/Z9xgiCkSUVFPsdhZLzvBT69b4ucJWKeJ/2COP/KnZWOxErkFiWwpm
p0tFIZMYLIO6UlnbRocxNbK1VOFqyq+GV4Y8ufu7B/B5sAh53DU=
=qUaA
-----END PGP SIGNATURE-----
```

andrewdavidwong:
Hi all,

There are a ton of errors in the CI log, but I think this is because CI is checking an outdated version of the repository. (For example, I see it flagging something that I know has already been removed.) I'm *guessing* this is because this PR has not been updated with the master branch, so I'm going to try pushing to this PR to see if that updates things correctly. (Unfortunately, this is cumbersome to do, but oh well.)

andrewdavidwong:
Wow, that's a lot of commits (and, unfortunately, some unrelated comments came along with it). If there's a better way to handle this in the future, please let me know.

Now to wait and see whether this solves the CI problem...

andrewdavidwong:
> Wow, that's a lot of commits (and, unfortunately, some unrelated comments came along with it). If there's a better way to handle this in the future, please let me know.

@marmarek, I have the vague sense that you might have already told me how I should handle this, but I can't remember and can't find the conversation anywhere... I hate having a terrible memory. :frowning_face: 

This time I'll document it so that I can reference it later.

andrewdavidwong:
On the plus side, no more CI errors!

fepitre:
@andrewdavidwong to be sure you can run 'PipelineRetry'

andrewdavidwong:
> @andrewdavidwong to be sure you can run 'PipelineRetry'

Wait, is that necessary? If so, why? I believe it just ran a fresh check after I pushed all the commits from origin/master, so shouldn't the current (passing) check be good enough?

fepitre:
> > @andrewdavidwong to be sure you can run 'PipelineRetry'
> 
> Wait, is that necessary? If so, why? I believe it just ran a fresh check after I pushed all the commits from origin/master, so shouldn't the current (passing) check be good enough?

Yes it's fine it you merged it recently.

andrewdavidwong:
> > > @andrewdavidwong to be sure you can run 'PipelineRetry'
> > 
> > 
> > Wait, is that necessary? If so, why? I believe it just ran a fresh check after I pushed all the commits from origin/master, so shouldn't the current (passing) check be good enough?
> 
> Yes it's fine it you merged it recently.

Well, I haven't merged this PR yet, but there was some kind of automerge when I pulled this PR. Is that what you mean?

fepitre:
> > > > @andrewdavidwong to be sure you can run 'PipelineRetry'
> > > 
> > > 
> > > Wait, is that necessary? If so, why? I believe it just ran a fresh check after I pushed all the commits from origin/master, so shouldn't the current (passing) check be good enough?
> > 
> > 
> > Yes it's fine it you merged it recently.
> 
> Well, I haven't merged this PR yet, but there was some kind of automerge when I pulled this PR. Is that what you mean?

Yes I meant when merge some commits into this PR.

marmarek:
> There are a ton of errors in the CI log, but I think this is because CI is checking an outdated version of the repository.

There shouldn't be the need to do anything manually, CI tests PR branch merged into master, so recent changes in master should be included too. But, there is some issue with CI that we haven't figured out yet and sometimes it uses outdated branch. The workaround is to trigger the run again with "PipelineRetry" comment.

andrewdavidwong:
> There shouldn't be the need to do anything manually, CI tests PR branch merged into master, so recent changes in master should be included too. But, there is some issue with CI that we haven't figured out yet and sometimes it uses outdated branch. The workaround is to trigger the run again with "PipelineRetry" comment.

Thank you! I've documented the `PipelineRetry` command here: https://github.com/QubesOS/qubes-doc/commit/decc4135f963bca88f65d364d85610c3b18385c7. @marmarek, @fepitre, please add any other CI commands that should be documented.

PR #1153, opened 2021-06-14
Add github-isues#6694

hackerncoder:
- [x] Rename "Introduction" to "Working with qubes and templates"
- [x] Move "GUI & Command Line Tools" to become the first section
- [x] Rename "GUI & Command Line Tools" to "User Interface"
- [x] In the first paragraph, speak explicitly to technical users desiring command-line tools to navigate.
- [ ] introduce paragraph that speaks to the following points [...]


andrewdavidwong:
I think it might be easier for me just to handle this one myself, as I'm already knee-deep in revising the text with @ninavizz over on the GDoc, which has already changed from the version here. Thank you, though, @hackerncoder!

PR #1157, opened 2021-06-17
remove "install windows or mac"

deeplow:
Generally people don't install these systems. They just come per-installed.


andrewdavidwong:
Good point.

PR #1160, opened 2021-06-19
Mention why one would use minimal templates

deeplow:
Perhaps it is obvious for some why minimal templates make sense, but for some that are just exploring this may not be as obvious. Because if this it may be wise to mention some use-cases at the top so people know if minimal templates make sense for them.

@SvenSemmler I hope you don't mind me totally "borrowing"  your reply from https://qubes-os.discourse.group/t/debian-10-minimal-configuration/2603/54


andrewdavidwong:
There's already a big problem with people trying to use minimal templates and paying no heed to the health warning near the top of this page. I fear your addition would add fuel to the fire.

andrewdavidwong:
> There's already a big problem with people trying to use minimal templates and paying no heed to the health warning near the top of this page. I fear your addition would add fuel to the fire.

I have an idea for how to address this.

deeplow:
> There's already a big problem with people trying to use minimal templates

True.

> I have an idea for how to address this.

Nice. I quite like your solution.

andrewdavidwong:
> Nice. I quite like your solution. Perhaps this type of warning should be used on more pages.

It already is. 🙂

PR #1161, opened 2021-06-20
Update how-to-get-started.md

ninavizz:
- Updated custom Qubes widgets image to match wording to names shown above & remove highlights on user menu/battery 
- Edited "custom" to "unique" wrt "unique to qubes," as sdwdate is not custom _to_ qubes, but is unique(ish) to Qubes.
- Included bullet for Whonix widget, w/ link-out
- Capitalized "A" on Awesome in desktop environments

### DIFF
--- a/user/advanced-topics/awesomewm.md
+++ b/user/advanced-topics/awesomewm.md
@@ -1,11 +1,12 @@
 ---
 lang: en
 layout: doc
-permalink: /doc/awesome/
+permalink: /doc/awesomewm/
 redirect_from:
+- /doc/awesome/
 - /en/doc/awesome/
 ref: 179
-title: awesome (window manager)
+title: AwesomeWM (window manager)
 ---
 
 
@@ -17,17 +18,17 @@ title: awesome (window manager)
 
 ## Installation
 
-awesome can be installed with the standard dom0 installation mechanisms.
+AwesomeWM can be installed with the standard dom0 installation mechanisms.
 
 ```shell_session
 $ sudo qubes-dom0-update awesome
 ```
 
-That's it. After logging out, you can select awesome in the login manager.
+That's it. After logging out, you can select AwesomeWM in the login manager.
 
 ## Development
 
-To [contribute code](/doc/contributing/) you may clone the awesome repository as follows:
+To [contribute code](/doc/contributing/) you may clone the AwesomeWM repository as follows:
 
 ```shell_session
 $ git clone https://github.com/QubesOS/qubes-desktop-linux-awesome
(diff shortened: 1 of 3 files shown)


andrewdavidwong:
Thanks for the PR, @ninavizz!

Regarding "awesome" (the window manager), I recently grappled with this myself. After scouring their web presences and Wikipedia entry, I discovered that project is not consistent in the way it names itself. All three of these variations are used: `awesome`, `Awesome`, and `AwesomeWM`. The goal of your change here seems to be to distinguish the name of the window manager from the regular English word. I agree that that's desirable. In that case, I think we should just change all instances of the name to `AweomseWM` to remove all ambiguity. I'll make that change here and in the rest of the docs.

Regarding your new image, we have specific instructions for [how to add images](https://www.qubes-os.org/doc/doc-guidelines/#how-to-add-images). There are two important reasons for these specific instructions:

1. Many of the visitors to our website are understandably privacy- and security-conscious, so we locally self-host as many resources as possible, including things like fonts and images (see our new [Privacy Policy](https://www.qubes-os.org/privacy/)). This way, the visitor's browser doesn't have to connect to a bunch of potentially untrusted third-party CDNs and domains. For this reason, all images must be hosted from [qubes-attachment](https://github.com/QubesOS/qubes-attachment), which means that you should submit a separate PR adding the image to that repo, then link to it in your PR here. I'll do it for you this time. I just wanted to let you know for the future.
2. If you look closely at the image syntax, what you added is different from what it replaces. The line you replaced (like all of the other images on this page) turns the image itself into a hyperlink directly to the image. Your replacement does not do this. This is important, because turning the image into a link allows a visitor to click on the image in order to view the full version. Following best practices, our website has a responsive design, which allows the website to render appropriately across all screen sizes. When viewing this page on a smaller screen, the image will automatically shrink down to fit the screen. If visitors cannot click on the image to view it in full size, then, depending on their device, they may have no way see the details in the image clearly. I will also fix this for you. Again, just wanted to let you know for the future.

andrewdavidwong:
Oh, I just remembered another reason that it's important to add images in the correct way:

The way you added this image means that the old image would continue to exist at `/attachment/doc/r4.0-q40_widgets.png`. Since you didn't *replace* the old image, the old image would become an outdated and incorrect "trap" that other unsuspecting contributors might think is a good thing to use in other documents. (They would have no way of knowing that it's outdated, and they would be reasonable to think it's a good image, since it's in `/attachment/doc/`.) Also, leaving our old junk around just makes things messy and cluttered compared to cleaning up after ourselves. :slightly_smiling_face: 

(I'm replacing the old image with your new one right now.)

andrewdavidwong:
Another plus of replacing the image in `qubes-attachment` is that we actually don't have to change the image syntax in this file at all.

ninavizz:
You are _far_ too kind and thoughtful with all your guidance, Andrew—TY!

PR #1162, opened 2021-06-21
Fix redirect to 404 from /doc/disposablevm

noskb:
The link appears on the first page of a search result for "qubes disposable", but it is not redirecting correctly.


andrewdavidwong:
Thank you.

PR #1164, opened 2021-06-23
Make replaceable text like <THIS> instead of <this>

deeplow:
[A user](https://qubes-os.discourse.group/t/criando-maquinas-virtuais-de-template-de-qualquer-sistema-operacional/4691) (in Portuguese) didn't see they had to replace <text>, possibly because it was not visibly out of place. If my impression is correct other documentations pages use the <CAPS_FORMATTING> which could help somehow this.

### DIFF
--- a/user/advanced-topics/standalones-and-hvms.md
+++ b/user/advanced-topics/standalones-and-hvms.md
@@ -44,7 +44,7 @@ You can create a standalone in the Qube Manager by selecting the "Type" of "Stan
 Alternatively, from the dom0 command line:
 
 ```
-qvm-create --class StandaloneVM --label <label> --property virt_mode=hvm <vmname>
+qvm-create --class StandaloneVM --label <LABEL> --property virt_mode=hvm <VMNAME>
 ```
 
 (Note: Technically, `virt_mode=hvm` is not necessary for every standalone.
@@ -145,7 +145,7 @@ This mode can be used for any HVM (e.g. FreeBSD running in a HVM).
 In order to create a TemplateHVM you use the following command, suitably adapted:
 
 ~~~
-qvm-create --class TemplateVM <qube> --property virt_mode=HVM --property kernel=''  -l green
+qvm-create --class TemplateVM <QUBE> --property virt_mode=HVM --property kernel=''  -l green
 ~~~
 
 Set memory as appropriate, and install the OS into this template in the same way you would install it into a normal HVM.


andrewdavidwong:
> If my impression is correct other documentations pages use the <CAPS_FORMATTING> which could help somehow this.

No need for impressions: It's explicitly documented!

https://www.qubes-os.org/doc/doc-guidelines/#style-guidelines

user/advanced-topics/standalones-and-hvms.md
```diff
@@ -44,7 +44,7 @@ You can create a standalone in the Qube Manager by selecting the "Type" of "Stan
 Alternatively, from the dom0 command line:
 
 ```
-qvm-create --class StandaloneVM --label <label> --property virt_mode=hvm <vmname>
+qvm-create --class StandaloneVM --label <LABEL> --property virt_mode=hvm <VMNAME>
```
andrewdavidwong, line 47: Out of curiosity, why `<VMNAME>` here but `<QUBE>` below?

andrewdavidwong: Ah, guessing it was just a capitalization change of the existing words without further thought given to changing them.

deeplow: No idea. I was in auto-pilot -- just made it caps everywhere. Probably `<QUBE>` would be better. Feel free to change it.

andrewdavidwong: No worries. Thank you.

deeplow:
@andrewdavidwong this check could be added to the linter. Not sure how easy that would be to do, though.

andrewdavidwong:
> @andrewdavidwong this check could be added to the linter. Not sure how easy that would be to do, though.

No idea how to do that. Might result in too many false positives to be worth it.

What do you think, @marmarek?

andrewdavidwong:
It's also unclear whether this practice is optimal, consider that users seem to keep typing the `<` and `>` characters:

https://qubes-os.discourse.group/t/how-can-i-create-a-template-hvm-of-any-os/4676/4?u=adw

andrewdavidwong:
Maybe the better approach would just be to choose a real value as a "default." If the user knows how to change it and wants to, they will. If not, they can just live with the default.

In other words, just:

```
--label black
```

Sure, some users won't know that they can change `black` to a different color, but least the command will work. And it's not like using `black` instead of another color will hurt them. (Well, maybe in some edge cases, but generally not. This might be a reason to default to red when no other color is a good obvious default.)

andrewdavidwong:
Ah, but there are cases in which we can't choose a default for the user. For example:

```
3. If you have an ISO image of the installation media located in a qube (the
   qube where the media is located must be running):

   qvm-start <YOUR_HVM> --cdrom=<YOUR_OTHER_QUBE>:/home/user/<YOUR_INSTALLER.ISO>
```

We could choose a default name for the new HVM, but obviously we have no way of knowing which other qube houses the installer ISO or what the installer ISO is called.

PR #1169, opened 2021-07-02
Update how-to-use-block-storage-devices.md

ninavizz:
Remove use of the phrase "GUI" and specifically speak to the "Devices WIdget." Especially since it is now referenced in the Getting Started section, which should cross-link, here.

### DIFF
--- a/user/how-to-guides/how-to-use-block-storage-devices.md
+++ b/user/how-to-guides/how-to-use-block-storage-devices.md
@@ -39,7 +39,7 @@ Click on one and your USB drive will be attached!
 However, it often means the app qube won't detect the new partition and you will need to manually mount it inside the app qube.
 See below for more detailed steps.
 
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
> which should cross-link, here.

Please add the cross-links when you get a chance.

user/how-to-guides/how-to-use-block-storage-devices.md
```diff
@@ -39,7 +39,7 @@ Click on one and your USB drive will be attached!
 However, it often means the app qube won't detect the new partition and you will need to manually mount it inside the app qube.
 See below for more detailed steps.
 
-## Block Devices in VMs
+## Block devices in VMs
```
andrewdavidwong, line 42: I don't understand why this word is being lowercased when the heading at the same level above it is in title case.

PR #1170, opened 2021-07-02
Update support.md

ninavizz:
- Added link to glossary to the top of the page
- Re-formatted H2s on top section, to active-voice topical headers
- Removed some informal "aw, shucks" verbiage and replaced with simpler phrasing

### DIFF
--- a/introduction/support.md
+++ b/introduction/support.md
@@ -21,18 +21,14 @@ get the support you need, and we want to make sure our community remains a
 friendly place by ensuring we all follow the [Code of
 Conduct](/code-of-conduct/).
 
-## How to get help and support
+## Glossary of terms
+What does something mean? There may be a simple answer to that, in our [glossary of terms](https://www.qubes-os.org/doc/glossary/).
 
-First, let's see what kind of help you need.
+## Help and support
 
-### I would like to report a security vulnerability
+### Find a solution to a problem
 
-That sounds more like you helping us! Thanks! Please see [Reporting Security
-Issues in Qubes OS](/security/#reporting-security-issues-in-qubes-os).
-
-### I have a problem or a question
-
-No worries! Here's how we recommend proceeding:
+Qubes OS can be tough to learn and navigate. Here are some steps we suggest:
 
 1. Check the [documentation](/doc/). There may already be a page about it.
    Specifically, check out the [How-to Guides](/doc/#how-to-guides) and
@@ -53,33 +49,33 @@ No worries! Here's how we recommend proceeding:
    archives](https://www.mail-archive.com/qubes-users@googlegroups.com/). There
    may have already been a relevant thread.
 
-### I didn't find a solution or an answer!
+### Ask the community for help
 
-Sorry to hear that! In that case, we recommend asking for help on the [Qubes
+If you are not able to find an answer in our docs or community archives, we recommend asking for help on the [Qubes
 Forum](https://forum.qubes-os.org/) or on the [`qubes-users` mailing
 list](#qubes-users). Choose the venue you prefer, but please don't ask on both
-at the same time! Before you ask, please review our [discussion
+at the same time. Before you ask, please review our [discussion
 guidelines](#discussion-guidelines) and StackOverflow's advice on [how to ask
 good questions](https://stackoverflow.com/help/how-to-ask). Don't forget to
 [stay safe](#staying-safe)!
 
-### I don't need support, but I think I found a bug
+### Report a bug
 
 We'd be grateful if you reported it (but please make sure no one else has
 already reported it first)! Please see [Issue Tracking](/doc/issue-tracking/)
 for details.
 
-### I don't need support, but I'd like to request a feature
+### Report a security vulnerability
+
+That sounds more like you helping us! Thanks! Please see [Reporting Security
+Issues in Qubes OS](/security/#reporting-security-issues-in-qubes-os).
+
+### Request a feature
 
 No promises, but we'd be happy to consider it! Please see [Issue
 Tracking](/doc/issue-tracking/) for details.
 
-### I just want to discuss Qubes!
-
-Great! Head on over to the [Qubes Forum](https://forum.qubes-os.org/) or
-the [`qubes-users` mailing list](#qubes-users).
-
-### How can I get involved and contribute?
+### Get involved and contribute
 
 Thank you for asking! Please see [How to Contribute](/doc/contributing/) for
 all the ways you can do so.


introduction/support.md
```diff
@@ -21,18 +21,14 @@ get the support you need, and we want to make sure our community remains a
 friendly place by ensuring we all follow the [Code of
 Conduct](/code-of-conduct/).
 
-## How to get help and support
+## Glossary of terms
+What does something mean? There may be a simple answer to that, in our [glossary of terms](https://www.qubes-os.org/doc/glossary/).
```
andrewdavidwong, line 25: Always use relative links in the documentation (e.g., `/doc/glossary/`).

andrewdavidwong:
> Added link to glossary to the top of the page

I don't think it makes sense to have this in a separate section before everything else. What exactly is the goal here?

> Re-formatted H2s on top section, to active-voice topical headers

The headings were already in active voice. Both the old and new headings are in active voice. I think you mean a different grammatical term, but I'm not sure which one.

In any case, why are these two headings better than the old ones?

> Removed some informal "aw, shucks" verbiage and replaced with simpler phrasing

Why is this an improvement?

introduction/support.md
```diff
@@ -21,18 +21,14 @@ get the support you need, and we want to make sure our community remains a
...
-### I would like to report a security vulnerability
+### Find a solution to a problem
 
-That sounds more like you helping us! Thanks! Please see [Reporting Security
-Issues in Qubes OS](/security/#reporting-security-issues-in-qubes-os).
-
-### I have a problem or a question
-
-No worries! Here's how we recommend proceeding:
+Qubes OS can be tough to learn and navigate. Here are some steps we suggest:
```
andrewdavidwong, line 31: It seems weird for us to say of our own OS that it's tough to learn and navigate. A reader might think, "If even *you* think it's tough to learn and navigate, then why did you *make* it that way?!"

introduction/support.md
```diff
@@ -53,33 +49,33 @@ No worries! Here's how we recommend proceeding:
...
-No promises, but we'd be happy to consider it! Please see [Issue
-Tracking](/doc/issue-tracking/) for details.
+That sounds more like you helping us! Thanks! Please see [Reporting Security
+Issues in Qubes OS](/security/#reporting-security-issues-in-qubes-os).
 
-### I just want to discuss Qubes!
+### Request a feature
 
-Great! Head on over to the [Qubes Forum](https://forum.qubes-os.org/) or
-the [`qubes-users` mailing list](#qubes-users).
```
andrewdavidwong, line 80: What's the rationale for eliminating this section? There have been many times in the past when people have asked where they should go to discuss Qubes, especially after being told that qubes-issues is *not* that place.

introduction/support.md
ninavizz: Empathy matters, and this is nicely empathetic. It takes a long time for software to become usable. Qubes does something no other software really does in a GUI. I think everyone understands that it's real humans who make it, and that we're not resource-rich enough to have something ultra-polished. This is good. :) Thank you for eliminating one exclamation point. There are currently very many—too many—in the docs. :)

introduction/support.md
ninavizz: I don't remembering doing this, tbh. "I just want to discuss Qubes!" is a strange header. The exclamation-point in a header also feels extraneous.

introduction/support.md
andrewdavidwong: > Empathy matters, and this is nicely empathetic. It takes a long time for software to become usable. Qubes does something no other software really does in a GUI. I think everyone understands that it's real humans who make it, and that we're not resource-rich enough to have something ultra-polished. 

Oh, if only that were true! In my experience, it's quite the opposite.

introduction/support.md
andrewdavidwong: > I don't remembering doing this, tbh. "I just want to discuss Qubes!" is a strange header. The exclamation-point in a header also feels extraneous. 

Agreed, but I couldn't think of anything better. ☹️

introduction/support.md
andrewdavidwong: Also, I agree that empathy matters, which is why I tried to write this page in a more empathetic tone. It seems to me that many of the changes in this PR are moving from more empathetic to less empathetic. If that's not the case, I'd like to understand how it's not.

SvenSemmler:
On 7/4/21 2:50 AM, Nina Eleanor Alter wrote:
> Qubes OS can be tough to learn and navigate.

This will turn away people who would otherwise try Qubes OS.

It's fine or maybe even appropriate for a community member to show 
empathy for a struggling new user and make them feel welcome in a 
conversation on the forum or on the mailing list.

It's quite another thing for the Qubes OS project to say so on it's 
official website!

In most cases it's not even Qubes OS folks are struggling with, but 
Linux/XFCE. They would have the exact same struggles trying to use 
Xubuntu or any of the other distributions. I don't see them writing "can 
be tough to learn and navigate" on their websites.

It's also frankly disrespectful to all contributors working very hard to 
make Qubes OS as easy to use as possible. This one deserves strong push 
back.

-- 

introduction/support.md
ninavizz: That's good feedback! Intentions don't always come through as impact, and I'm not immune to that in my contributions.

ninavizz:
All of the above is good/solid feedback, and rather than itterate on these specific changes I'd rather just step back and keep this discussion on hand as a closed PR... but revisit some of these things, in other PRs. If that makes sense. I'll be mindful to look at these past discussion points to try and not repeat the same things folks object to, here. 

@SvenSemmler It's the mental-models involved in using any hypervisor system, that are hard to learn—and then Linux-ings and XFCE, only compound that. I don't disagree, at all, with any of what you're saying, though. If anything, Joanna and Marek have been guiding Qubes, from the start, _to be_ as easy and approachable as possible. But in many ways, that's like trying to breed Jalapeños that won't also clear your sinuses. It's entirely possible—just very hard.

andrewdavidwong:
One thing I do like in this PR is moving the "report a security issue" section down toward the bottom rather than having it at the top, so I'm going to make that change here and in other locations (e.g., the sidebar and the footer links) in a separate commit right now.

My initial rationale for putting it on top is because it seems more important. If anyone wants to responsibly disclose a security vulnerability, I want to make sure it's obvious how to do so. The last thing I want is for a security researcher to publicly say, "I found a vuln in Qubes, but they don't even have a way to report security issues! And they call themselves a secure OS!" So I've generally put it near the top.

However, @ninavizz makes a good point about prioritizing the things that most users will need to find most of the time. There will be vastly more cases of regular users needing to find out how to get help than there will be of security researchers disclosing vulns, *and* the security researchers are much more likely to be savvy enough to find the information they need, whereas regularly users often need the "big flashing neon sign" treatment to not miss stuff.

PR #1173, opened 2021-07-07
Update templates.md

ninavizz:
- Added pieces composed for "Getting Started" to the introduction, to build language/concept parity/consistency
- Added my own language, stating what was very much _not_ clear to me in the current documentation—that "Creating" a Template is not a thing one does. They're already created... end  users download and install "Add" what others "Create." Of course, I could be a lone outlier user in that read on today's docs. I also wanted this to be stated early, so that short-attention-span users (cough, cough) are sure to not miss it when visiting this page in search of instructions for adding (cough, or creating) more templates. 
- Adapted existing intro language, to follow new intro language, such that concepts are not lost.
- Tweaked a few of the headers throughout, to more clearly anchor page-skimmers to where sought content is.
  - Used Title Case to boost header/topic discoverability among page-skimmers.

### DIFF
--- a/user/templates/templates.md
+++ b/user/templates/templates.md
@@ -11,12 +11,9 @@ ref: 131
 title: Templates
 ---
 
-In [Getting Started](/doc/getting-started/), we covered the distinction
-in Qubes OS between where you *install* your software and where you *run* your
-software. Your software is installed in [templates](/doc/glossary/#template).
-Each template shares its root filesystem (i.e., all of its programs and system
-files) with all the qubes based on it. [App qubes](/doc/glossary/#app-qube) are
-where you run your software and store your data.
+In [Getting Started](/doc/getting-started/) it was explained that each [app qube](/doc/glossary/#app-qube) is based on another type of qube called a [template](/doc/glossary/#template). Templates are usually named after their operating system (often a [Linux distribution](https://en.wikipedia.org/wiki/Linux_distribution)) and corresponding version number. "Creating" [your own template](/doc/template-implementation) is not a trivial process, and something best left to developers mindful of security best practices. There are many ready-to-use templates to choose from, however, and you can download and have as many as you like (see below). 
+
+More than one qube can be based on the same template. An app qube cannot modify its template in any way. This means that if a qube is ever compromised, its template will remain safe. For that reason, in Qubes OS the template is where you *install* your software; the app qube, conversely, is where you *run* your software and also store your files. Each template shares its root filesystem (i.e., all of its programs and system files) with all the app qubes based on it. 
 
 The template system has significant benefits:
 
@@ -50,7 +47,7 @@ to have multiple templates installed in order to provide:
 * Different environments (e.g., Fedora, Debian, Whonix)
 * Different tools (e.g., office, media, development, hardware drivers)
 
-## Official
+## Officially Supported Templates
 
 These are the official Qubes OS Project templates. We build and release updates
 for these templates. We guarantee that the binary updates are compiled from
@@ -62,7 +59,7 @@ exactly the same source code as we publish.
 * [Debian](/doc/templates/debian/)
 * [Debian Minimal](/doc/templates/minimal/)
 
-## Community
+## Community Created Templates
 
 These templates are supported by the Qubes community. Some of them are
 available in ready-to-use binary package form (built by the Qubes developers),
@@ -83,7 +80,7 @@ developers do not test them.
 * [Gentoo](/doc/templates/gentoo/)
 * [Gentoo Minimal](/doc/templates/minimal/)
 
-## Installing
+## Installing Templates
 
 Certain templates come preinstalled with Qubes OS. However, there may be times
 when you wish to install a fresh template from the Qubes repositories, e.g.:
@@ -132,7 +129,7 @@ After installing a fresh template, we recommend performing the following steps:
 
 3. If desired, [uninstall the old template](#uninstalling).
 
-## Updating
+## Updating Templates
 
 Please see [How to Update](/doc/how-to-update/).
 
@@ -140,7 +137,7 @@ Please see [How to Update](/doc/how-to-update/).
 
 Please see [How to Install Software](/doc/how-to-install-software).
 
-## Uninstalling
+## Uninstalling Template
 
 The procedure for uninstalling a template depends on how it was created.
 
@@ -200,11 +197,11 @@ Applications Menu entries for backups of removed qubes can also be found in
 $ rm /usr/local/share/applications/<TEMPLATE_NAME>
 ```
 
-## Reinstalling
+## Reinstalling Templates
 
 Please see [How to Reinstall a Template](/doc/reinstall-template/).
 
-## Switching
+## Switching Templates
 
 When you install a new template or upgrade a clone of a template, it is
 recommended that you switch everything that was set to the old template to the
@@ -241,11 +238,11 @@ new template:
     [user@dom0 ~]$ qubes-prefs default-dispvm <NEW_DISPOSABLE_TEMPLATE>
     ```
 
-## Advanced
+## Advanced Topics
 
 The following sections cover advanced topics pertaining to templates.
 
-### Inheritance and persistence
+### Inheritance and Persistence
 
 Whenever an app qube is created, the contents of the `/home` directory of its
 parent template are *not* copied to the child app qube's `/home`. The child app
@@ -270,7 +267,7 @@ manner, you must make those changes in the parent template.
 <sup>2</sup>Following shutdown  
 <sup>3</sup>Includes [disposable templates](/doc/glossary/#disposable-template)
 
-### Trusting your templates
+### Trusting Your Templates
 
 As the template is used for creating filesystems for other app qubes where you
 actually do the work, it means that the template is as trusted as the most


ninavizz:
...also, @andrewdavidwong if a Qubes OS Foundation as a business entity does not exist (I do not think that it does?) the foundational concepts spoken to in the below sentence in "Trusting Your Templates" should be tweaked:

>>   But why trust Fedora?
> 
> Because we chose to use Fedora as a vendor for the Qubes OS foundation (e.g. for dom0 packages and for app qube packages).

h01ger:
2 comments

1. creating templates might be left to experienced people but certainly copying and modifying templates is a basic task. (eg adding wireshark to a debian-11 template and calling the template debian-11-wireshark is a simply task any qubes user should be able to to do.) - I think this should be explained here and I don't think it is.
2. it's totally valid to install stuff in app qubes. I do it all the time. sometimes because I want the software to be gone after i used, sometimes I install it in /usr/local or $HOME. - I'm not sure this should be explained here, maybe just hinting the fact is got enough for getting started.

andrewdavidwong (changes requested):
> - Added pieces composed for "Getting Started" to the introduction, to build language/concept parity/consistency

We want to be very careful with duplicating content, because it's almost invariably problematic. The small amount here is _probably_ okay, though.

> - Added my own language, stating what was very much _not_ clear to me in the current documentation—that "Creating" a Template is not a thing one does. They're already created... end  users download and install "Add" what others "Create." Of course, I could be a lone outlier user in that read on today's docs. I also wanted this to be stated early, so that short-attention-span users (cough, cough) are sure to not miss it when visiting this page in search of instructions for adding (cough, or creating) more templates. 

In the context of Qubes templates, "create" often means "clone from existing." From what you have here, a reader might conclude that customizing templates is extremely difficult, which would not be true. It will probably require further explanation to be really clear about this.

> - Adapted existing intro language, to follow new intro language, such that concepts are not lost.
> 
> - Tweaked a few of the headers throughout, to more clearly anchor page-skimmers to where sought content is.
> 

This is going to break a bunch of links. :/

Why is it better to repeat the word "template" in subheadings so many times on a page already called "Templates"?

>   - Used Title Case to boost header/topic discoverability among page-skimmers.

This is not a good reason to use title case. Besides, we already have a (mostly) established practice of using sentence case for subheadings within pages, and we've already discussed changing the page titles to sentence case.

ninavizz:
TY Holger for stating #2! I actually could not figure-out how to get Signal to install in a Debian Template in a fashion I could then access in a dependent app-vm—so ended-up installing it in the app-vm, directly.

Re: #1... I agree in principle and am opem to different wording... but feel it needs to be made clear, "This is how this system was designed to work, and here is where you go to get those resources;" without clouding that by also ensuring edge cases are covered. Even if it is common for people to do the things you described, in the first few paragraphs folks need to hear how the system was designed to work out-of-the-box.

A problem I keep seeing in a lot of FOSS documentation, is an insistence on telling users allthethings they can do in a single gulp—without separating "Here is the by-design path, and here is the ambitious-developer path, and here is the hacker path." I want to help non-technical users more clearly see their paths, as they are also more likely to want the quickest read and have the least interest in learning "all the things." More advanced users, conversely, are more likely willing to read it all and to look for opportunities to hack.

ninavizz:
(and, yep, will ping once I update the text per the suggestions—unless GH does that for me?)

h01ger:
I do agree that documentation often is and should not be overloaded. However
I also think the "Getting started" page should not state things wrong.
(Software *can* be run in templates, software *can* be installed in app qubes, etc.)

Maybe this can be achieved by adding a sentence like "These are general
recommendations/best practices, other usages are possible" or something like that.


-- 
cheers,
	Holger

 ⢀⣴⠾⠻⢶⣦⠀
 ⣾⠁⢠⠒⠀⣿⡁  holger@(debian|reproducible-builds|layer-acht).org
 ⢿⡄⠘⠷⠚⠋⠀  OpenPGP: B8BF54137B09D35CF026FE9D 091AB856069AAA1C
 ⠈⠳⣄

If a monkey hoarded more bananas than it could eat, while most of the other
monkeys starved, scientists would study that monkey to figure out what the
heck was wrong with it. When humans do it, we put them on the cover of Forbes.

ninavizz:
> We want to be very careful with duplicating content, because it's almost invariably problematic. The small amount here is _probably_ okay, though.

It's only intended to be just enough to connect the narrative in user minds. Agreed wrt copious verbatim redundancy.

> 
> In the context of Qubes templates, "create" often means "clone from existing." From what you have here, a reader might conclude that customizing templates is extremely difficult, which would not be true. It will probably require further explanation to be really clear about this.

Things need to be discussed and presented to folks on "basic" vs "advanced" pages, to align more with common mental models folks can relate to, than domain-specific mental models. Eg: I was stuck for a week, trying to figure-out how to "create" a Debian-9 Template on my system that did not have one. It is totally unclear to a new user, as things are, that the primary path is downloading or cloning, and the secondary path is the fancy stuff. Totally kosher giving this another spin, but "create" is a very loaded word I'd rather not confuse an intro-level doc page with, in speaking to the primary path. If that makes sense.

> > - Tweaked a few of the headers throughout
> 
> This is going to break a bunch of links. :/
> 
> Why is it better to repeat the word "template" in subheadings so many times on a page already called "Templates"?

Well shoot, that sucks. :/

Because it makes finding things when quickly skimming the page,  easier. The current formatting works for people who carefully read a whole page—not for folks opening a page, then quickly skimming and looking for phrases that resonate with them to find what they need.

Do we want to help folks find what they need, quickly, or to punish them with frustration for not having established documentation consumption habits we believe they should follow? The latter just feels really crummy, to me.

The broken link thing has me not wanting to do this right now, but as a best practice for usable text-based content to facilitate skimming, I'd like to encourage it, going forward.

> >   - Used Title Case to boost header/topic discoverability among page-skimmers.
> 
> This is not a good reason to use title case. 

Usability isn't a good reason to do something? This is a usability best practice.

> we've already discussed changing the page titles to sentence case.

Yes, page titles have very different usability needs from sub-headers within pages that are 90% text-based.

ninavizz:
@h01ger Yes, yes, yes. The "Getting Started" page just got a total rewrite (thx to Andrew!), and could totally use some settling-in. Thx for the observation and suggestion!

andrewdavidwong:
> Because it makes finding things when quickly skimming the page, easier. The current formatting works for people who carefully read a whole page—not for folks opening a page, then quickly skimming and looking for phrases that resonate with them to find what they need.

Still not seeing how this helps. If anything, the shorter headings are easier to read and convey the same information as the longer headings. The longer headings actually take longer to read because of the redundant words. They also make the URLs look a bit silly

> Do we want to help folks find what they need, quickly, or to punish them with frustration for not having established documentation consumption habits we believe they should follow? The latter just feels really crummy, to me.

That's a straw man and a false dichotomy.

> > > * Used Title Case to boost header/topic discoverability among page-skimmers.
> > 
> > This is not a good reason to use title case.
> 
> Usability isn't a good reason to do something? This is a usability best practice.
> 

When it conflicts with the best practice of using either sentence case or title case consistently for the same heading type throughout a set of documentation, yes. Now, if you're advocating that we use title case consistently for *all* subheadings, that would be better.

> > we've already discussed changing the page titles to sentence case.
> 
> Yes, page titles have very different usability needs from sub-headers within pages that are 90% text-based.

I would be skeptical about using sentence case for all page titles and title case for all subheadings (just as I am skeptical of the opposite). My sense is that picking *either* sentence or title case and using it consistently *throughout* is the right thing to do. Would like to see some sources backing up the pro-inconsistency position, if available.

andrewdavidwong:
PipelineRetry

ninavizz:
> Still not seeing how this helps. If anything, the shorter headings are easier to read and convey the same information as the longer headings. The longer headings actually take longer to read because of the redundant words. They also make the URLs look a bit silly.

1. Headings should not be shaped such that they make for good URLs. Their objective, should be to facilitate user skimming, first, and thorough-reading, second. Far fewer people sit down to linearly "read" documentation, it needs to function primarily as reference material. 
2. I am sharing what I have observed in many years of user research and ux design. Humans are not logical creatures—so consistently applying what feels logical against what I've learned in the field, is often going to return conflicting results. 

The whole point of the UX discipline, is to move away from _pure logic_ motivating everything, and to move towards how humans really behave—as evidenced in user research. 

I appreciate that to your very logic-rooted POV, this feels illogical. You are also speaking to a process of thoroughly reading everything—not coming to a page with a thing in the back of your mind that you're looking for, and hoping to skim the entire page in 10 seconds or less to find what you need. 

> I would be skeptical about using sentence case for all page titles and title case for all subheadings (just as I am skeptical of the opposite). My sense is that picking either sentence or title case and using it consistently throughout is the right thing to do. Would like to see some sources backing up the pro-inconsistency position, if available.

"Sources" are not available, unfortunately. I am conveying what I have learned over many, many years of practice. UX practitioners rarely write whitepapers, and our best practices also tend to be cultivated in practice and then shared in talks and articles—not in canonical resources. Which selfishly suits "us" well for our own day-to-day, but then suits our projects horribly by not giving us solid reference material. Design has also traditionally been practiced in a "dictatorship/expert" sort of a model, with executives telling middle management to do what the designer says—and not collaboratively across disciplines. So, materials to support those of us working collaboratively with non-designers, are currently few and far between.

I sent you to some links for books a while ago, about Information Architecture. I'd recommend reading anything by Abby Covert, or from government digital service agencies about IA, to learn more about general best practices. UX is a "craft" practice, not an academic or research discipline—and is also a very young practice, relative to others.

ninavizz:
> I would be skeptical about using sentence case for all page titles and title case for all subheadings (just as I am skeptical of the opposite). My sense is that picking either sentence or title case and using it consistently throughout is the right thing to do. Would like to see some sources backing up the pro-inconsistency position, if available.

For users, page titles and subheadings serve a radically different purpose. That is how/why this matters. Please try to see that.

andrewdavidwong:
This argument for using sentence case sounds pretty convincing to me:

https://www.sallybagshaw.com.au/articles/sentence-case-v-title-case/

What do you think?

user/templates/templates.md
```diff
@@ -83,7 +80,7 @@ developers do not test them.
 * [Gentoo](/doc/templates/gentoo/)
 * [Gentoo Minimal](/doc/templates/minimal/)
 
-## Installing
+## Installing Templates
```
unman, line 83: Certainly, include Kali.
ParrotOS and Mint templates are also available.

user/templates/templates.md
```diff
@@ -132,15 +129,15 @@ After installing a fresh template, we recommend performing the following steps:
...
+## Updating Templates
 
 Please see [How to Update](/doc/how-to-update/).
 
 ## Installing Software
 
 Please see [How to Install Software](/doc/how-to-install-software).
 
-## Uninstalling
+## Uninstalling Template
```
unman, line 140: Cant we say "Removing"?
"Templates" or "a Template"

user/templates/templates.md
```diff
@@ -132,15 +129,15 @@ After installing a fresh template, we recommend performing the following steps:
...
```
unman, line 141: You should make it clear that you cannot remove a template that is used by an app qube, (including e.g. mgmt-disp-vm). There you have to switch the template before attempting removal

PR #1180, opened 2021-07-20
Add another possible reason for "WARNING: This key is not certified...

dksmiffs:
...with a trusted signature! There is no indication that the signature
belongs to the owner."

With this edit, I'm aiming to assist the beginner reader who walked the
following breadcrumbs:

Should I trust this website?
==> verify the PGP signatures on the commits and/or tags

Detailed steps suggested by the docs along these breadcrumbs:
1) `git clone git@github.com:QubesOS/qubesos.github.io.git`
2) Verify the PGP sigs on the commits and/or tags
  a) get properly validated GPG keys (available in the Qubes Security Pack)
    i.   `git clone https://github.com/QubesOS/qubes-secpack.git`
    ii.  `gpg --import qubes-secpack/keys/*/*`
    iii. Verify/trust the QMSK (details given on the page)
  b) `cd qubesos.github.io`
  c) `git verify-commit 45ca80e8` (one of Andrew's recent commits)

RESULT: The WARNING pops up, because the user has not yet ultimately PGP
  trusted Andrew's E11D 15C6 D204 3576 9FFA  A456 8CE1 3735 2A01 9A17
  key.

### DIFF
--- a/project-security/verifying-signatures.md
+++ b/project-security/verifying-signatures.md
@@ -595,10 +595,13 @@ first.](#3-verify-your-qubes-iso))
 ### Why am I getting "WARNING: This key is not certified with a trusted
 signature! There is no indication that the signature belongs to the owner."?
 
-Either you don't have the [Qubes Master Signing
-Key](#1-get-the-qubes-master-signing-key-and-verify-its-authenticity), or you
-didn't [set its trust level
-correctly](#1-get-the-qubes-master-signing-key-and-verify-its-authenticity).
+Several possibilities:
+* you don't have the [Qubes Master Signing
+Key](#1-get-the-qubes-master-signing-key-and-verify-its-authenticity)
+* you didn't [set its trust level
+correctly](#1-get-the-qubes-master-signing-key-and-verify-its-authenticity)
+* in the case of verifying a Git commit or tag, you haven't yet chosen to
+place ultimate PGP trust in the individual signer's key
 
 ### Why am I getting "X signature not checked due to a missing key"?
 


project-security/verifying-signatures.md
```diff
@@ -595,10 +595,13 @@ first.](#3-verify-your-qubes-iso))
...
-Key](#1-get-the-qubes-master-signing-key-and-verify-its-authenticity), or you
-didn't [set its trust level
-correctly](#1-get-the-qubes-master-signing-key-and-verify-its-authenticity).
+Several possibilities:
+* you don't have the [Qubes Master Signing
+Key](#1-get-the-qubes-master-signing-key-and-verify-its-authenticity)
+* you didn't [set its trust level
+correctly](#1-get-the-qubes-master-signing-key-and-verify-its-authenticity)
+* in the case of verifying a Git commit or tag, you haven't yet chosen to
+place ultimate PGP trust in the individual signer's key
```
andrewdavidwong, line 604: The third possibility you just added is actually the same as the second possibility that was already present. Placing ultimate trust in a key is the same as setting the trust level of that key (to "ultimate").

I'm guessing this wasn't clear (or else you wouldn't have been motivated to add the third reason), so perhaps we just need to clarify that existing second reason.

andrewdavidwong (changes requested):

andrewdavidwong:
(On a separate note: Reviewing this PR, I just realized that my prior line wrapping broke a bunch of these headings. Fixing that separately.)

project-security/verifying-signatures.md
dksmiffs: My experience:

1. I **did** have the QMSK
2. I **did** set its (meaning the QMSK's) trust level correctly to "ultimate"
3. when verifying a Git commit by Andrew, before giving Andrew's key explicit "ultimate" trust, I got the warning

So in this sequence, the second reason wasn't the cause for the warning I received.  It was the third reason.  Ultimately trusting your doc signing key eliminated the warning for me.

andrewdavidwong: It appears that's because my doc signing key is not signed by the QMSK (nor are any of the other doc signing keys). I guess that's our policy. Only some of the keys in the secpack are signed by the QMSK.

I'll document this as part of the revamp I'm doing for this page.

PR #1183, opened 2021-07-25
add extra case for installation freezing

deeplow:
If installing in external SSD sometimes and when the user forgets to disable sys-usb the initial setup also
fails after setup networking.

[Another user and myself](https://forum.qubes-os.org/t/aborts-installation-at-installing-network-drivers-and-then-enters-into-a-black-screen-with-a-single-curser-in-the-top-left-corner/5238/4) have experience this issue after installing Qubes on an external SSD and forgetting to disable sys-usb during the initial setup. @hyperfekt was this your case as well?


andrewdavidwong:
Typo: do -> to

I wonder if there's a way for the installer to detect and warn the user about this automatically. Might be worth an enhancement issue.

deeplow:
> I wonder if there's a way for the installer to detect and warn the user about this automatically. Might be worth an enhancement issue.

Done: https://github.com/QubesOS/qubes-issues/issues/6798

PR #1184, opened 2021-07-26
Don't repeat prereqs in 'How to verify the [...] hash values' section...

dksmiffs:
...Also make other minor improvements.

### DIFF
--- a/project-security/verifying-signatures.md
+++ b/project-security/verifying-signatures.md
@@ -41,7 +41,7 @@ decision of whether to trust any given party is beyond the scope of digital
 signatures. It's more of a social and political decision.
 
 Once we make the decision to trust certain parties, digital signatures are
-useful, because they make it possible for us to limit our trust to only those
+useful, because they make it possible for us to limit our trust only to those
 few parties we choose and not to worry about all the bad things that can happen
 between us and them, e.g., server compromises (qubes-os.org will surely be
 compromised one day, so [don't blindly trust the live version of this
@@ -656,11 +656,11 @@ the arguments to `gpg2`. (The signature file goes first.)
 
 There are several possibilities:
 - You don't have the [Qubes Master Signing
-  Key (QMSK)](#how-to-import-and-authenticate-the-qubes-master-signing-key).
-- You have [not set the trust level of the QMSK
+  Key](#how-to-import-and-authenticate-the-qubes-master-signing-key).
+- You have [not set the Qubes Master Signing Key's trust level
   correctly.](#how-to-import-and-authenticate-the-qubes-master-signing-key)
-- In the case of a key that is not directly signed by the QMSK,
-  you have [not set that key's trust level
+- In the case of a key that is not directly signed by the Qubes Master Signing
+  Key, you have [not set that key's trust level
   correctly.](#how-to-verify-signatures-on-git-repository-tags-and-commits)
 
 ### Why am I getting "X signature not checked due to a missing key"?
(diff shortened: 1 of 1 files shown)


project-security/verifying-signatures.md
```diff
@@ -41,7 +41,7 @@ decision of whether to trust any given party is beyond the scope of digital
 signatures. It's more of a social and political decision.
 
 Once we make the decision to trust certain parties, digital signatures are
-useful, because they make it possible for us to limit our trust only to those
+useful, because they make it possible for us to limit our trust to only those
```
andrewdavidwong, line 44: This is less precise.

project-security/verifying-signatures.md
```diff
@@ -480,22 +480,18 @@ malicious ISO, computed the hash values for that malicious ISO, and replaced
...
+```shell_session
+$ gpg2 -v --verify Qubes-RX-x86_64.iso.DIGESTS
+gpg: armor header: Hash: SHA256
+gpg: armor header: Version: GnuPG v2
+gpg: original file name=''
+gpg: Signature made Tue 20 Sep 2016 10:37:03 AM PDT using RSA key ID 03FA5082
+gpg: using PGP trust model
+gpg: Good signature from "Qubes OS Release X Signing Key"
+gpg: textmode signature, digest algorithm SHA256
+```
```
andrewdavidwong, line 494: Why not repeat those steps here? Some people will surely miss them and ask silly questions as a result. In fact, this has already happened several times.

project-security/verifying-signatures.md
```diff
@@ -660,11 +656,11 @@ the arguments to `gpg2`. (The signature file goes first.)
...
 - You don't have the [Qubes Master Signing
-  Key](#how-to-import-and-authenticate-the-qubes-master-signing-key).
-- [You have not set the Qubes Master Signing Key's trust level
+  Key (QMSK)](#how-to-import-and-authenticate-the-qubes-master-signing-key).
+- You have [not set the trust level of the QMSK
   correctly.](#how-to-import-and-authenticate-the-qubes-master-signing-key)
-- [In the case of a key that is not directly signed by the Qubes Master Signing
-  Key, you have not set that key's trust level
+- In the case of a key that is not directly signed by the QMSK,
+  you have [not set that key's trust level
```
andrewdavidwong, line 663: These were intentional, as a reader might find these FAQ entries without having been introduced to the acronym. Also, the people most likely to need these are the easily confused, and they may not pick up on the acronym. I'm open to be convinced otherwise, though.

andrewdavidwong (changes requested):

project-security/verifying-signatures.md
dksmiffs: We did use the acronym throughout this page, but I didn't notice before that this would've been its first introduction into the FAQ.  I agree with your comment, I'm changing it back.

project-security/verifying-signatures.md
dksmiffs: Those steps are listed as prereqs in the narrowest linkable scope, so if they're not going to read those, then they're probably equally likely not to read these.  Also, the flow of repeating those steps here reads awkwardly - we're introduced with "we can use GPG to verify it from the command line:", and then there's an unspoken "wait, before you do that, did you remember to do these prereqs?"

andrewdavidwong: I do think some people will miss it or forget without the reminder, but perhaps it's best not to let such readers' shortcomings unduly influence the writing. Let's try it your way. 🙂

PR #1187, opened 2021-08-07
Update hardware-related docs

andrewdavidwong:
- Add links to "just works" list
- Improve organization
- Standardize orthography
- Wrap text
- Remove very old ML link
- Minor wording improvements

### DIFF
--- a/user/hardware/certified-hardware.md
+++ b/user/hardware/certified-hardware.md
@@ -19,8 +19,7 @@ We aim for these vendors to be as diverse as possible in terms of geography, cos
   We take no responsibility for any vendor's manufacturing, shipping, payment, or other practices; nor can we control whether physical hardware is modified (whether maliciously or otherwise) <i>en route</i> to the user.
 </div>
 
-There are also other hardware models on which we have tested Qubes OS.
-See [Hardware Testing](/doc/hardware-testing/) for details.
+You may also be interested in the community-curated list of [laptops that "just work" with Qubes OS](https://forum.qubes-os.org/t/laptops-that-just-work-with-qubes-os-r4-0-4/5560).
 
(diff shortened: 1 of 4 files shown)


andrewdavidwong:
FYI @SvenSemmler, @deeplow

andrewdavidwong:
Cross-link for future reference:
https://forum.qubes-os.org/t/laptops-that-just-work-with-qubes-os-r4-0-4/5560/4?u=adw

andrewdavidwong:
Oh, crap. I just realized that `r4-0-4` is hard-coded into the URL. :frowning_face: 

However, I see that the URL will redirect to the correct thread as long as the number ("5560") is there, so I suppose I'll just edit these URLs.

SvenSemmler:
@andrewdavidwong, @deeplow: please use https://forum.qubes-os.org/t/5560

-- 

andrewdavidwong:
> @andrewdavidwong, @deeplow: please use https://forum.qubes-os.org/t/5560
> 

I considered that too, but the current URL at least gives some preview of the title. There is no difference in functionality. Do we expect to change even the part that remains? I suppose it's redundant with the anchor text, but then that would go out of date too.

andrewdavidwong:
@SvenSemmler: Ok, I've pushed an update (https://github.com/QubesOS/qubes-doc/commit/d75a0dcfe27a02b554adef70f60ac99439de8747) that uses the shorter URL (least chance of things going wrong in the future that way, I suppose). I also decided to make the anchor text more generic, since we might decide to change the thread title in the future. For example, we might start including non-laptop hardware. For another example, someone might complain that they don't all "just work" (since you have to do "minor" things to make some of them work), so we might change the title in response to that, etc.

PR #1191, opened 2021-08-31
Add debian-security rewrite

SvenSemmler:
When upgrading from buster to bullseye the 'debian-security buster/updates' must change into 'debian-security bullseye-security'. I tested this with my 10+ debian-minimal templates on R4.0.4, but see no reason this should be any different for the full Debian template. Needs review by @unman.

I used 'bullseye' instead of '\<new-name\>' so it would silently fail in case of any other upgrade.

### DIFF
--- a/user/templates/debian/debian-upgrade.md
+++ b/user/templates/debian/debian-upgrade.md
@@ -38,6 +38,19 @@ In general, upgrading a Debian template follows the same process as [upgrading a
 [user@debian-<new> ~]$ sudo apt dist-upgrade
 [user@dom0 ~]$ qvm-shutdown debian-<new>
 ```
+**Note:** when upgrading from Debian 10 "buster" to Debian 11 "bullseye" an additional `sed` command is needed before running `sudo apt update`: 
+
+```
+[user@dom0 ~]$ qvm-clone debian-<old> debian-<new>
+[user@dom0 ~]$ qvm-run -a debian-<new> gnome-terminal
+[user@debian-<new> ~]$ sudo sed -i 's/<old-name>/<new-name>/g' /etc/apt/sources.list
+[user@debian-<new> ~]$ sudo sed -i 's/debian-security bullseye\/updates/debian-security bullseye-security/g' /etc/apt/sources.list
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
> I used 'bullseye' instead of '<new-name>' so it would silently fail in case of any other upgrade.

Why is that desirable?

SvenSemmler:
On 8/31/21 2:20 PM, Andrew David Wong wrote:
>> I used 'bullseye' instead of '<new-name>' so it would silently fail in case of any other upgrade.
> Why is that desirable?

Presumably the instructions as they where *before* my change worked for Jessie->Stretch and Stretch->Buster, because the security repo was under:

	../debian-security buster/updates


... or "jessie/updates" or "stretch/updates". Going forward it is hosted at:


	../debian-security bullseye-security


.. and presumably will continue to be under "bookworm-security" and "trixie-security" etc.

So the actual rewrite is exclusively needed in the buster->bullseye transition.

-- 

SvenSemmler:
> would silently fail

A better forumulation would have been "have no effect" as it won't actually fail, but simply not match and hence not effect any change.

SvenSemmler:
I just realized that there is a "release-specific notes" section at the bottom of the page. Maybe this should be mentioned there?

andrewdavidwong:
> I just realized that there is a "release-specific notes" section at the bottom of the page. Maybe this should be mentioned there?

Ah, yes, that's what that section is for (as you can see from the other subsections for prior Debian releases). Might be better to move this down there, creating a new section for Debian 11. I don't think it's a good idea to duplicate the content in both places. Better to pick one (probably "release-specific notes") and point to it from the other (e.g., "If upgrading to Debian 11, see [here].").

unman:
@SvenSemmler  Are you going to make the change Andrew suggested?

SvenSemmler:
> @SvenSemmler  Are you going to make the change Andrew suggested?

I will some time today. Sorry for the delay.

-- 

unman:
On Wed, Sep 15, 2021 at 07:54:24AM -0700, Sven Semmler wrote:
> 
> > @SvenSemmler  Are you going to make the change Andrew suggested?
> 
> I will some time today. Sorry for the delay.
> 

I wasn't complaining, just asking.

SvenSemmler:
@unman please review #1197 instead.

PR #1193, opened 2021-09-07
Update how-to-update.md

ninavizz:
Per https://github.com/QubesOS/qubes-issues/issues/#6871 added a blurb in the CLI section, to advise users that 1. The terminal in the update-vm will in fact open, and 2. Where the logs will be saved to. My notation on two may be incorrect, so pls check?

### DIFF
--- a/user/how-to-guides/how-to-update.md
+++ b/user/how-to-guides/how-to-update.md
@@ -76,9 +76,9 @@ Advanced users may also be interested in learning [how to enable the
 testing repos](/doc/testing/).
 
 ### Qubes 4.1
-The recommendedation and relevant SALT forumulae and command from 4.0 remain the same.
+The recommendedation, commands, and SALT forumulae from 4.0 remain the same.
 
-However, backend updates to 4.1 now automatically open a terminal window in your update vm (the qube that proxies your updates, or the qube you are seeking to update if you have no proxies setup). The log from that terminal will be saved back to the VM that was updated.
+dom0 updates in 4.1 will demonstrate a new behavior. A terminal window is now opened in your update vm, from which you may observe the progress of tasks and logs as they are compiled. At the end of the process, logs are sent back to dom0. Any y/n commands you may need to respond to, should be responded to from your dom0 terminal window. 
 
 ## Upgrading to avoid EOL
 
--- a/user/how-to-guides/how-to-update.md
+++ b/user/how-to-guides/how-to-update.md
@@ -72,7 +72,9 @@ There are two Salt formulae and two corresponding `qubesctl` commands:
  - [`update.qubes-dom0`](/doc/salt/#updatequbes-dom0)
  - [`update.qubes-vm`](/doc/salt/#updatequbes-vm)
 
-In addition, advanced user may be interested in learning [how to enable the
+Bear in mind that when updates are performed from the command line, a terminal will be automatically opened in your update vm (the qube that proxies your updates, or the qube you are seeking to update if you have no proxies setup). The log from that terminal will be saved back to the VM that was updated.
+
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
Since this only applies to 4.1, I'll have to figure out the best way to handle this. See:

https://www.qubes-os.org/doc/documentation-style-guide/#release-specific-documentation

DemiMarie:
Does this only apply to dom0 updates?

ninavizz:
@andrewdavidwong I went ahead and took a stab at an update, mostly following the styleguide. Unfortunately, users tend to visually snapshot redundant text, and skip-over sections that appear redundant of text they've already read. Because of that, I did not want to 1:1 follow the styleguide, and create a block of text that's 80-90% redundant with what is up above. 

I appreciate your rationale in the docs styleguide for retaining histories per version-specific guidance, with the 100% redundancy guidance. It just works against usability with how humans consume written content; a bummer of a phenomenon long observed in user research.

ninavizz:
@DemiMarie Err... I'd thought you were the expert on that? Andrew and I are just making updates to the docs per the original issue this pointed to. If my text is incorrect, pls let me know! :D

andrewdavidwong:
> Unfortunately, users tend to visually snapshot redundant text, and skip-over sections that appear redundant of text they've already read. Because of that, I did not want to 1:1 follow the styleguide, and create a block of text that's 80-90% redundant with what is up above.
> 
> I appreciate your rationale in the docs styleguide for retaining histories per version-specific guidance, with the 100% redundancy guidance. It just works against usability with how humans consume written content; a bummer of a phenomenon long observed in user research.

That's part of the reason we *don't* want to do it this way anymore. We want to have release-specific docs instead: https://github.com/QubesOS/qubes-issues/issues/5308

Might try to implement this for 4.1.

DemiMarie:
> @DemiMarie Err... I'd thought you were the expert on that? Andrew and I are just making updates to the docs per the original issue this pointed to. If my text is incorrect, pls let me know! :D

Sorry about that!  There are two different concepts here:

1. The updates proxy, which is implemented as an HTTP proxy tunneled over qrexec.  This is used so that `dnf`, `apt`, and other package managers can obtain packages from the Internet without template qubes having to have a network connection.
2. The UpdateVM, which is what dom0 uses to obtain updates.  It uses a completely different architecture: the metadata is not processed in dom0, but rather in the UpdateVM, which computes dependencies and sends the appropriate packages to dom0 over qrexec (`qubes.ReceiveUpdates` service).  dom0 then canonicalizes the packages (using `rpmcanon`), checks their signatures (both in `rpmcanon` and again via `rpmkeys -K`), and finally calls `createrepo_c` to create a whole new package repository of already-verified packages.  It is *that* repository that `dnf` in dom0 actually pulls from.  This is the case that pops up a terminal in the UpdateVM when run interactively, so that the user has a progress bar to see the UpdateVM downloading packages (the slow part). 
 Once the packages are downloaded, the remaining steps in the process are quite fast.

DemiMarie:
Logs aren’t actually sent back to dom0.  Instead, dom0 recomputes the list of tasks on its own, and it is that new list that is displayed in the dom0 terminal window.

andrewdavidwong:
@ninavizz, could you revise this PR so that it doesn't reference different Qubes releases and instead just speaks to the user on Qubes 4.1 (so, no past/future or version number references; just assume the user us on 4.1).

andrewdavidwong:
> @ninavizz, could you revise this PR so that it doesn't reference different Qubes releases and instead just speaks to the user on Qubes 4.1 (so, no past/future or version number references; just assume the user is on 4.1).

Actually, scratch that. The way you've done it here is already correct. My mistake.

PR #1194, opened 2021-09-10
correct Qubes-Whonix support status for Qubes R4.1

adrelanos:
https://github.com/QubesOS/qubes-issues/issues/6890


andrewdavidwong:
Already added 16. Will remove 15 when it reaches EOL.

PR #1195, opened 2021-09-12
Describe qubes-builder usage as is necessary today.

ydirson:
This reflects in the doc the consequences of issue #6522, and
describes how to build a kernel today.  It also gives some hints for
 #6486 as to what to improve in next generation of qubes-builder.

The kernel build should be linked from main documentation page on the
website, but sadly that page is not in this repo.

I flagged this PR as "Draft", as this won't work for possibly a new bug: `make qubes-dom0`
always rebuilds `vmm-xen-dom0`, only to fail on `core-libvirt-dom0` because of missing sig.
So after `make sign-dom0` a new `make qubes-dom0` run will just regenerate the just-signed
rpms with new unsigned ones and fail again.

Also it is not clear to me if we want to unset `NO_SIGN` by default, which would have an impact on this doc.

### DIFF
--- a/developer/building/qubes-builder.md
+++ b/developer/building/qubes-builder.md
@@ -47,20 +47,33 @@ The build system creates build environments in chroots and so no other packages 
 All files created by the build system are contained within the qubes-builder directory.
 The full build requires some 25GB of free space, so keep that in mind when deciding where to place this directory.
 
+One additional useful requirement is that 'sudo root' must work without any prompt, which is default on most distros (e.g. 'sudo bash' brings you the root shell without asking for any password).
+This is important as the builder needs to switch to root and then back to user several times during the build process.
+
 The build system is configured via builder.conf file.
 You can use the setup.sh script to create and modify this file.
 Alternatively, you can copy the provided default builder.conf, and modify it as needed, e.g.:
 
 ```shell
 cp example-configs/qubes-os-master.conf builder.conf
-# edit the builder.conf file and set the following variables:
-NO_SIGN=1
 ```
 
-One additional useful requirement is that 'sudo root' must work without any prompt, which is default on most distros (e.g. 'sudo bash' brings you the root shell without asking for any password).
-This is important as the builder needs to switch to root and then back to user several times during the build process.
+The default configurations (as of `qubes-builder#master@b9f0322e4f3`)
+disables signing of generated packages, through `NO_SIGN ?= 1`.  This
+is [known not to
+work](https://github.com/QubesOS/qubes-issues/issues/6522) if you want
+to build a package that depends on another one you also build (that is
+necessary to build only some packages, for Fedora-based VMs, but some
+prominent ones like `linux-kernel` do have dependencies and cannot be
+built with `NO_SIGN=1`), and package signing must be enabled, by
+setting this variable to an empty value (0 will not work), by
+commenting the line or clearing it:
 
-Additionally, if building with signing enabled (NO\_SIGN is not set), you must adjust `\~/.rpmmacros` file so that it points to the GPG key used for package signing, e.g.:
+```makefile
+NO_SIGN =
+```
+
+If building with signing enabled (`NO_SIGN` is not set), you must adjust `\~/.rpmmacros` file so that it points to the GPG key used for package signing, e.g.:
 
 ```bash
 %_signature gpg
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
FYI, you can open a PR as a draft rather than manually adding "WIP" to the title. This is generally less prone to misunderstanding, since it makes it clear to everyone at a glance when you do and don't want your PR to be reviewed and merged.

andrewdavidwong:
PipelineRetry

developer/building/qubes-builder.md
```diff
@@ -47,20 +47,28 @@ The build system creates build environments in chroots and so no other packages
...
 ```
 
-One additional useful requirement is that 'sudo root' must work without any prompt, which is default on most distros (e.g. 'sudo bash' brings you the root shell without asking for any password).
-This is important as the builder needs to switch to root and then back to user several times during the build process.
+The default configurations (as of `qubes-builder#master@b9f0322e4f3`)
+disables signing of generated packages, through `NO_SIGN ?= 1`.  This
+is [known not to
+work](https://github.com/QubesOS/qubes-issues/issues/6522) if you want
+to build a package that depends on another one you also build
+(eg. `linux-kernel`), and package signing must be enabled, by setting
```
unman, line 66: You should make it explicit that this only affects Fedora based builds, and some packages. Users wanting to modify many packages, and other templates, don't need to do this.

developer/building/qubes-builder.md
```diff
@@ -47,20 +47,28 @@ The build system creates build environments in chroots and so no other packages
...
+this variable to an empty value (0 will not work), by commenting the
+line or writing `NO_SIGN=` instead.
+
 
-Additionally, if building with signing enabled (NO\_SIGN is not set), you must adjust `\~/.rpmmacros` file so that it points to the GPG key used for package signing, e.g.:
+Additionally, if building with signing enabled (`NO_SIGN` is not set), you must adjust `\~/.rpmmacros` file so that it points to the GPG key used for package signing, e.g.:
```
unman, line 71: "Additionally" is unnecessary

developer/building/qubes-builder.md
```diff
@@ -71,6 +79,10 @@ Additionally, if building with signing enabled (NO\_SIGN is not set), you must a
 It is also recommended that you use an empty passphrase for the private key used for signing.
 Contrary to a popular belief, this doesn't affect your key or sources security -- if somebody compromised your system, then the game is over anyway, whether you have used an additional passphrase for the key or not.
 
+You will also need include your key in each chroot used for the build
+(which for this reason have to be generated first), by running `rpm
+--import` in each of those chroots.
```
unman, line 84: I think this is clumsy.
You will need to generate each chroot used for the build, and include your key in each, by running `rpm --import`

This has the advantage of matching the instructions you provide. Again, specify that this applies only to Fedora based chroots, or provide instructions for other templates.

unman (changes requested):

developer/building/qubes-builder.md
```diff
@@ -47,20 +47,28 @@ The build system creates build environments in chroots and so no other packages
...
+
 The build system is configured via builder.conf file.
 You can use the setup.sh script to create and modify this file.
 Alternatively, you can copy the provided default builder.conf, and modify it as needed, e.g.:
 
 ```shell
 cp example-configs/qubes-os-master.conf builder.conf
-# edit the builder.conf file and set the following variables:
-NO_SIGN=1
 ```
```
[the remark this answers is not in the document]
ydirson: ok, reorganizing the next paragraph for this

developer/building/qubes-builder.md
ydirson: AFAICT and correct me if missed something:
- the whole NO_SIGN concept today only impacts Fedora builds, by pulling `rpm-sign` and impacting `make sign`
- the `linux-yum` and `installer-qubes-os` are the only components making use of it, with code to allow unsigned repositories, but this is apparently not working today, overriden by the general setting forcing their use

right?

developer/building/qubes-builder.md
ydirson: Looks like signing of debian packages will have to be added separately :)

unman: >  signing of debian packages 

You'll know that this is rarely done in Debian, and there is not a standard on how to do it.

andrewdavidwong:
Just checking in because I haven't seen any communication on this PR in a while. @ydirson, what's the current status from your side?

unman:
Absent any communication, and with the move to bulderv2 I am closing this PR.

ydirson:
Ah, looks like I had missed the question :)

PR #1206, opened 2021-11-13
Clarification Points

ninavizz:
- Added a table to the "Debian Full" page, to manage user expectations around which releases have templates available for them
- Edited beginning of main "Templates" page, to explain templates in more human terms and to explicitly state they're _installed packages_ built by developers, not downloaded distros. Because—well, yes.

### DIFF
--- a/user/templates/debian/debian.md
+++ b/user/templates/debian/debian.md
@@ -12,14 +12,14 @@ title: Debian templates
 ---
 
 The Debian [template](/doc/templates/) is an officially [supported](/doc/supported-releases/#templates) template in Qubes OS.  
-This page is about the standard (or "full") Debian template.
-For the minimal version, please see the [Minimal templates](/doc/templates/minimal/) page.
+This page is about the standard (or "Full") Debian template.
+For the Minimal version, please see the [Minimal templates](/doc/templates/minimal/) page.
 There is also a [Qubes page on the Debian Wiki](https://wiki.debian.org/Qubes).
 
 
-| Release Version        |  Initial Release Date & Status*       |  Does a Qubes OS "full" Template yet exist for this? |  Does a Qubes OS "minimal" Template yet exist for this? |
+| Release Version        |  Initial Release Date & Status*       |  "Full" Qubes OS template available? |  "Minimal" Qubes OS template available? |
 | :---                   |  :---                                 |  :---                                                |  :---                                                  |
-| Debian 11 "bullseye"   | 14 Aug 2021, current stable           |  No                                                  |  No                                                     |
+| Debian 11 "bullseye"   | 14 Aug 2021, current stable           |  —Not Yet—                                           |  —Not Yet—                                                     |
 | Debian 10 "buster"     | 06 Jul 2019, current oldstable            |  Yes                                             |  Yes                                                    |
 | Debian 9 "stretch"     | 27 Jun 2017, oldoldstable under LTS       |  Decomissioned                                   | Decomissioned                                           |
 | Debian 8 "jessie"      | 26 Apr 2015, archived under Extended LTS  |  Decomissioned                                   | Decomissioned                                           |
--- a/user/templates/debian/debian.md
+++ b/user/templates/debian/debian.md
@@ -16,7 +16,6 @@ This page is about the standard (or "full") Debian template.
 For the minimal version, please see the [Minimal templates](/doc/templates/minimal/) page.
 There is also a [Qubes page on the Debian Wiki](https://wiki.debian.org/Qubes).
 
-Building a template in Qubes OS is not a trivial process. Below is a (tally?) of known Debian releases, and the status of their Qubes OS templates. Release-specific installation notes, are at the bottom of the page below the general installation notes. 
 
 | Release Version        |  Initial Release Date & Status*       |  Does a Qubes OS "full" Template yet exist for this? |  Does a Qubes OS "minimal" Template yet exist for this? |
 | :---                   |  :---                                 |  :---                                                |  :---                                                  |
--- a/user/templates/debian/debian.md
+++ b/user/templates/debian/debian.md
@@ -11,10 +11,20 @@ ref: 134
 title: Debian templates
 ---
 
-The Debian [template](/doc/templates/) is an officially [supported](/doc/supported-releases/#templates) template in Qubes OS.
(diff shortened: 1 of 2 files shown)


user/templates/debian/debian.md
```diff
@@ -11,11 +11,20 @@ ref: 134
 title: Debian templates
 ---
 
-The Debian [template](/doc/templates/) is an officially [supported](/doc/supported-releases/#templates) template in Qubes OS.
-This page is about the standard (or "full") Debian template.
-For the minimal version, please see the [Minimal templates](/doc/templates/minimal/) page.
+The Debian [template](/doc/templates/) is an officially [supported](/doc/supported-releases/#templates) template in Qubes OS.  
+This page is about the standard (or "Full") Debian template.
+For the Minimal version, please see the [Minimal templates](/doc/templates/minimal/) page.
```
andrewdavidwong, line 16: I don't think these terms should be capitalized.

user/templates/debian/debian.md
```diff
@@ -11,11 +11,20 @@ ref: 134
...
+| Release Version        |  Initial Release Date & Status*       |  "Full" Qubes OS template available? |  "Minimal" Qubes OS template available? |
+| :---                   |  :---                                 |  :---                                                |  :---                                                  |
+| Debian 11 "bullseye"   | 14 Aug 2021, current stable           |  —Not Yet—                                           |  —Not Yet—                                                     |
+| Debian 10 "buster"     | 06 Jul 2019, current oldstable            |  Yes                                             |  Yes                                                    |
+| Debian 9 "stretch"     | 27 Jun 2017, oldoldstable under LTS       |  Decomissioned                                   | Decomissioned                                           |
+| Debian 8 "jessie"      | 26 Apr 2015, archived under Extended LTS  |  Decomissioned                                   | Decomissioned                                           |
+| Debian 7 "wheezy"      | 04 May 2013, obsolete stable              |  Decomissioned                                   | Decomissioned
+
```
andrewdavidwong, line 27: This appears to largely duplicate the table we already have on the "supported releases" page. I propose we just link to that existing table instead. (Note that there's also Fedora to consider.)

user/templates/templates.md
```diff
@@ -13,10 +13,13 @@ title: Templates
 
 In [Getting Started](/doc/getting-started/), we covered the distinction
 in Qubes OS between where you *install* your software and where you *run* your
-software. Your software is installed in [templates](/doc/glossary/#template).
-Each template shares its root filesystem (i.e., all of its programs and system
-files) with all the qubes based on it. [App qubes](/doc/glossary/#app-qube) are
-where you run your software and store your data.
+software. Software that you use in most everyday tasks, is installed within [templates](/doc/glossary/#template).
+
+When using Qubes OS, you will typically begin a task by opening an application in an [app qube](/doc/glossary/#app-qube). App qubes are built from a *template* qube (or more simply, just *a template*). As such, app qubes they _inherit_ all data from their template—from the template's OS or distro, to installed applications, to stored files—so its "[root filesystem](https://opensource.com/life/16/10/introduction-linux-filesystems)." App qubes do not, however, _persist_ their data back to the Template.
```
andrewdavidwong, line 18: I think the phrasing here might be a bit confusing for the intended audience.

user/templates/templates.md
```diff
@@ -13,10 +13,13 @@ title: Templates
...
+
+An understanding of [Inheritance and Persistence](#inheritance-and-persistence) concepts will help inform a clear understanding of this essential concept within Qubes OS. [Standalones](/doc/glossary/#standalone) are the only qubes in the Qubes OS ecosystem that neither _persist_ nor _inherit_ their filesystems to or from other qubes. 
+
+Templates are always built by developers, and are typically categorized by which OS or Linux distro each is based on. Qubes OS officially supports Debian and Fedora templates, and our community of developers has built and made available many more. When installing Qubes OS, officially supported templates may be installed. If you wish to add additional templates, you will need to download and install them.
```
andrewdavidwong, line 22: Probably better to link to the main template page instead of listing officially supported templates here. This is the sort of duplication that's likely to lead to erroneous information if we ever add to or remove from that list, so it's best to keep it in one place.

andrewdavidwong (changes requested):

user/templates/templates.md
ninavizz: I only just found the "Supported Versions" page, and had no idea to look for 3rd party things, there. IMHO it'd make more sense to move its content, onto this page. If you'd be comfortable with that. Then Debian and Fedora specific content, there. 

Since users don't read whole sections of documentation and will often just look at the page they feel is most relevant to an immediate need, I feel that would better support usability—while also keeping things maintainable. Thoughts?

user/templates/templates.md
ninavizz: I guess that calls into question who the intended audience is—just developers, native Linux users, or also people new to Linux, as well? My own feeling is that it's currently written to support only very technical users, to the exclusion of folks unfamiliar with phrasing like "file systems" or "root." 

I feel like more technical users can do a better job working backwards to make sense of things, than more  novice users; and I only edited the introduction to make for an easier entry experience.

user/templates/debian/debian.md
ninavizz: It feels like a contradiction of our own consistency standard to speak to 3rd party products, on a Qubes docs page on "Supported Releases." That page has enough about just Qubes and dom0 on it, already—and the rest of its contents I feel would make more sense on the Templates page, and the Debian and Fedora specific pages. See other comment. Can update PR, accordingly if you'd be open to that.

user/templates/debian/debian.md
ninavizz: Cool!

andrewdavidwong:
@ninavizz, I think we already discussed all of your questions here in our other chat. Let me know if any still linger.

user/templates/debian/debian.md
maiska: @ninavizz Thak you for your suggestions! Would you like to update your PR with the current main version of the documentation so that we can continue the conversation and review the PR? :) All the best!

maiska:
@ninavizz Hello! Would you be so kind to update the PR to reflect the current state of the origin main branch while keeping the suggestions you consider important? :)

maiska:
@unman  @rapenne-s  I will list the PR for after RST doc conversion to incorporate some of the suggestions with taking time to consider them given the time passed, github issue regarding my suggestions for after RST doc conversion is coming these days

PR #1215, opened 2021-12-07
Doc: accelerated video w/o GPU on fedora-minimal

hexagonrecursion:
* Many applications will fail to run or have reduced functionality if OpenGL and/or Vulkan is not available.
* Due to how QubesOS handles GPU giving a qube access to a real GPU is not an option for some users.
* Software implementations of the OpenGL and Vulkan API exist, but are not installed by default in minimal templates. ~TODO: check if full templates have them.~

Solution: add instructions on how to install mesa-dri-drivers to the customization guide

### DIFF
--- a/user/templates/minimal-templates.md
+++ b/user/templates/minimal-templates.md
@@ -194,6 +194,9 @@ Also, there are packages to provide additional services:
   idle.
 - `qubes-mgmt-salt-vm-connector`: If you want to use salt management on the
   template and qubes.
+- `mesa-dri-drivers`: If you want to run applications (e.g. videogames)
+  that require OpenGL or Vulkan API without attaching a real video card
+  to the qube.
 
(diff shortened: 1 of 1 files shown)


hexagonrecursion:
For reference: a full Fedora 34 template has mesa-dri-drivers installed by default.

marmarek:
I don't think this list should describe generic packages, going this way we should describe most/all packages available in Fedora/Debian...
Maybe there should be some page for extended examples of custom templates based on the minimal ones? But likely at https://github.com/Qubes-Community/Contents

hexagonrecursion:
I disagree with closing this. Correct me if I'm wrong, but as far as I can tell this is required precisely because Qubes does not give DomU access to the GPU (except via PCI, which requires two GPUs - a luxury not everyone has)

PR #1219, opened 2021-12-30
add details for dom0-approved sudo

grey-olli:
chmod 755 required for new policy file,
show exact line to replace in sudoers.d/qubes

### DIFF
--- a/user/security-in-qubes/vm-sudo.md
+++ b/user/security-in-qubes/vm-sudo.md
@@ -113,6 +113,7 @@ Do not rely on this for extra security.**
 
     ```
     [root@dom0 /]# echo "/usr/bin/echo 1" >/etc/qubes-rpc/qubes.VMAuth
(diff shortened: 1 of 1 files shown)


grey-olli:
On Thu, Feb 10, 2022, 16:57 Marek Marczykowski-Górecki <
***@***.***> wrote:

> ***@***.**** requested changes on this pull request.
>
Didn't get what changes are requested:

Those ROLE= and TYPE= entries needs to stay, for the case of enabled
> SELinux (which we are working on).
>
my pull request does require only modification of usual, not extended file
permissions. W/o chmod the policy file is not executable at all, at least
I'm not aware of SELinix permission to execute a file that has no usual
execute bit set.

Or you're requesting 'Those ROLE= and...' to be added to the .md file
itself?

marmarek:
The text now suggests to drop `ROLE=` and `TYPE=` parameters.

unman:
This has been moved to community docs, now in the Forum.

PR #1224, opened 2022-01-20
check the Intel chipset as well.

egberts:


egberts:
my Dell Optiplex 790 (circa 2011j has the necessary VT-d at CPU-level (i7-2450), but no VT-d support at the motherboard chipset level (Intel Q65 Express).

It appears to run 4.0 fine but 4.1 extremely sluggishly.

marmarek:
> It appears to run 4.0 fine but 4.1 extremely sluggishly.

It may be related to qubesos/qubes-issues#7075, not lack of VT-d.

Without VT-d, sys-net / sys-usb would not work.

egberts:
In this case, sys-net appears to be working fine on Optiplex 790 with i7-2450.  

I’ll have this pull-request retracted.

andrewdavidwong:
> I’ll have this pull-request retracted.

If I've understood correctly, you'd like this PR closed, so I'll do that now.

PR #1227, opened 2022-02-02
update language re: tor user counting

mfc:
since the red line no longer exists in the graph, there's no need to explain the different tor user counting methodologies.

### DIFF
--- a/introduction/statistics.md
+++ b/introduction/statistics.md
@@ -28,11 +28,9 @@ Since the graph is updated daily, the bar for the current month will be very low
 
 We simply count the number of unique IPv4 addresses that connect to the Qubes update servers each month (except for Tor connections; see [below](#how-has-the-methodology-for-counting-tor-users-changed)).
 
-### How has the methodology for counting Tor users changed?
+### How are Tor users counted?
 
-Before, we simply counted the number of unique Tor exit node IPv4 addresses that connected to the Qubes update servers each month.
-However, this underestimated the actual number of Tor users, since many Tor users can use the same exit node.
-The new methodology is to estimate the number of Tor users as a proportion of the total number of *requests* from Tor exit nodes on the assumption that the proportion of users to requests is roughly the same for both clearnet and Tor users.
+We estimate the number of Tor users as a proportion of the total number of *requests* from Tor exit nodes on the assumption that the proportion of users to requests is roughly the same for both clearnet and Tor users.
 To be precise, the formula is:
 
 ```
--- a/introduction/statistics.md
+++ b/introduction/statistics.md
@@ -28,6 +28,8 @@ Since the graph is updated daily, the bar for the current month will be very low
 
 We simply count the number of unique IPv4 addresses that connect to the Qubes update servers each month (except for Tor connections; see [below](#how-has-the-methodology-for-counting-tor-users-changed)).
 
+Users who have manually configured to use a mirror (whether clearnet or onion mirrors) are not counted.
+
 ### How are Tor users counted?
 
 We estimate the number of Tor users as a proportion of the total number of *requests* from Tor exit nodes on the assumption that the proportion of users to requests is roughly the same for both clearnet and Tor users.


tlaurion:
I see a problem, not linked to this PR directly, but from usage of tor hidden update servers not logging basic information, which is then dismissed here for calculation of user base.

Is there a way to at least estimate that user base? Otherwise as of today there is 40k+ qubes users from, where tor hidden uodate servers are dismissed from calculation.

@marmarek? @mfc?

mfc:
that is a *very* excellent point. at minimum we should add language noting that there is likely undercounting. not sure if anything else is possible re: the onion servers

andrewdavidwong:
> since the red line no longer exists in the graph, there's no need to explain the different tor user counting methodologies.

This needs to be changed in @woju's image.

> I see a problem, not linked to this PR directly, but from usage of tor hidden update servers not logging basic information, which is then dismissed here for calculation of user base.
> 
> Is there a way to at least estimate that user base? Otherwise as of today there is 40k+ qubes users from, where tor hidden uodate servers are dismissed from calculation.
> 
> @marmarek? @mfc?

> that is a _very_ excellent point. at minimum we should add language noting that there is likely undercounting. not sure if anything else is possible re: the onion servers

I don't understand. Isn't that exactly what is explained in the very text that this PR is attempting to remove?

I also don't understand the rationale for the actual change this PR is making (which seems unrelated to the original comment at the start of this discussion).

mfc:
> > since the red line no longer exists in the graph, there's no need to explain the different tor user counting methodologies.
> 
> This needs to be changed in @woju's image.
> 
> > I see a problem, not linked to this PR directly, but from usage of tor hidden update servers not logging basic information, which is then dismissed here for calculation of user base.
> > Is there a way to at least estimate that user base? Otherwise as of today there is 40k+ qubes users from, where tor hidden uodate servers are dismissed from calculation.
> > @marmarek? @mfc?
> 
> > that is a _very_ excellent point. at minimum we should add language noting that there is likely undercounting. not sure if anything else is possible re: the onion servers
> 
> I don't understand. Isn't that exactly what is explained in the very text that this PR is attempting to remove?
> 
> I also don't understand the rationale for the actual change this PR is making (which seems unrelated to the original comment at the start of this discussion).
> 
> Edit 1: Oh, I think I see now. The idea is that "old vs. new methodology" is correlated with "before vs. after the red line," so no more red line means no more old methodology. It's not obvious to me that there's no value in explaining the old methodology anymore. However, even if we decide not to, the valuable part that this PR seems to remove is an explanation for the underestimation.
> 
> Edit 2: Actually, I don't see how the old or new text fails to explain the undercounting.

the undercounting thierry (and i) are referring to are the .onion update servers which (if i'm not mistaken) are hosted by @unman and don't have any logging and so not included at all in these statistics. please correct me if i'm wrong

https://www.qubes-os.org/news/2019/04/17/tor-onion-services-available-again/

andrewdavidwong:
> the undercounting thierry (and i) are referring to are the .onion update servers which (if i'm not mistaken) are hosted by @unman and don't have any logging and so not included at all in these statistics. please correct me if i'm wrong

I don't really know any more than you do about how the onion servers work. Perhaps @unman can fill us in.

However, the inference "don't have any logging and so not included at all in these statistics" seems questionable in light of the text (after your PR):

> We estimate the number of Tor users as a proportion of the total number of *requests* from Tor exit nodes on the assumption that the proportion of users to requests is roughly the same for both clearnet and Tor users.

In other words, the onion server hits are not counted *directly*, but they are *estimated* based on the clearnet hits (which *are* directly counted). I'm not saying this is a good or bad system, but it *does* include the onion server hits in these statistics (via the estimation method just described).

Having said that, I certainly have no opposition to providing further explanation on the page (e.g., why the count might be underestimated for whatever reason you have in mind).

mfc:
> In other words, the onion server hits are not counted _directly_, but they are _estimated_ based on the clearnet hits (which _are_ directly counted). I'm not saying this is a good or bad system, but it _does_ include the onion server hits in these statistics (via the estimation method just described).

no they are not estimated, what is estimated is how many tor users are hitting the clearnet update server.

andrewdavidwong:
> no they are not estimated, what is estimated is how many tor users are hitting the clearnet update server.

Oh, I think I finally understand what you mean. I guess this has been a problem ever since the onion update servers debuted, and no one has pointed it out until now?

mfc:
> > no they are not estimated, what is estimated is how many tor users are hitting the clearnet update server.
> 
> Oh, I think I finally understand what you mean. I guess this has been a problem ever since the onion update servers debuted, and no one has pointed it out until now?

yeah exactly. dunno if possible/desirable to be able to include them in the stats, @unman would know. otherwise at minimum i can add a line mentioning this undercounting

unman:
I don't host the onion service. It's provided by a company that supports
Qubes, for whom I sometimes work.
My understanding is that hits are not logged - they certainly are not
included in these statistics.

The estimate of Tor users is based on hits on the clearnet server
controlled by ITL coming from Tor.
Are users who use some other mirror (over Tor or clearnet) also excluded
from these numbers? I would have thought so.

unman:
I could ask for logging to be turned on, but as they provide other onion
services, I doubt this would be done. But I can ask.

unman:
Neat - but users who directly access a mirror will be excluded.

mfc:
> I could ask for logging to be turned on, but as they provide other onion services, I doubt this would be done. But I can ask.

from my perspective wouldn't hurt to ask.

> Neat - but users who directly access a mirror will be excluded.

yeah i will add language about this.

PR #1228, opened 2022-02-08
Update glossary.md

ninavizz:
- Changed all terms to "sentence case," as lowercase-only reads as quite confusing to non-developers. Exceptions are terms such as dom0 or domU where the specific use of casing or mixed alpha/numeric characters counters the use of the sentence case best practice.
- Added "Policy," as how Qubes uses Policies maps to no easy public definition.
- Added "Policies" definition. May (probably?) need Marek/Andrew special touch. Goal is to keep text as short as possible, but also as clear as possible to users of all skill levels. Did as much, to cross-reference in FAQ.

### DIFF
--- a/user/reference/glossary.md
+++ b/user/reference/glossary.md
@@ -10,13 +10,13 @@ ref: 140
 title: Glossary
 ---
 
-## Admin qube
+## admin qube
 
 A type of [qube](#qube) used for administering Qubes OS.
 
 * Currently, the only admin qube is [dom0](#dom0).
 
-## App qube
+## app qube
 
 Any [qube](#qube) that does not have a root filesystem of its own. Every app
 qube is based on a [template](#template) from which it borrows the root
@@ -27,7 +27,7 @@ filesystem.
 * Historical note: This term originally meant "a qube intended for running user
   software applications" (hence the name "app").
 
-## Disposable
+## disposable
 
 A type of temporary [app qube](#app-qube) that self-destructs when its
 originating window closes. Each disposable is based on a [disposable
@@ -37,7 +37,7 @@ See [How to Use Dispoables](/doc/how-to-use-disposables/).
 
 * Previously known as: `DisposableVM`, `DispVM`.
 
-## Disposable template
+## disposable template
 
 Any [app qube](#app-qube) on which [disposables](#disposable) are based. A
 disposable template shares its user directories (and, indirectly, the root
@@ -68,7 +68,7 @@ relative to other domains, such as direct access to most hardware.
 * The term "dom0" is a common noun and should follow the capitalization rules
   of common nouns.
 
-## Domain
+## domain
 
 In Xen, a synonym for [VM](#vm).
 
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
> Exceptions are terms such as dom0 or domU where the specific use of casing or mixed alpha/numeric characters counters the use of the sentence case best practice.

This isn't right, though. Those should be sentence case too.

ninavizz:
Question to @andrewdavidwong: So... if I do those in sentence case too, they'll show as `Dom0` and `DomU`, which is not how users visually see either name in context. Thoughts? That was literally the only reason I didn't do those, to facilitate recognition from having seen them in context. You're the docs master, so I'll follow your decision!

andrewdavidwong:
> Question to @andrewdavidwong: So... if I do those in sentence case too, they'll show as `Dom0` and `DomU`, which is not how users visually see either name in context.

But sentence case *should* be how users see those terms in context. (Same with "qube," by the way.)

We've discussed this before, but I'll reiterate. A word like "dom0," for example, is just a term of art that's short for "domain zero." It's a common noun, like "cat," as opposed to a proper noun, like "Garfield." However, since it's unfamiliar, some people don't know how to handle it, so they always capitalize it, e.g., "Open a terminal in Dom0." You've probably seen me criticize such usage by saying that "dom0" shouldn't be capitalized in those cases for the same reason "cat" shouldn't be capitalized in "Look at that Cat over there."

I suspect what has happened is that your interpretation of my criticism has led you to internalize an *overgeneralized* rule to the effect that terms like "dom0" should *never* be capitalized. Perhaps you've also noticed that some stylized names like "iPhone" always have intentionally lowercase initial letters and subconsciously assimilated "dom0" to this group. However, "iPhone" is a proper noun, while "dom0" is a common noun.

In short, there is no rule that says that terms like "dom0" should always be lowercase, but there *are* rules that say they should be capitalized consistently with other common nouns, including at the beginnings of sentences and in sentence-case lists of words.

By the way, the reason all the common nouns in this glossary are lowercase is because that's a standard practice for glossaries ([example](https://style.mla.org/formatting-a-glossary/)). However, I've also seen some glossaries with sentence-case entries, and I'm not certain there are any hard-and-fast rules about it.

ninavizz:
> I suspect what has happened is that your interpretation of my criticism has led you to internalize an overgeneralized rule to the effect that terms like "dom0" should never be capitalized. Perhaps you've also noticed that some stylized names like "iPhone" always have intentionally lowercase initial letters and subconsciously assimilated "dom0" to this group. However, "iPhone" is a proper noun, while "dom0" is a common noun.

Not at all. One, I never learned grammar or linguistics, so much of the mechanics of language are foreign to me—and as an adult, despite trying to learn these things, it's been extraordinarily difficult... so I remember some things, and not others. 1970s childhood in a hippie town. 

I'm going purely on pattern-matching what I see in the wild, and trying to match words to how users are accustomed to "seeing" them (not _reading_ them). I suppose in part, too, because of my own learning tendency to remember the visual forms of letters, and not the language of the word. If that makes any sense. TL;DR, it sounds like it'd be best for me to revert the capitalization of the majority of terms, to the exclusion of the common noun qube names—so, I'll do that! Thx, and sorry to be a bit of a scatterbrain.

PR #1229, opened 2022-02-08
Update faq.md

ninavizz:
Hey Andrew—this is kind of a lot. Trying to get some of the basic changes to existing-but-unclear or outdated questions, into incremental PRs. No rush. Curious what your thoughts are!

- Updated "What is Qubes?" to more clearly speak to less technical users, and users unfamiliar with OSS norms; while also providing a clearer answer to the specific question.
- Separated speaking to Qubes' FOSS status, as its own (following) question.
- Moved "How is Qubes different from other security solutions" up in the list
- Moved "What about other approaches to security?" up in the list
- Killed "Why does Qubes use Xen instead of KVM or some other hypervisor?" question, because it refernces the 0.3 architecture spec—which, even if that information is still correct, feels horribly out of date. 
- Killed "What about safe languages and formally verified microkernels" question, as it directs to an article composed in 2010 for Qubes 0.3
- Killed "What is the concept behind Qubes" as it's redundant with "How does Qubes OS provide security?" (both speak to security by compartmentalization)
- Killed "Could you please make my preference the default?" as answer sounds scoldy... and, imho anybody asking that in a reciprocal community space doesn't deserve a response. I also **_do_** want people to ask those things in user research or surveys (so, w/o expectations of reciprocity).
- Updated XSA answer (near line 201), as current one points to a broken link, and to better adhere to UX best practices for FAQs and content writing (Don't simply give users an answer to link to another page, w/o first giving them a TL;DR high-level response where they expect one. Referring users simply to a link, frustrates folks).
- Edited answer to Passwordless Sudo (near line 247) for same UX best-practices reason cited in XSAs answer.
- Edited section name "Users" to "Using Qubes" (near line 319) as "Users" is not an intuitive section name for end users to identify their needs with. Will that break things, Andrew?
- Re-worded multi-user system question to speak more to how less-technical people model that question in their own heads (eg: only nerds or IT people think of this concept as "multi-user system." Will also help SEO for people wanting to discover this question with Command-F
- Added (near line 372) new question to speak to governance idea Andrew proposed, of only speaking to version-specific compatibility between things, in the actual Docs


ninavizz:
Oops—meant to include this link in the first (I think): https://en.wikipedia.org/wiki/Hypervisor
Currently there is a placeholder in place of the actual link.

andrewdavidwong:
> Killed "Why does Qubes use Xen instead of KVM or some other hypervisor?" question, because it refernces the 0.3 architecture spec—which, even if that information is still correct, feels horribly out of date.

But why ignore correct information just because it's "old" (which just means it's remained true for a long time)? From that perspective, the study of history has no value.

In this case, other parts of that document are actually outdated, and we've decided to delist it from the doc index, so it makes sense not to reference it, but that doesn't mean the entire question has to go, just the reference. This does still seem to be a frequently-asked question, FWIW.

> Killed "Could you please make my preference the default?" as answer sounds scoldy... and, imho anybody asking that in a reciprocal community space doesn't deserve a response. I also do want people to ask those things in user research or surveys (so, w/o expectations of reciprocity).

I have to link to that one fairly often. Sometimes people deserve to be scolded, but I'd be fine with toning it down. "Doesn't deserve a response" simply doesn't work with some people. They'll assume they're right and we're just too lazy or cowardly to respond (because deep down we know we're wrong and they're right). This emboldens them to beat the drum louder and make my life more difficult. No thanks.

> Updated XSA answer (near line 201), as current one points to a broken link, and to better adhere to UX best practices for FAQs and content writing (Don't simply give users an answer to link to another page, w/o first giving them a TL;DR high-level response where they expect one. Referring users simply to a link, frustrates folks).

In theory, I agree, but in practice, offering just a link is the lesser evil compared to content duplication. This is a general point about the documentation as a whole. Ideally, we would have modular blocks of content that could be dynamically included in various places, but that would require a lot of work, and we're not really set up to do it right now.

> Edited section name "Users" to "Using Qubes" (near line 319) as "Users" is not an intuitive section name for end users to identify their needs with. Will that break things, Andrew?

Probably. Have a look at the CI log to see.

One non-technical thing it breaks is the parallelism between "users" and "developers," as well as introducing a gerund while the other headings at this heading level are not.

andrewdavidwong:
> Oops—meant to include this link in the first (I think): https://en.wikipedia.org/wiki/Hypervisor
> 
> Currently there is a placeholder in place of the actual link.

Go ahead and update the placeholder.

ninavizz:
> One non-technical thing it breaks is the parallelism between "users" and "developers," as well as introducing a gerund while the other headings at this heading level are not.

This is one thing I will push back on. Folks don't think through language on webpages like that; words either resonate with them, or they don't. That's a big tenant of usability; to design how users really think. Rationally and irrationally. I don't believe the word "users" is as easily recognizable to non-techies, looking for guidance... whereas "developers" will resonate with developers. 

I don't know what a a gerund is. Would you prefer if I change developers to "Development" or something else to better align with "Using Qubes"? 

> In theory, I agree, but in practice, offering just a link is the lesser evil compared to content duplication. 

Dig, will revert!

> I have to link to that one fairly often. Sometimes people deserve to be scolded, but I'd be fine with toning it down.

Good lord. Yeah, making your life easier is the goal with much of these! Ok, I'll re-write the answer and the question, both, so it more clearly speaks to what happens in those situations, without also discouraging honest feedback in the context of user feedback _when it is sought_. Which is not in the places you encounter it (whereas in surveys or user interviews, it's totally desired).

> But why ignore correct information just because it's "old" (which just means it's remained true for a long time)? From that perspective, the study of history has no value.
> 
> In this case, other parts of that document are actually outdated, and we've decided to delist it from the doc index, so it makes sense not to reference it, but that doesn't mean the entire question has to go, just the reference. This does still seem to be a frequently-asked question, FWIW.

Because that had been my takeaway as your guidance from my question yesterday in Wire? :D 

My only goal, is to keep out-of-date information out of the FAQ. Is your suggestion to then amend the question to simply `In short: we believe the Xen architecture allows for the creation of more secure systems (i.e. with a much smaller TCB, which translates to a smaller attack surface).` ??

andrewdavidwong:
I'll do a copyediting pass for language style and mechanics when merging, but here's my content feedback for now. Thanks for your willingness to help improve the FAQ! 🙂

andrewdavidwong:
> > One non-technical thing it breaks is the parallelism between "users" and "developers," as well as introducing a gerund while the other headings at this heading level are not.
> 
> This is one thing I will push back on. Folks don't think through language on webpages like that; words either resonate with them, or they don't. That's a big tenant of usability; to design how users really think. Rationally and irrationally. I don't believe the word "users" is as easily recognizable to non-techies, looking for guidance... whereas "developers" will resonate with developers.
> 
> I don't know what a a gerund is. Would you prefer if I change developers to "Development" or something else to better align with "Using Qubes"?
> 

I see what you're saying. I'd just prefer to maintain a consistent pattern, all else being equal, like "Using Qubes" and "Developing Qubes" (although the latter sounds a bit odd), and we still have that weird "General/Security" section, so maybe it's too hard to find a consistent pattern at this level.

This is more of a stylistic preference, not a deal-breaker.

> > I have to link to that one fairly often. Sometimes people deserve to be scolded, but I'd be fine with toning it down.
> 
> Good lord. Yeah, making your life easier is the goal with much of these! Ok, I'll re-write the answer and the question, both, so it more clearly speaks to what happens in those situations, without also discouraging honest feedback in the context of user feedback _when it is sought_. Which is not in the places you encounter it (whereas in surveys or user interviews, it's totally desired).

What if we just remove the last sentence or two?

> > But why ignore correct information just because it's "old" (which just means it's remained true for a long time)? From that perspective, the study of history has no value.
> > In this case, other parts of that document are actually outdated, and we've decided to delist it from the doc index, so it makes sense not to reference it, but that doesn't mean the entire question has to go, just the reference. This does still seem to be a frequently-asked question, FWIW.
> 
> Because that had been my takeaway as your guidance from my question yesterday in Wire? :D
> 

I thought you asked about a different question rather than the KVM vs. Xen one. Anyway, I think the principle still holds. Correct information = good. Referencing outdated docs = bad. We can preserve the former while avoiding the latter.

> My only goal, is to keep out-of-date information out of the FAQ. Is your suggestion to then amend the question to simply `In short: we believe the Xen architecture allows for the creation of more secure systems (i.e. with a much smaller TCB, which translates to a smaller attack surface).` ??

Something like that, yeah. Unless, of course, this turns out to be too brief of an answer to be useful, in which case we could remove the entry, but I suspect we'd end up writing a new answer and adding it back later (after receiving that question too many times and not having an easy answer to link, which is why we added it in the first place).

ninavizz:
Bah! I just posted an update, with a long new answer to the "Make my preference the default" question. 

My goal with the new answer, is to provide insight into _why_ it is lame for folks to make those requests—as it's been my own assumption from emails I see and survey responses, that people just expect everything gets made like Mac and Windoze systems do. Preposterous and obviously ridiculous as that might seem to us (tho you personally have earned yourself a Lambourghini with your Qubes contribs, and then some!). 

Super curious to hear what you think! I also really do not want to discourage folks from responding with honesty, where appropriate. Those channels did not exist before, and now they do. Not sure how much that may matter tho, just guessing. And I respect that you're the one who has to take the brunt of the selfish bozos. 

I see what you're saying. I'd just prefer to maintain a consistent pattern, all else being equal, like "Using Qubes" and "Developing Qubes" (although the latter sounds a bit odd), and we still have that weird "General/Security" section, so maybe it's too hard to find a consistent pattern at this level.

> (Users vs Using Qubes) This is more of a stylistic preference, not a deal-breaker.

All of these changes, for me, are incremental changes working towards an improved design for the page itself. With that updated design, I'm hoping to [have more categories](https://github.com/ninavizz/ninavizz/wiki/Qubes-FAQ-Migration-Content-Prep), and separate categories for "General" and "Security." So, I'm not at all married to making the "Users" to "Using Qubes" change, now. Whatever you're comfortable with, I'm happy to go with. 

Mostly just trying to make some easy and incremental changes, with this. I leave it to your calls, from here! Unless you'd prefer to put anything back in my court, in which I'd be happy to oblige.

andrewdavidwong (changes requested):

andrewdavidwong:
Oh, I think you missed my inline comments. I'm not sure if GitHub wasn't displaying them to you because I didn't submit the review earlier.

introduction/faq.md
```diff
@@ -610,13 +603,14 @@ Qubes is slower when reading from the disk because of the VM overhead, which is
...
+
+In 2019 we began our first structured research effort to learn more about [who's using Qubes](https://www.qubes-os.org/news/2020/11/26/qubes-survey-results/). Between that survey and follow-up interviews, we've learned that some of our users are security researchers in academia, while others are human rights workers still struggling to get their heads around using GPG; many of our users are developers, while many others are less-technically adept but still die-hard Linux and OSS enthusiasts. Many others are workers in government or enterprise environments. Some of our users can't imagine using Qubes with anything other than an i3 DE with D-Menu, while others still struggle with Qubes leaving KDE for XFCE as our default DE. 
+
+TL;DR, chances are that if thousands of users will truly benefit from something, that there's already a GitHub issue for it. That issue could be many months (or years) old, too. Or, there's a non-negotiable reason something is not the default already. We love our user community. We're just a very small, modest team, and ask for your humble understanding in not overwhelming us with what we'll never be able to meaningfully act upon.
```
andrewdavidwong, line 612: This is a bit too rambling and hand-wringing for my taste. It sounds too apologetic about things that merit no apology and beats around the bush instead of giving a straight answer.

introduction/faq.md
```diff
@@ -19,23 +19,24 @@ title: Frequently asked questions (FAQ)
...
-Depending on what kind of malware it is, it might do anything from showing you unwanted advertisements to logging your keystrokes to taking over your entire computer.
-This could jeopardize all the information stored on or accessed by this computer, such as health records, confidential communications, or thoughts written in a private journal.
-Malware can also interfere with the activities you perform with your computer.
-For example, if you use your computer to conduct financial transactions, the malware might allow its creator to make fraudulent transactions in your name.
+Most people use an operating system like Windows or macOS on their desktop and laptop computers. These OSes are popular because they tend to be easy to use and usually come pre-installed on the computers people buy. However, they present problems when it comes to security. Most common, anybody can open an innocent-looking file, email attachment, or website, not realizing that it's actually allowing malware—a catch-all phrase for any/all software created for malicious use, unbeknownst to an end user—to run on their computer.
```
unman, line 34: OSs, not OSes
"Most Common, anybody" - ? What about "For example, someone.."
"any/all" = "any"
Space either side of the hyphen.
There should be a hyphen after "use" instead of a comma, or put that in parentheses.
"unbeknownst to an end user" is wrong.

introduction/faq.md
```diff
@@ -19,23 +19,24 @@ title: Frequently asked questions (FAQ)
...
+
+### Is it safe for code to be so open like that? What about bad actors?
+The source code for Qubes is openly available so that anyone can contribute to and audit it. Only the CTO of Invisible Things Lab has the _[signing key](https://en.wikipedia.org/wiki/Code_signing)_ to create certified packages for the public to download and install. Signing code is a normal process that most organizations do, to garuntee the integrity of what users install on their machines. So, don't worry—it is a good thing that "just anyone" can audit and contribute to Qubes OS! Qubes is only possible, in fact, because of the large and generous community that contributes to building, auditing, and continually improving it.
```
unman, line 30: guarantee
space around hyphen

introduction/faq.md
```diff
@@ -19,23 +19,24 @@ title: Frequently asked questions (FAQ)
...
+
+Malware can do anything from showing you unwanted advertisements, to logging your keystrokes to taking over your entire computer. If you use your computer to conduct financial transactions, malware can allow its creator to make fraudulent transactions in your name. There are many kinds of malware, made by folks and organizations from all over the world; some for targeted harm, some more simply for passive mischief. Those are just a couple of common examples. 
+
+The specific risks malware presents to an operating system, is that it can jeopardize all the information stored on or accessed by a computer; information such as one's health records, confidential communications, or thoughts written in a private journal. In the same way a virus mutates, the makers and types of malware are also changing every minute of every day, to adapt to known software mitigations. So, considering the vulnerability of an entire operating system as one giant exposed hub to a much larger a digital ecosystem, is an important step in a reasonable approach to digital security.
```
unman, line 38: risk ...is that

> jeopardize all the information stored on or accessed by a computer; information such as one's health records, 

What about - "jeopardize all the information stored on or accessed by a computer: e.g. health records, banking information, confidential communications, or thoughts written in a private journal"

makers and types - what is the relevance of "makers"? Why not just say, "malware also changes..."
"So" isn't needed.
Something instead of "hub?" 
"a" is duplicated
what about "toward" in place of "in a"?

unman (changes requested):

introduction/faq.md
andrewdavidwong: I'll rewrite all of this in my copy editing pass after the substantive content matters have been worked out.

introduction/faq.md
andrewdavidwong: (Ditto.)

introduction/faq.md
ninavizz: @unman "For example" does not make clear why something is important, in an ordering of priorities. FAQs exist to deliver a TL;DR explainer to folks, and with infosec its important to clearly communicate that we're not willy-nilly making suggestions because they align with willy-nilly interests of ours, but because things really do happen that often. 

All of my recommendations are based on many years of user interviews in enterprise UX, that include dispelling IT myths end-users believe, because how IT likes to communicate often doesn't make sense to them. Some of my recommendations, are directly from interviewing journalists to learn about their security habits/beliefs. 

Hyphen was supposed to be an em-dash. 

Unsure how "unbeknownst to an end user" is incorrect? Malware doesn't ask for permission. It does things w/o consent. That is what I'm trying to communicate; that the user won't see what's happening.

introduction/faq.md
ninavizz: Perhaps we could discuss some more, perhaps offline? 

The way the current answer is written, assumes the end user has some (any) visibility into how software is made. From my experiences interviewing users for user research, most have none. Or, their assumptions are misaligned with how big a team is, how many users there really are, or what the diversity of the user base is. So, it's not being apologetic at all; it's providing information to give the user an opportunity to see why we may think their request is selfish. 

Especially as more and more non-technical and not-FOSS-community users come into the Qubes fold, I'd like to see more of the FAQ and intro content assume less about a user's knowledge of how our sausage gets made. If that makes sense.

introduction/faq.md
ninavizz: @andrewdavidwong "I'll rewrite all of this;" all of what I've written, or all of the mechanical language things Unman is pointing out? Unclear what, specifically, you're referring to. If my writing improvements just aren't desired at all, I'd prefer not to spend the time, here, doing them.

andrewdavidwong:
Spoke with @ninavizz about this PR. I'm going to try a revision soon and see how it goes.

ninavizz:
@unman Apologies for the exclusion, by doing a phone call with Andrew. Andrew is an expert wordsmith, and I am an expert OMGIWANTALLTEHTHINGS-smith. It really helped hearing his feedback on my writing, after I shared with him my goals in composing the blurbs I had. I'm also not an async-native worker, so the telephone voice conversation and hearing one of the most articulate writers I've encountered on GitHub speak in verbal conversation (he speaks in distinctly Western American accent, with "like" as every-other word) and with a ready laugh, was strangely soothing. :) 

We also discussed a strategy that could well work better for my interests in seeing writing to address both user concerns I observe in survey responses, and user-centric/ux-best-practice needs, here. So, I'll be writing issues, mostly, for written content, going forward.

You also have a much stronger grasp of writing mechanics, than I do—and I love your writing contributions. And, you have a direct connection to many of Qubes' users, that I do not. :)

andrewdavidwong:
Revised first two entries.

andrewdavidwong:
Fixed wording and typo.

andrewdavidwong:
Revised third entry.

ninavizz:
@andrewdavidwong Will review tonite, thx for the ping!

introduction/faq.md
```diff
@@ -248,7 +258,7 @@ At the same time, due to the smart use of Xen shared memory, our GUI implementat
 
 ### Why passwordless sudo?
 
-Please refer to [this page](/doc/vm-sudo/).
+We have found that there's really no point in user/root isolation, because all the user data (and VM management interface) is already accessible from dom0 user level; so, there is nothing more to get from a dom0 root account. Read more about: [Passwordless root access in Qubes](https://www.qubes-os.org/doc/vm-sudo/)
```
tlaurion, line 261: @ninavizz  `/doc/vm-sudo/`  instead of direct link `https://www.qubes-os.org/doc/vm-sudo/`

maiska:
@ninavizz Hello! Would you be so kind to update the PR to reflect the current state of the origin main branch while keeping the suggestions you consider important? :)
P.S. Thanks for all the work!

maiska:
@ninavizz  Thank you for the suggestions! :) I will close the PR but we will re-consider those after rst doc conversion.

PR #1231, opened 2022-02-20
Added install command sudo apt install gnupg2

kangarooo:


andrewdavidwong:
This assumes the user is using Debian or another APT-based system, which is not always the case, and we don't want to try to maintain a list of commands for every Linux package manager here.

kangarooo:
Then can fix to add install command of fedora also. So 2 commands needed? Here is one already, can add 2nd also.

andrewdavidwong:
> Then can fix to add install command of fedora also. So 2 commands needed? Here is one already, can add 2nd also. If command is mentioned, then need install command of correct package for both systems.

No, because there are also many other Linux distros and package managers, and we don't want to saddle ourselves with maintaining a list of commands for all of them.

kangarooo:
What others there are then debian and fedora?

andrewdavidwong:
> What others there are then debian and fedora?

Have a look at this site, which maintains an extensive list:

https://distrowatch.com

They also maintain pages about package management, e.g.:

https://distrowatch.com/dwres-mobile.php?resource=package-management

andrewdavidwong:
> So then 3 most popular would be needed to write?
> 
> debian apt
> 
> fedora rpm
> 
> arch pacman
> 
> Any other is very popular?
> 
> For rest people would understand what they use and that they need to install with their default package manager.

Don't those three distros already come with GPG preinstalled in their mainstream flavors? I'd be really surprised if they didn't.

andrewdavidwong:
> For rest people would understand what they use and that they need to install with their default package manager.

Why wouldn't this also apply to users of the three distros you listed?

The main distro that caters to newbies where this might not apply is Ubuntu. Btw, how is this not in the top three of popularity?

andrewdavidwong:
I'm aware that Ubuntu is based on Debian and that they both use apt, but that doesn't mean they're the same OS. There are a great many distros based on Debian.

PR #1235, opened 2022-03-02
add link to 2020 article for new 4.1 policy

taradiddles:
quick-and-dirty edit - following a [discussion ](https://forum.qubes-os.org/t/wip-windows-qwt-user-reports/9687/16?u=taradiddles) with @andrewdavidwong on the forum...
(feel free to change the text!).

### DIFF
--- a/user/advanced-topics/rpc-policy.md
+++ b/user/advanced-topics/rpc-policy.md
@@ -6,6 +6,8 @@ ref: 178
 title: RPC policies
 ---
 
+:warning: *Qubes OS 4.1 introduced a new qrexec policy system and policy format; please see [this article](https://www.qubes-os.org/news/2020/06/22/new-qrexec-policy-system/) for more details.*
+
 This document explains the basics of RPC policies in Qubes.
 For more information, see [Qrexec: command execution in VMs](/doc/qrexec3/).
 


andrewdavidwong:
Thanks. For future reference, please make sure to use [relative links](https://www.qubes-os.org/doc/documentation-style-guide/#relative-vs-absolute-links) for internal website links.

taradiddles:
> For future reference, please make sure to use [relative links](https://www.qubes-os.org/doc/documentation-style-guide/#relative-vs-absolute-links) for internal website links.

Ah - forgot about that as the link's first path component was `/news`, not `/doc`. I see you haven't changed the link to make it relative, do you want me to send a PR to fix it ?

andrewdavidwong:
Btw, is the `:warning:` syntax from somewhere else? AFAIK that wouldn't do anything here.

andrewdavidwong:
> It's a pretty standard markdown emoji (at least here on github and on discourse) -  - so I didn't even pay attention it might not work with the markdown to html conversion for qubes doc...

I don't believe that's part of the Markdown spec. At least, I've never encountered it after reading a fair amount of Markdown syntax documentation. I tried Ctrl+F searching for `emoji` in the [GitHub Flavored Markdown Spec](https://github.github.com/gfm/) and got zero hits. I'm guessing it's some kind of additional emoji system that GitHub and many other apps and platforms who happen to use Markdown support. I don't believe this has anything to do with "Markdown to HTML conversion," which is really just the Jekyll engine rendering our site from (largely) Markdown source files. I believe we also use a fair amount of GitHub Flavored Markdown, so it's probably not even a Markdown spec difference. That said, I've never actually tried the `:warning:` syntax on the Qubes site because it never occurred to me that it would do anything other than display literally. (We use FontAwesome for most of our icon needs.)

taradiddles:
> I don't believe that's part of the Markdown spec

I don't think it is either. It's a convenient feature though as markdown lacks a way to display warning boxes (known workarounds: inline html code or [abuse tables](https://stackoverflow.com/questions/58737436/how-to-create-a-good-looking-notification-or-warning-box-in-github-flavoured-mar#58758521)).

> I don't believe this has anything to do with "Markdown to HTML conversion,"

I meant I'd expect any markdown to html "rendering" engine to ignore non-standard markdown spec stuff. Jekyll doesn't seem to support emojis natively but there are plugins like [this popular one](https://github.com/jekyll/jemoji). Using this in qubes doc is debatable though and warrants its own topic :)

PR #1239, opened 2022-03-14
fix name

deeplow:
According to @marmarek's comment in https://github.com/QubesOS/qubes-issues/issues/6507#issuecomment-1067135102


andrewdavidwong:
Thanks!

PR #1240, opened 2022-03-16
Fix stage numbering; copyedit

andrewdavidwong:
See: https://forum.qubes-os.org/t/10156


andrewdavidwong:
@marmarek, is this correct?

1. STAGE 0
2. STAGE 1
3. STAGE 2
4. STAGE 3
5. STAGE 4
6. STAGE 5
7. `sudo qubes-dist-upgrade --resync-appmenus-features`

So there are actually seven stages in total, but the last "stage" isn't labeled `STAGE 6`?

PR #1254, opened 2022-05-26
Describe adding a menu entry for applications under wine

GWeck:


andrewdavidwong:
If I understand correctly, this is about creating a menu entry for a third-party program running under another third-party program running under a third-party guest OS. While all the third-party stuff is quite removed from Qubes, the menu entry in dom0 perhaps is not. Still, I have no idea who on our team would be right to review this sort of content. Any suggestions, @marmarek? Does this indicate that this sort of content instead belongs in the community documentation?

GWeck:
It's not so much about creating a new menu entry but just a workaround for a minor bug in the automatic creation of such menu entries.

Perhaps @fepitre or @DemiMarie could have a look at this?

user/troubleshooting/app-menu-shortcut-troubleshooting.md
```diff
@@ -115,3 +115,22 @@ Actual command lines for the menu shortcuts involve `qvm-run` command which star
...
+~~~
+[Desktop Entry]
+Version=1.0
+Type=Application
+Terminal=false
+X-Qubes-VmName=personal
+Icon=/usr/share/icons/hicolor/48x48/apps/Excel.png
+Name=Microsoft Excel
+Categories=X-Qubes-VM;
+Exec=env WINEPREFIX="/home/user/.wine" /usr/bin/wine C:\\\\windows\\\\command\\\\start.exe /Unix /home/user/.wine/dosdevices/c:/users/user/Start\\ Menu/Programs/Microsoft\ Excel.lnk
```
DemiMarie, line 132:
```suggestion
Exec=env WINEPREFIX="/home/user/.wine" /usr/bin/wine C:\\\\windows\\\\command\\\\start.exe /Unix /home/user/.wine/dosdevices/c:/users/user/Start\\ Menu/Programs/Microsoft\\ Excel.lnk
```

(more generally, is this the correct number of backslashes everywhere?)

marmarek:
> The problem that I am addressing is that this needs a corresponding menu definition in the Template/AppVM that may be located in the wrong place, causing the dom0 definition to point into the void.

Subdirectories should be handled correctly now (it was an issue before, but is fixed in R4.1 already). What I see is not handled correctly is file names with spaces. In your example, I guess that was `Microsoft Excel.desktop`. Try renaming it to `Microsoft_Excel.desktop` or similar.

I was curious what "Desktop Entry Specification" says about file names there, but it wasn't very helpful: the original spec haven't specified it, and version 1.2 (in 2017) added naming based on D-Bus names, which isn't really applicable here. Anyway, it feels like everybody assumes there are no spaces in `.desktop` file names (and so do we...).

@DemiMarie any idea for some sensible escaping? Perhaps use `_` for space? If there are both `Microsoft_Excel.desktop` and `Microsoft Excel.desktop`, it would be undefined which one would be used, but _if_ such conflict is rare, I think it's okay to leave it as "undefined behavior" (the "Desktop Entry Specification" says the same about `/` vs `-`).

DemiMarie:
> @DemiMarie any idea for some sensible escaping? Perhaps use `_` for space? If there are both `Microsoft_Excel.desktop` and `Microsoft Excel.desktop`, it would be undefined which one would be used, but _if_ such conflict is rare, I think it's okay to leave it as "undefined behavior" (the "Desktop Entry Specification" says the same about `/` vs `-`).

I would prefer to avoid undefined behavior.  For one, it leads to corner cases that do not get tested, and untested corner cases tend to be full of bugs and even security holes.

One approach would be to replace space with `_`, and append `_` and the SHA256 hash of the name.  If the name exceeds 194 bytes, the remaining bytes are discarded after hashing, but before escaping.  This ensures that desktop names are unique, and also avoids problems with long paths.

GWeck:
Escaping blanks, especially through several layers of scripts is a nightmare! Here are two examples that work:
~~~
Exec=env WINEPREFIX="/home/user/.wine" wine C:\\\\ProgramData\\\\Microsoft\\\\Windows\\\\Start\\ Menu\\\\Programs\\\\Avanti\\ Virtual\\ Alpha\\\\Virtual\\ Alpha\\ Configuration\\ Utility.lnk
~~~
~~~
Exec=env WINEPREFIX="/home/user/.wine" wine C:\\\\windows\\\\command\\\\start.exe /Unix /home/user/.wine/dosdevices/c:/ProgramData/Microsoft/Windows/Start\\ Menu/Programs/IrfanView/IrfanView\\ 64\\ 4.58.lnk
~~~
Somehow I seem to have hit the right number of backslashes. I have no idea why the format of the two start commands is so different, one using the Windows utility `start.exe` and the other one directly positioning to the link - they were created just like that on the installation of the software.

Regarding submenus: If, for instance, the desktop file is stored in the AppVM  as
`~/.local/share/applications/wine/Programs/SomeProgramDir/SomeProgram.desktop`
it will be found if the program starter in dom0 uses
`qubes.StartApp+wine+Programs+SomeProgramDir+SomeProgram`
I couldn't check if this is always built correctly on program installation under wine, but if it is set like that, perhaps manually, subdirectories and submenus work.

unman:
@GWeck I ma sorry this has rotted for so long.
Would it be possible for you to review this under 4.2 and confirm that the process works under Wine?

GWeck:
I checked now with Qubes R4.2.1 and wine 9.0. The menus are built correctly, and the applications can be started. There is just still the one exception discussed above: If the desktop file has a space in its name, like `Microsoft Excel.desktop`, the application cannot be started. Removing the space and then refreshing the menus via `qvm-sync-appmenus` or using the Qube Manager rebuilds the menus, and then these applications can be started, too. No further action is needed.

PR #1258, opened 2022-07-09
Update proxy documentation clarification

nyxnor:
It had deprecated information and also was not very clear of what is protects against, fixes https://github.com/QubesOS/qubes-issues/issues/7610

Also used marmarek's quote from here https://forum.qubes-os.org/t/why-does-the-updateproxy-not-have-a-whitelist/11087/4

### DIFF
--- a/user/how-to-guides/how-to-install-software.md
+++ b/user/how-to-guides/how-to-install-software.md
@@ -144,11 +144,11 @@ Please see [How to Update](/doc/how-to-update/).
 
 In order to protect you from performing risky activities in templates, they do
 not have normal network access by default. Instead, templates use an [updates
-proxy](#updates-proxy) that allows you to install and update software without
-giving the template direct network access. **The updates proxy is already set
-up to work automatically out-of-the-box and requires no special action from
-you.** Most users should simply follow the normal instructions for [installing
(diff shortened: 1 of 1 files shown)


user/how-to-guides/how-to-install-software.md
```diff
@@ -304,19 +304,23 @@ This is like the simple revert, except:
...
+to applications trusted by the user, configured to use the proxy.
+The http proxy (tinyproxy) does not filter traffic because it is hard to list
+all the repository mirrors and keep that list up to date). it also does not
+cache anything.
 
 The proxy is running in selected VMs (by default all the NetVMs (1)) and
-intercepts traffic directed to 10.137.255.254:8082. Thanks to such
+intercepts traffic directed to 127.0.0.1:8082. Thanks to such
 configuration all the VMs can use the same proxy address, and if there is a
 proxy on network path, it will handle the traffic (of course when firewall
```
marmarek, line 321: The "if there is a proxy on network path" part is not applicable anymore (the updates proxy is using qrexec now, as documented below, not intercepted TCP connection).

marmarek (changes requested):

PR #1262, opened 2022-08-08
Update minimal-templates.md

TheWanderer1983:
Yubikey works out of the box for 2fa responses in debian-11.template but not in the minimal. It isn't straight forward to know which package to include for that functionality to work. Others on the forum have had this issue also. I've found that package and included it here so others can easily add that functionality.

### DIFF
--- a/user/templates/minimal-templates.md
+++ b/user/templates/minimal-templates.md
@@ -254,7 +254,7 @@ list of packages to be installed):
   to configure it.
 - `default-mgmt-dvm`: requires `qubes-core-agent-passwordless-root` and
   `qubes-mgmt-salt-vm-connector`.
-- [Yubikey](/doc/yubikey/): You may need to install `xserver-xorg-input-all` for 2FA responses to work in web browsers like Firefox.
+- [Yubikey](/doc/yubikey/): You may need to install `xserver-xorg-input-libinput` for 2FA responses to work in web browsers like Firefox.
 
 In Qubes 4.0, additional packages from the `qubes-core-agent` suite may be
 needed to make the customized minimal template work properly. These packages
--- a/user/templates/minimal-templates.md
+++ b/user/templates/minimal-templates.md
@@ -254,6 +254,7 @@ list of packages to be installed):
   to configure it.
 - `default-mgmt-dvm`: requires `qubes-core-agent-passwordless-root` and
   `qubes-mgmt-salt-vm-connector`.
+- [Yubikey](/doc/yubkiey/), you may need to install `xserver-xorg-input-all` for 2fa responses to work in web-browsers like Firefox.
 
 In Qubes 4.0, additional packages from the `qubes-core-agent` suite may be
 needed to make the customized minimal template work properly. These packages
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
> Description: X.Org X server -- input driver metapackage
 This package depends on the full suite of input drivers for the X.Org X server
 (Xorg).  It does not provide any drivers itself, and may be removed if you wish
 to only have certain drivers installed.

Not sure if this means that this package by itself couldn't be responsible for getting your YubiKey to work, so I guess I'll ask @marmarek to review this PR as a quick sanity check.

TheWanderer1983:
Hey @andrewdavidwong 
I believe xserver-xorg-input-libinput is required but xserver-xorg-input-all installs that package as well as wacom.  Installing that package worked for me and is the package installed on the debian-11 template that comes with R1. 
The yubikey won't 2fa with any web-browser 2fa (lastpass, twitter etc)  when you press the button unless that package is installed on a debian minimal template.  
[Yubikey](https://wiki.debian.org/Smartcards/YubiKey4) want libu2f-host0 installed to work but that isn't installed on the default debian-11 template and it's working. It appears to be using libinput instead.

TheWanderer1983:
This is actually  a bug not a fix. I have sys-usb with all my usb controllers running through it. I just noticed yesterday even If I don't have the yubico attached to a specific VM it still feeds in my yubikey response with a finger press. There should be no input if i haven't attached the yubico to a AppVM.  As you can see below it isn't attached to anything but when I press the yubikey button it feeds in my correct response. I just tested it with three separate VMs and its inputting from the yubico. It appears xserver-xorg-input-all is breaking sys-usb seperation.

qvm-pci list 

`dom0:00_14.0 USB Controller: Intel Corporation Comet Lake PCH-LP USB 3.1 xHCI Host controller Sys-usb (no-strict-reset= True)`

qvm-usb
```
BACKEND:DEVID DESCRIPTION USED BY
sys-usb:2-3 Yubico_Yukikey_4_OPT+U@F+CCID  
```

TheWanderer1983:
Sorry to keep doing this. It seems I was wrong and it isn't a bug. Yubikey acts as a USB keyboard and that is why adding xserver-xorg-input-all fixes the problem with minimal. I have USB keyboard and mouse and therefore it is accepted in dom0. It's not breaking anything. The original point should still stand though about needing that package to work.

marmarek:
> I believe xserver-xorg-input-libinput is required

This looks like a better idea than recommending -all (in the spirit of keeping minimal template minimal).

user/templates/minimal-templates.md
```diff
@@ -254,6 +254,7 @@ list of packages to be installed):
   to configure it.
 - `default-mgmt-dvm`: requires `qubes-core-agent-passwordless-root` and
   `qubes-mgmt-salt-vm-connector`.
+- [Yubikey](/doc/yubikey/): You may need to install `xserver-xorg-input-all` for 2FA responses to work in web browsers like Firefox.
```
TheWanderer1983, line 257: - [Yubikey](/doc/yubikey/): You may need to install `xserver-xorg-input-libinput` for 2FA responses to work in web browsers like Firefox.

PR #1265, opened 2022-09-01
Fix style and add instructions for using Qube Manager to find USB controller/address

qubesbugreport:
For issue https://github.com/QubesOS/qubes-issues/issues/7711

### DIFF
--- a/user/how-to-guides/how-to-use-usb-devices.md
+++ b/user/how-to-guides/how-to-use-usb-devices.md
@@ -28,9 +28,9 @@ Examples of valid cases for USB-passthrough:
 (If you are thinking to use a two-factor-authentication device, [there is an app for that](/doc/u2f-proxy/).
 But it has some [issues](https://github.com/QubesOS/qubes-issues/issues/4661).)
 
-## Attaching And Detaching a USB Device
+## Attaching and detaching a USB device
 
-### With Qubes Device Manager
+### With Qubes device manager
 
 Click the device-manager-icon: ![device manager icon](/attachment/doc/media-removable.png)
 A list of available devices appears.
@@ -48,7 +48,7 @@ Hover on the attached device to display a list of running VMs.
 The one to which your device is connected will have an eject button ![eject icon](/attachment/doc/media-eject.png) next to it.
 Click that and your device will be detached.
 
-### With The Command Line Tool
+### With the command line tool
 
 In dom0, you can use `qvm-usb` from the commandline to attach and detach devices.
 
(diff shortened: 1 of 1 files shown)


user/how-to-guides/how-to-use-usb-devices.md
```diff
@@ -28,9 +28,9 @@ Examples of valid cases for USB-passthrough:
 (If you are thinking to use a two-factor-authentication device, [there is an app for that](/doc/u2f-proxy/).
 But it has some [issues](https://github.com/QubesOS/qubes-issues/issues/4661).)
 
-## Attaching And Detaching a USB Device
+## Attaching and detaching a USB device
 
-### With Qubes Device Manager
+### With Qubes device manager
```
andrewdavidwong, line 33: I'm not certain, but I think "Qubes Device Manager" is the name of the thing, so should remain in title case.

user/how-to-guides/how-to-use-usb-devices.md
```diff
@@ -155,10 +159,26 @@ This should output something like:
 ../../../devices/pci-0/pci0000:00/0000:00:1a.0/usb3
 ```
 
+If the output format does not match this example, or you are unsure if it contains the correct BDF address, you can try finding the address using method 2 using the Qube Manager instead.
+
```
andrewdavidwong, line 163: I might be missing something, but if Method 2 always works reliably, while Method 1 sometimes gives the wrong output, then why even have two methods? Why not just scrap Method 1 and only present Method 2? (Of course, in that case, we wouldn't call it "Method 2," we should just refer to it as *the* recommended method.)

andrewdavidwong (changes requested):

user/how-to-guides/how-to-use-usb-devices.md
qubesbugreport: Yes, you are right. That was a mistake.

user/how-to-guides/how-to-use-usb-devices.md
qubesbugreport: Well, I didn't want to take out the original method because I don't know why it was recommended in the first place instead of just looking in the Qubes Manager GUI. Maybe there are cases in which its not possible to use the GUI or maybe at the time this was written it was not yet possible to see the device addresses in the Qubes Manager?

andrewdavidwong: I'm not sure either. Perhaps @marmarek can steer us in the right direction.

marmarek: The described method is about finding to which USB controller given device is connected, not about just identifying some USB controller. The Method 2 described below does not help with this at all ("you can probably guess").

As for not working reliably, it does work, it just needs a little clarification regarding situation in https://github.com/QubesOS/qubes-issues/issues/7711: if there are multiple addresses in the path, the _last_ one is the BDF of the controller. Taking example from that issue, given `../../../devices/pci0000:00/0000:00:1c.0/0000:01:00.0/usb1`, the answer is `01:00.0`, not `00:1c.0` (this one is a PCI bridge to which the controller is connected).

PR #1266, opened 2022-09-04
doc: Replace highstate occurences with apply

sheerun:
As per salt docs state.apply replaced less readable state.apply ( yes, this happened to me :smile: ) 

> This action is referred to as a "highstate", and can be run using the [state.highstate](https://docs.saltproject.io/en/latest/ref/modules/all/salt.modules.state.html#salt.modules.state.highstate) function. However, to make the usage easier to understand ("highstate" is not necessarily an intuitive name), a [state.apply](https://docs.saltproject.io/en/latest/ref/modules/all/salt.modules.state.html#salt.modules.state.apply_) function was added in version 2015.5.0, which when invoked without any SLS names will trigger a highstate. [state.highstate](https://docs.saltproject.io/en/latest/ref/modules/all/salt.modules.state.html#salt.modules.state.highstate) still exists and can be used, but the documentation (as can be seen above) has been updated to reference [state.apply](https://docs.saltproject.io/en/latest/ref/modules/all/salt.modules.state.html#salt.modules.state.apply_), so keep the following in mind as you read the documentation:

 >   [state.apply](https://docs.saltproject.io/en/latest/ref/modules/all/salt.modules.state.html#salt.modules.state.apply_) invoked without any SLS names will run [state.highstate](https://docs.saltproject.io/en/latest/ref/modules/all/salt.modules.state.html#salt.modules.state.highstate)

>    [state.apply](https://docs.saltproject.io/en/latest/ref/modules/all/salt.modules.state.html#salt.modules.state.apply_) invoked with SLS names will run [state.sls](https://docs.saltproject.io/en/latest/ref/modules/all/salt.modules.state.html#salt.modules.state.sls)

### DIFF
--- a/user/advanced-topics/salt.md
+++ b/user/advanced-topics/salt.md
@@ -61,7 +61,7 @@ enforcing the state in a particular area.
 It exposes some *imperative* functions for the administrator.
 For example, there is a `system` module that has a `system.halt` function that,
 when issued, will immediately halt a domain.
-There is another function called `state.highstate` which will synchronize the
+There is another function called `state.apply` which will synchronize the
 state of the system with the administrator's configuration/desires.
 
 ### Configuration
@@ -184,7 +184,7 @@ $ qubesctl top.disable my-new-vm
 To apply the states to dom0 and all VMs:
 
 ```
-$ qubesctl --all state.highstate
+$ qubesctl --all state.apply
 ```
 
 (More information on the `qubesctl` command further down.)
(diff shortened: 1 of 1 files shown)


marmarek:
There is at least [one place](https://github.com/QubesOS/qubes-mgmt-salt/blob/master/qubessalt/__init__.py#L296) where we check for `state.highstate` specifically, so before changing the documentation, this needs to be converted too.

rapenne-s:
if the change was done in https://github.com/QubesOS/qubes-mgmt-salt/blob/master/qubessalt/__init__.py#L296 to unstuck this PR, would it be required to write release notes somewhere? Users will need to know that `state.highstate` has been replaced by `state.apply` (it's only true for a single use case though, but enough to deserve an entry)

sheerun:
I think your e-mail client doesn't recognize utf-8 characters. I've sent following emoji: https://emojipedia.org/party-popper

PR #1271, opened 2022-09-29
Modifications for RTD conversion

marmarek:
Some source modifications to make conversion to ReST format easier. In principle, they should be valid also without the conversion.

Marking as draft, as I'll add more do this PR.


marmarek:
PipelineRetry

marmarek:
PipelineRefresh

marmarek:
testDeploy

marmarek:
This is not relevant anymore (the converter tool was improved to deal with those cases), and more importantly outdated.

PR #1272, opened 2022-10-02
replace all files with redirects to RTD

marmarek:
This is preview of redirects to readthedocs. The final version will be put as doc.qubes-os.org (or similar), so this PR will need to be regenerated.


marmarek:
testDeploy

marmarek:
testDeploy

rapenne-s:
Can we close this PR?

PR #1273, opened 2022-10-02
Update stanalones-and-hvms.md 

nescravana:
To fix image size mismatch error when creating VM : "app: Error importing root volume (but VM created): Data import failed: not enough data (copied XXXX bytes, expected 10737418240)


marmarek (changes requested):
This is actually a bug that should be reported in qubes-issues. I did it now here: https://github.com/QubesOS/qubes-issues/issues/7821
Having workaround documented is fine, but please include the above issue reference.

unman:
Underlying issue was fixed by https://github.com/QubesOS/qubes-core-admin-client/commit/8078bded81ec5593a787ed9a657002ad109a2e6d

PR #1275, opened 2022-10-31
Minor update to qubes-iso-building.md

Gottale:
Change reference from 4.0 > 4.1, adjust template examples to match current releases, updated expired GPG key reference.


andrewdavidwong (approved):
Looks good to me, but let's also let @marmarek give it a quick look just in case.

marmarek (approved):

PR #1277, opened 2022-11-03
builder: add note about default selinux policy blocking the builder on fedora host

neowutran:
On modern fedora system like Fedora 37, SeLinux is enforced by default and is blocking the qubes-builder with error like this "can't create transaction lock on /.../rpm/.rpm.lock (Permission denied)". 

Adding a note so other people will spend less time understanding why the builder doesn't work out of the box. 
I personally just switched SeLinux to Permissive

### DIFF
--- a/developer/building/qubes-iso-building.md
+++ b/developer/building/qubes-iso-building.md
@@ -19,6 +19,11 @@ Fedora 32 has been successfully used to build Qubes R4.1 with the below steps.
 Other rpm-based operating systems may also work.
 Travis-CI uses Ubuntu 18.04 to perform test builds, except it can not test the `./setup` script.
 
+**Notes:** On modern Fedora system (like Fedora 37) SeLinux is enforced by
+default and is blocking the build system. You would get error like 
+"can't create transaction lock on /.../rpm/.rpm.lock (Permission denied)". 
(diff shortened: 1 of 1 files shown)


developer/building/qubes-iso-building.md
```diff
@@ -19,6 +19,11 @@ Fedora 32 has been successfully used to build Qubes R4.1 with the below steps.
 Other rpm-based operating systems may also work.
 Travis-CI uses Ubuntu 18.04 to perform test builds, except it can not test the `./setup` script.
 
+**Notes:** On modern Fedora system (like Fedora 37) SeLinux is enforced by
+default and is blocking the build system. You would get error like 
+"can't create transaction lock on /.../rpm/.rpm.lock (Permission denied)".
```
andrewdavidwong, line 24: Should a recommended workaround be included?

marmarek: The workaround is to switch selinux to permissive mode: `sudo setenforce 0`. And yes, I think it's a good idea to include it here.

Otherwise looks good.

neowutran: Ok :) added the workaround

PR #1282, opened 2022-12-25
Update installation-guide.md

jfldsf:
maybe just delete whole alert idk


andrewdavidwong:
Oh, since 4.0 is EOL, this alert text should be updated so that it doesn't mention 4.0. Do you want to do that, or shall I?

andrewdavidwong:
> maybe just delete whole alert idk

Okay, let's just go with this, since 4.0 is EOL.

PR #1289, opened 2023-01-25
Contributing: fix links

MarnixCroes:
current links give empty results


andrewdavidwong:
Thanks for catching these!

PR #1292, opened 2023-02-02
Add note to bind-dirs when dir does not exist yet

deeplow:
The doc on bind-dirs only provided an example where `/var/lib/tor` was made persistent in sys-whonix. However, if one tries to make persistent any directory that wasn't already present in the template qube, it would not persist as expected. The issue QubesOS/qubes-issues#5862 demonstrates this exact confusion.

This commit adds a note about this situation.

Fixes QubesOS/qubes-issues#5862

### DIFF
--- a/user/advanced-topics/bind-dirs.md
+++ b/user/advanced-topics/bind-dirs.md
@@ -44,6 +44,14 @@ Inside the app qube.
 
 4. Save.
 
+> **Note**: If the directory you are trying to persist doesn't already exist in the app qube, you'll need to create the directory with its full path, under `/rw/bind-dirs`. In this case you would do:
+>
+>    ```
+>    sudo mkdir -p /rw/bind-dirs/var/lib/tor
+>    ```
+>
+> This case happens only when the template on which this app qube is based does have this directory.
+
 5. Reboot the app qube.
 
 6. Done.


andrewdavidwong:
Thank you for the PR! Would you mind removing the block quotation syntax (the `>` characters)? I've noticed that it's common to want to use block quotation syntax with these sorts of notes, so I decided to add a [section](https://www.qubes-os.org/doc/documentation-style-guide/#do-not-use-markdown-syntax-for-styling) about this to the documentation style guide explaining why it should be avoided. (I just had to ask someone else to change this in a different PR the other day, so don't worry, it's not just you!)

user/advanced-topics/bind-dirs.md
```diff
@@ -44,6 +44,14 @@ Inside the app qube.
...
 4. Save.
 
+> **Note**: If the directory you are trying to persist doesn't already exist in the app qube, you'll need to create the directory with its full path, under `/rw/bind-dirs`. In this case you would do:
+>
+>    ```
+>    sudo mkdir -p /rw/bind-dirs/var/lib/tor
+>    ```
+>
+> This case happens only when the template on which this app qube is based does have this directory.
+
```
andrewdavidwong, line 54: Should this just be its own step, i.e., number 5 in the list? (If the concern is that some users may not need to take any action, I think that's handled by the fact that this is phrased as a conditional, i.e., "*If* the directory you are trying to persist doesn't already exist....")

deeplow: Sure!

deeplow:
> Would you mind removing the block quotation syntax (the > characters)?

Sure! I added it as step 5 as you suggested. I was trying to avoid alerts, since it's HTML in markdown and it will make migration to another platform harder.

user/advanced-topics/bind-dirs.md
```diff
@@ -44,9 +44,15 @@ Inside the app qube.
 
 4. Save.
 
-5. Reboot the app qube.
+5. If the directory you are trying to persist doesn't already exist in the app qube, you'll need to create the directory with its full path, under `/rw/bind-dirs`. In this case you would do:
+   ```
+   sudo mkdir -p /rw/bind-dirs/var/lib/tor
+   ```
+   This case happens only when the template on which this app qube is based does have this directory.
```
andrewdavidwong, line 51: Just to confirm, is it supposed to be "does have this directory" or "does not have this directory"?

deeplow: Good catch. I meant to say _does not_

deeplow: Fixed now via [eee99c7](https://github.com/QubesOS/qubes-doc/pull/1292/commits/eee99c70bb0dbce8a1a3de0b9c2b75ff72dbb8f5).

user/advanced-topics/bind-dirs.md
```diff
@@ -44,9 +44,17 @@ Inside the app qube.
...
 
-5. Reboot the app qube.
+5. If the directory you are trying to persist doesn't already exist in the app qube, you'll need to create the directory with its full path, under `/rw/bind-dirs`. In this case you would do:
 
-6. Done.
+   ```
+   sudo mkdir -p /rw/bind-dirs/var/lib/tor
+   ```
+   
+   This is required only when this directory does not already exist in the template on which the app qube is based.
```
andrewdavidwong, line 53: Okay, I just realized that something is still unclear to me. The first clause is:

> If the directory you are trying to persist doesn't already exist in the app qube

But the last sentence (now) is:

> This is required only when this directory does not already exist in the template on which the app qube is based.

So, which is the actual condition?

Is this step required when (a) the directory doesn't exist in the app qube, (b) the directory doesn't exist in the parent template, or (c) both?

deeplow: Both, where (a) is a consequence of (b). Maybe we can suppress the last part. It's just some contextual information as to why that happens. But as you pointed out, it may not be super clear.

If you agree with this, feel free to remove the last part.

andrewdavidwong: > Both, where (a) is a consequence of (b).

Okay, so as long as the directory exists in *either* the app qube *or* its template, this step is not required? In other words, this step is required only if the directory does not exist in *either* location?

Are you sure that's correct, or do you mean to say that the direct needs to exist in *both* locations?

deeplow: Sorry for not making myself clear. Generally, if the directory exists in the template, then it will also exist in the app qube because it's is inherited.

This step is not required when the directory already exists in the template (and thus, will already exist in the app qube).

andrewdavidwong: Okay, I've attempted to clarify the instructions based on our discussion. How does it look now?

andrewdavidwong: Ah, wait, I misunderstood again, but I think I've got it now. Added another commit.

deeplow: Yes, that's it. Thanks!

user/advanced-topics/bind-dirs.md
```diff
@@ -24,34 +24,42 @@ In this way sys-whonix can benefit from the Tor anonymity feature 'persistent To
...
    binds+=( '/var/lib/tor' )
    ```
 
 4. Save.
 
-5. Reboot the app qube.
+5. If the directory you wish to make persistent doesn't exist in the template on which the app qube is based, you'll need to create the directory (with its full path) under `/rw/bind-dirs` in the app qube. For example, if `/var/lib/tor` didn't exist in the template, then you would execute the following command in your app qube:
+
+   ```
+   sudo mkdir -p /rw/bind-dirs/var/lib/tor
```
rapenne-s, line 52: Maybe it should be mentioned that the permissions of these directories will be reused by bind-dirs, so you may have to take care of owner/group/permissions.

deeplow: That makes sense. But I'm not sure what expectations bind-dirs has regarding owner/group/permissions.

rapenne-s: The script will simply mount `/rw/bind-dirs/foobar` on top of `/foobar`, inheriting the permissions of `/rw/bind-dirs/foobar`.

So, if `/foobar` is owned by root:user and you mount `/rw/bind-dirs/foobar` owned by user:user, the latter is what you will see on `/foobar`

rapenne-s: good :) thanks

just to be pedantic, this can be done in one shot using `install -d -D -o owner -g group -m 755 /rw/bind-dirs/var/lib/tor`

rapenne-s (approved):
looks good to me

user/advanced-topics/bind-dirs.md
deeplow: Good to know! Although, I think I'll keep it as-is to make the commands simpler.

rapenne-s: I also think it should be kept as understandable as possible, user with basics skills should know about chown / chmod

unman:
@deeplow  I've reread the current state, and it doesnt cover the case where you want to use bind-dirs in a Template.
I'm not sure on how best to approach this - rewrite (again) to cover that case, or clarify your instructions to make it clear that they only cover the Template based qube position.

deeplow:
You're right @unman. I'll rework this.

deeplow:
> @deeplow I've reread the current state, and it doesnt cover the case where you want to use bind-dirs in a Template.
I'm not sure on how best to approach this - rewrite (again) to cover that case, or clarify your instructions to make it clear that they only cover the Template based qube position.

Thanks for taking a look @unman. However I was now trying to address this and I'm not sure I follow what you mean, since:
1. the original guide was already aimed at app qubes / Template based Qubes (see [here](https://github.com/QubesOS/qubes-doc/pull/1292/files#diff-776c235d9e747f9a841d070ca31d0e50b9867f6af032de406022a4ecf86725a0L29))
2. TemplateVMs don't really need bind-dirs AFAIK because everything (that usually matters -- aka: root) is already persistent

DemiMarie:
Time go merge this?

PR #1295, opened 2023-02-13
remove obsolete Whonix specific steps from vm-sudo

adrelanos:

PR #1296, opened 2023-02-15
Passwordless root

andrewdavidwong:
See the discussion on #1295 for background. 

It sounds like the original intention for this page was just to document passwordless root itself and the configuration changes associated with it. When the unofficial workaround to replace passwordless root with a dom0 prompt was devised, this seemed like the natural place to document it. In retrospect, though, that turned out to be scope creep (for this doc page). It sounds like the right thing to do is to move this page to the developer docs and move the section on replacing passwordless root with a dom0 prompt to the community docs.


unman:
I don't agree that this should be moved to developer docs.
It's a page linked from the FAQ ,and it's *still* an issue for many
people coming to Qubes from other Linux or BSD. It's an issue for many
users.
I think it's best left under Security.

andrewdavidwong:
> I don't agree that this should be moved to developer docs. It's a page linked from the FAQ ,and it's *still* an issue for many people coming to Qubes from other Linux or BSD. It's an issue for many users. I think it's best left under Security.

Do you think that the "configuration changes" section is appropriate for user documentation? If so, why? If not, then perhaps a further split is in order:

- Page documenting passwordless root itself in user docs
- Page documenting technical configuration changes associated with passwordless root in dev docs
- Page documenting unofficial workaround (dom0 prompt) in community docs

unman:
On reflection I have decided to move the specifics of passwordless-root to a page under developer documentation.
This leaves the user page focussed on the *why*, and alternatives.

PR #1297, opened 2023-02-26
Update template switching instructions

andrewdavidwong:
For context, see: https://forum.qubes-os.org/t/17031

One flaw with these instructions is that the disposable template will still have the old release number in its name, even though it is based on the new template. @marmarek, do you see an easy way to fix this without making the instructions even more complicated and user-error-prone for USB input device users?

### DIFF
--- a/user/templates/templates.md
+++ b/user/templates/templates.md
@@ -223,22 +223,20 @@ When you install a new template or
 recommended that you switch everything that was using the old template to the
 new template:
 
-1. Make the new or upgraded template the default template. In the App Menu, go
+1. **Make the new template the default template.** In the App Menu, go
    to Qubes Tools, then click on Qubes Global Settings. In the Qube Defaults
-   section, next to Template, select the new or upgraded template from the
+   section, next to Template, select the new template from the
    drop-down list. Press OK.
 
-2. Base your [disposable templates](/doc/glossary/#disposable-template) on the
-   new template. In the App Menu, go to Qubes Tools, then click on Qube
-   Manager. In the Qube Manager, find your disposable templates. (By default,
-   they end in `-dvm`.) Right click, then click on Settings. Next to Template,
-   select your new or upgraded template from the drop-down list. Press OK.
+2. **Base your [disposable templates](/doc/glossary/#disposable-template) on
+   the new template.**
 
-   - If you receive an error message because there are running
-     [disposables](/doc/glossary/#disposable) based on this disposable
-     template, and your only keyboard and mouse are *not* connected through a
-     [USB qube](/doc/usb-qubes/), or that USB qube is *not* a disposable, then
-     shut down all disposables based on the disposable template and try again.
+   - If your only keyboard and mouse are *not* connected through a [USB
+     qube](/doc/usb-qubes/), or that USB qube is *not* a disposable, then shut
+     down all disposables. In the App Menu, go to Qubes Tools, then click on
+     Qube Manager. In the Qube Manager, find your disposable template(s). (By
+     default, they end in `-dvm`.) Right click, hover over Template, then click
+     on the new template. Repeat for each disposable template.
 
    - If your only keyboard or mouse *are* connected through a USB qube, and
      that USB qube *is* a disposable, then you will have to enter a special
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
> One flaw with these instructions is that the disposable template will still have the old release number in its name, even though it is based on the new template.

One way to avoid this problem is by keeping the existing instructions, which create a new disposable template:

```
 [user@dom0 ~]$ qvm-create -l red -t <NEW_TEMPLATE> <NEW_DISPOSABLE_TEMPLATE>
 [user@dom0 ~]$ qvm-prefs <NEW_DISPOSABLE_TEMPLATE> template_for_dispvms True
 [user@dom0 ~]$ qvm-features <NEW_DISPOSABLE_TEMPLATE> appmenus-dispvm 1
 [user@dom0 ~]$ qubes-prefs default-dispvm <NEW_DISPOSABLE_TEMPLATE>
```

However, if all the user wants to do is switch existing stuff to a new template, then creating a whole new disposable template is unnecessary and out-of-scope. Also, this big block of commands is probably scary for non-CLI users, and making it seem like it's a required step in routine template upgrades is probably misleading and needlessly intimidating.

marmarek:
I don't have a better idea for R4.1. But I have some ideas what we can improve in R4.2. Specifically:
 - add deferred template change (allow changing template while VM is running, but actually do the change when VM shutdown/reboot) - I thought we have an issue for that already, but can't find right now
 - create default disposable template without template version in name (so, fedora-dvm, or even default-dvm, not fedora-xx-dvm)

user/templates/templates.md
```diff
@@ -218,40 +218,58 @@ Please see [How to Reinstall a Template](/doc/reinstall-template/).
...
-    ```
+1. Make the new or upgraded template the default template. In the App Menu, go
+   to Qubes Tools, then click on Qubes Global Settings. In the Qube Defaults
+   section, next to Template, select the new or upgraded template from the
+   drop-down list. Press OK.
+
+2. Base your [disposable templates](/doc/glossary/#disposable-template) on the
+   new template. In the App Menu, go to Qubes Tools, then click on Qube
+   Manager. In the Qube Manager, find your disposable templates. (By default,
+   they end in `-dvm`.) Right click, then click on Settings. Next to Template,
```
marmarek, line 234: You can change template directly from right-click menu - there is "template" submenu.

user/templates/templates.md
```diff
@@ -218,40 +218,58 @@ Please see [How to Reinstall a Template](/doc/reinstall-template/).
...
+     (Warning: This is just an example. Do not attempt to use it.)
+
+     ```
+     qvm-shutdown --wait --all; qvm-prefs fedora-01-dvm template fedora-02; qvm-start sys-usb
+     ```
+
+3. Base your app qubes on the new template. In the App Menu, go to Qubes Tools,
+   then click on Qubes Template Manager. Select all the templates based on the
+   old template (shown in the Current column). Next to "Change all selected
+   to," select the new or upgraded template from the drop-down list. Press OK.
```
marmarek, line 272: This works, but since user has Qube Manager open already, one can sort by template, select all with the old template, and change it from right click menu.

marmarek (approved):
Approving as is, but it can be improved as commented below.

user/templates/templates.md
andrewdavidwong: Won't that require way too much clicking around if the user has a lot of templates to change? I thought that's the problem the Template Manager tool was created to address. Or am I missing something?

andrewdavidwong:
> I don't have a better idea for R4.1. But I have some ideas what we can improve in R4.2. Specifically:
> 
> * add deferred template change (allow changing template while VM is running, but actually do the change when VM shutdown/reboot) - I thought we have an issue for that already, but can't find right now
> * create default disposable template without template version in name (so, fedora-dvm, or even default-dvm, not fedora-xx-dvm)

Yes, both sound like great ideas. The second one sounds like a very simple "no code" solution.

user/templates/templates.md
andrewdavidwong: > That's true, if you want to change a lot of templates (not just a lot of VMs), then Template Manager is the right tool for the job.

Hm, I don't understand. How does changing a lot of VMs not require tons of clicking without the Template Manager?

marmarek: Because you can select multiple qubes and change their template at once. If you sort by template, then select first, and then last with shift.

andrewdavidwong: Oh, I didn't realize you could do that! :exploding_head:

user/templates/templates.md
andrewdavidwong: Discovered a problem with this method: https://github.com/QubesOS/qubes-issues/issues/8074.

andrewdavidwong: Ok, I think I was able to rewrite the instructions in a way that avoids triggering this bug.

PR #1302, opened 2023-04-13
Emergency backup restore v4 tweaks

rustybird:


user/how-to-guides/backup-emergency-restore-v4.md
```diff
@@ -66,44 +66,33 @@ any GNU/Linux system.
...
+ 1. Untar the backup metadata from the main backup file.
 
- 2. Untar the main backup file.
-
-        [user@restore ~]$ tar -i -xvf qubes-backup-2015-06-05T123456
+        [user@restore ~]$ tar -i -xvf qubes-backup-2015-06-05T123456 \
+            backup-header backup-header.hmac qubes.xml.000.enc
         backup-header
         backup-header.hmac
         qubes.xml.000.enc
```
andrewdavidwong, line 86: Is the indentation inconsistency here intentional?

andrewdavidwong, line 86: Is leaving these lines intentional?

andrewdavidwong: Oh, I see now. You're manually specifying the three files that should be extracted. In that case, shouldn't there be a `>` on the second line?

```
$ tar -i -xvf qubes-backup-2023-04-13T163220 \
> backup-header backup-header.hmac qubes.xml.000.enc
```

user/how-to-guides/backup-emergency-restore-v4.md
```diff
@@ -66,44 +66,33 @@ any GNU/Linux system.
 
         [user@restore ~]$ sudo dnf install rpmdevtools
 
- 5. Extract the `scrypt` binary from the RPM.
+ 5. Extract the `scrypt` binary from the RPM and make it conveniently
+    available.
 
         [user@restore ~]$ rpmdev-extract scrypt-*.rpm
+        [user@restore ~]$ alias scrypt="$PWD/scrypt-*/usr/bin/scrypt"
```
andrewdavidwong, line 73: This doesn't work for me in a Debian 11 minimal qube:

```
$ ls scrypt-1.2.1-2 
scrypt-1.2.1-2
$ alias scrypt="$PWD/scrypt-*/usr/bin/scrypt"
[follow instructions]
$ echo "backup-header!$backup_pass" |\
    scrypt dec -P backup-header.hmac backup-header.verified && \
    diff -qs backup-header backup-header.verified
bash: /home/user/scrypt-*/usr/bin/scrypt: No such file or directory
$ alias
alias scrypt='/home/user/scrypt-*/usr/bin/scrypt'
```

andrewdavidwong: Simply doing something like `alias scrypt="$PWD/scrypt-1.2.1-2"` works fine, though.

user/how-to-guides/backup-emergency-restore-v4.md
```diff
@@ -128,36 +117,53 @@ any GNU/Linux system.
...
 
     If this pipeline fails, it is likely that the backup is corrupted or has
     been tampered with.
 
     **Note:** If your backup was compressed with a program other than `gzip`,
     you must substitute the correct compression program in the command above.
-    This information is contained in `backup-header` (see step 5). For example,
+    This information is contained in `backup-header` (see step 4). For example,
     if your backup is compressed with `bzip2`, use `bzip2 -d` instead of `gzip
     -d` in the command above.
```
andrewdavidwong, line 139: FWIW, `bzip2` is not natively present in the Debian 11 minimal qube in which I'm testing these instructions, but that might be because it's minimal. I don't have a non-minimal VM handy to check.

```
$ echo "$backup_id!qubes.xml.000!$backup_pass" |scrypt dec -P qubes.xml.000.enc | bzip2 -d | tar -xv
bash: bzip2: command not found
tar: This does not look like a tar archive
tar: Exiting with failure status due to previous errors
```

user/how-to-guides/backup-emergency-restore-v4.md
```diff
@@ -128,36 +117,53 @@ any GNU/Linux system.
...
 
- 8. Enter the decrypted directory, mount `private.img`, and access your data.
+ 7. Search inside of `qubes.xml` for the `backup-path` property of the qube
+    whose data you wish to restore. Using the value of this property (e.g.
```
andrewdavidwong, line 142: I think we need to provide instructions for this part (i.e., how exactly to "search inside of `qubes.xml` for the `backup-path` property) and probably break it up into separate steps.

I'm guessing it should be something like this:

```
$ grep 'backup-path' qubes.xml
        <feature name="backup-path">vm75/</feature>
```

In this test run, my backup contains only a single VM, which I'm guessing is why this returned only one result.

If the backup contains multiple qubes, the user will naturally wonder, at this point, how she's supposed to know which `vm##` corresponds to the qube she wants. For example, "Okay, what's the number for my `work` qube?"

user/how-to-guides/backup-emergency-restore-v4.md
rustybird: Oooh you probably got that `scrypt-1.2.1-2` binary directly from your backup, not by running `rpmdev-extract scrypt-*.rpm`?

user/how-to-guides/backup-emergency-restore-v4.md
rustybird: > In that case, shouldn't there be a `>` on the second line (and likewise for other places where commands are broken up across multiple lines using `\`)?

That might cause some people to type in the `>` though

user/how-to-guides/backup-emergency-restore-v4.md
rustybird: Right, only the full templates have it by default. I've added a sentence about installing the package.

user/how-to-guides/backup-emergency-restore-v4.md
rustybird: The *implied* instructions are something like

"Open `qubes.xml` in a text viewer. As XML is a famously readable format, you will intuitively grasp its overall structure and notice multiple `domain` elements (a domain being a VM, which is of course a qube). Find your qube `foo` by looking for a `domain` element containing a descendant element `<property name="name">foo</property>`. The `domain` element will also contain a descendant element\* `<feature name="backup-path">vm123/</feature>`. We will not insult you by writing down any part of this."

\* Omg it isn't even a "property" but a "feature" :facepalm:

> (I know this is a preexisting problem and wasn't introduced by your changes, but these new instructions make it more salient to me now for some reason.)

Besides making it more visible, the changes also made UX worse when the backup file only contains a single qube anyway. Previously it was possible to ignore `qubes.xml` in that case, and just go with the only extracted `vm123/` directory.

Ugh. Here's an (already unpleasantly long) command that produces a list of every VM in `qubes.xml`, plus the directory name for those that are included in the backup:

```
$ xmlstarlet sel -T -t -m //domain -v 'concat(.//property[@name="name"], " ", .//feature[@name="backup-path"])' -n qubes.xml
```

user/how-to-guides/backup-emergency-restore-v4.md
andrewdavidwong: It might, but it might also cause people to wonder why theirs has a `>` when the instructions don't and try to get rid of theirs (which they won't be able to). Is this a distro-specific convention or something common to all Linux environments? If the former, we can exclude it, but if the latter, we should include, IMHO.

user/how-to-guides/backup-emergency-restore-v4.md
andrewdavidwong: I already had that binary. I've had it for so long that I can't remember how I originally got it. It could have very well originally come from running `rpmdev-extract scrypt-*.rpm`, but I can't recall, as it was years ago.

user/how-to-guides/backup-emergency-restore-v4.md
andrewdavidwong: > "[...]We will not insult you by writing down any part of this."

IMHO, that's the opposite attitude of the one we should take regarding these sorts of instructions, where the context is that the user has an emergency (by assumption, as these are *emergency* restore instructions) and is being walked through an unfamiliar procedure step-by-step, with no assumed skill level.

(It's kind of like the movie trope where a surgeon has to tell a panicked layperson how to perform surgery over the phone, or printing "in case of emergency, break glass" on a glass-encased object instead of assuming that people will figure it out in the moment. Better to over-explain even what seems obvious, just in case.)

rustybird: What, you don't agree that "draw the rest of the owl" is fine as a load-bearing step in data recovery, and that it would be gauche to assume that the reader hasn't already built a workable XML parser in their head? (I'm putting an `/s` here this time)

Latest commit has the `xmlstarlet` command.

user/how-to-guides/backup-emergency-restore-v4.md
rustybird: Seems to be pretty much universal. I'm not aware of any distro changing the secondary prompt string to something differnt.

One sharp edge might be that `>` is also the file redirection operator, so if `> foo bar baz` is entered as an independent line when the shell isn't in secondary prompt mode (e.g. forgot to type in the previous `\`) it would overwrite the file `foo`.

I can't tell anymore, multiline commands are weird. You decide :) Maybe after moving things back to #1279?

user/how-to-guides/backup-emergency-restore-v4.md
rustybird: Makes sense. If you aren't running the `rpmdev-extract` step at the moment (which would have created a directory tree `scrypt-VERSION/usr/bin/` and put the executable named `scrypt` inside of it) you have to place the differently obtained executable into an equivalent directory tree or adjust the alias to refer to where it ended up.

user/how-to-guides/backup-emergency-restore-v4.md
andrewdavidwong: What if we just didn't try to break up commands across multiple lines using `\` and instead left everything on a single line? If some commands seem too long for that, what if we directed the user to create a shell script file for them?

user/how-to-guides/backup-emergency-restore-v4.md
andrewdavidwong: Since we explicitly recommend that users keep a copy of the `scrypt` binary with their backups, I think we should assume that at least some of them will follow that advice, and therefore we shouldn't assume that they're doing `rpmdev-extract` at the moment. (We should consider, for example, that they might be trying to follow this procedure completely offline, with no access to the internet, just a saved copy of these instructions, their backup, and whatever binaries they have saved or installed on their local system, which might just be some old Ubuntu disc.)

user/how-to-guides/backup-emergency-restore-v4.md
andrewdavidwong: "Rest of the owl" is exactly what I thought of too. :joy:

andrewdavidwong: > Here's an (already unpleasantly long) command that depends on the `xmlstarlet` package and produces a list of every qube in `qubes.xml`, plus the directory name for those that are included in the backup: [...]

I like the neat and easily-readable output of this, but if we're concerned about the need for another uncommon package and a long command, what about something like this?

List all qubes included in the backup:

```
$ grep -B 14 'backup-path' qubes.xml | grep 'name="name"'
        <property name="name">work</property>
        <property name="name">untrusted</property>
        <property name="name">vault</property>
```

Find the QID for the one you want to restore:

```
$ grep -A 2 'vault' qubes.xml | grep 'qid'
        <property name="qid">75</property>
```

Then just add `vm` to the front and `/` to the end, so `vm75/`?

This way, we don't have to recommend that users store a copy of `xmlstarlet` with their backups just to be able to parse XML, and the commands are easier to type and understand. What do you think?

user/how-to-guides/backup-emergency-restore-v4.md
rustybird: That would feel the most natural to me. Might be tricky to get this to render ok on the website, if the long commands should have soft line wraps (not a mile-wide line to scroll through horizontally, but no hard line breaks either that would break copy and paste) yet still be distinguished from the output.

user/how-to-guides/backup-emergency-restore-v4.md
rustybird: That whole section might need a rewrite too, if it's to be reliable in both the "obtain necessary tools right now" and "(1) download them all in advance, (2) make them available when it's time" scenarios. E.g. that `scrypt` binary is dynamically linked so it depends on a compatible version of the OpenSSL libraries being installed in the rescue environment. (Hmm, maybe it should be statically linked)

user/how-to-guides/backup-emergency-restore-v4.md
rustybird: Adding yet another dependency on xmlstarlet (plus its own dependencies) does suck. Although it seems to be more widely available through package managers than scrypt.

I'm not sure yet if it's possible to rely on the order and format in which XML elements are written out to `qubes.xml` across all versions of the code that generates v4 backups. That's always an annoying part of interacting with XML on a string level.

Currently the `grep` commands are also loose, like "vault" would match "anothervault" or "vault2" as well. I'd guess that at the point where the commands have been narrowed down to avoid most bycatch, they'll be even gnarlier than the `xmlstarlet` command line. In comparing the two approaches it might come down to avoiding the extra package dependency.

user/how-to-guides/backup-emergency-restore-v4.md
andrewdavidwong: > Might be tricky to get this to render ok on the website, if the long commands should have soft line wraps (not a mile-wide line to scroll through horizontally, but no hard line breaks either that would break copy and paste) yet still be distinguished from the output.

I'm afraid I don't have a solution for that part, except using in-line code snippets (i.e., wrap the command in single backticks) rather than putting it in a code block (i.e., having lines with triple backticks before and after).

user/how-to-guides/backup-emergency-restore-v4.md
andrewdavidwong: Darn. :\

Maybe we should just provide the user with general instructions for using a text editor to visually read the XML and find what they need.

@marmarek, do you have any suggestions about this? (At the very least, this seems like something to keep in mind for any future design changes to the backup system.)

andrewdavidwong: Reposting my comment from [this discussion](https://forum.qubes-os.org/t/19395/):

> @rustybird: For what it's worth, if pushing the new instructions through even with the added `xmlstarlet` dependency is the lesser of two evils, I won't stand in the way. I prefer to avoid extra dependencies where feasible (to avoid exactly the "special software required"/"vendor lock-in" perception discussed above), but when you need 'em, you need 'em (like with `scrypt`).

rustybird: Yeah all in all `xmlstarlet` doesn't seem too bad to me.

I've added two more small commits incorporating some of the [feedback from @/tree in the forum](https://forum.qubes-os.org/t/getting-stuck-on-emergency-backup-recovery/18520/5).

How do you feel about setting aside other orthogonal improvements, like rewriting how to properly gather various package dependencies in advance (big can of worms...) and reformatting code blocks?

andrewdavidwong: That's also okay with me. I don't want to allow the perfect to be the enemy of the good.

rustybird: @marmarek This might have fallen through the cracks. Can you give us a yay or nay? I always feel weird linking people in the forum to my little branch instead of the official instructions.

unman:
@andrewdavidwong Why reopen?
I think this version is (relatively) clear, and it's time it was
merged.
I see suggestions in the comments that some areas need rewriting or
further work, but *this* version stands.

unman:
Hmm, #1302 HAS been merged, so I think we can safely close it now.

PR #1312, opened 2023-04-28
Write 'ITL' in full in vm-sudo.md

imme-emosol:


andrewdavidwong:
Actually, it's kind of weird to mention ITL here at all. I think what the original author meant was really something like "the Qubes developers."

PR #1314, opened 2023-05-03
add how-to-open-urls-in-other-qubes

taradiddles:
following up https://github.com/QubesOS/qubes-issues/issues/8166#issuecomment-1529972761

@andrewdavidwong

- feel free to edit out stuff, it might be faster than back and forth comments.
- my English phrasing/grammar is sometimes er... weird, so don't worry to fix whatever you want.
- doc `ref` number: TBD
- before delving into the details let me know if some sections should move back to the original community doc (I tried to reflect [your post](https://github.com/QubesOS/qubes-issues/issues/8166#issuecomment-1529957223))
- I updated the RPC policies to 4.x and tested them but it would be nice to have someone else confirm I haven't made any glaring mistake.
- once the doc is final I'll update the community doc as well.

### DIFF
--- a/user/how-to-guides/how-to-open-urls-in-other-qubes.md
+++ b/user/how-to-guides/how-to-open-urls-in-other-qubes.md
@@ -1,0 +1,174 @@
+---
+lang: en
+layout: doc
+permalink: /doc/how-to-open-urls-in-other-qubes/
+ref: TBD
+title: How to open URLs/files in other qubes
+---
+
+*This page is about opening untrusted files and URLs from "secure" offline or
+firewalled qubes in "general-purpose" qubes - or any qubes you see fit. The
+simplest solution is to [copy/paste](/doc/how-to-copy-and-paste-text/) URLs or
+[copy files](/doc/how-to-copy-and-move-files/) between qubes and manually open
+the resource in the destination qube. However this approach is error-prone so
+using formal RPC policies like described below is preferable.*
+
+Naming convention:
+
+- `srcQube` is the qube where the files/URLs are
+- `dstQube` is the qube we want to open them in
+
+## Configuring RPC policies
+
+The `qvm-open-in-vm` and `qvm-open-in-dvm` scripts are invoked in a qube to
+open files and URLs in another qube. Those scripts make use of the the
+`qubes.OpenInVM` and `qubes.OpenURL` [RPC
+services](/doc/qrexec/#qubes-rpc-services). Qubes [RPC
+policies](/doc/rpc-policy/) control which RPC services are allowed between
+qubes.
+
+Policy files are in `/etc/qubes/policy.d/`.
+
+### Using the `ask` action
+
+This action displays a selection widget with the list of allowed destination
+qubes each time the associated RPC service is called. This setup makes it
+possible to always control if, and on which qube and network (eg. "clearnet",
+TOR, VPN) an URL is requested or file opened.
+
+The selected qube will autostart if it wasn't running. 
+
+Note: when using `ask`, the destination qube given as argument to
+`qvm-open-in-vm` is ignored if no `allow` rule matches the current RPC service
+and source/destination qubes.
+
+### Using the `allow` action
+
+This action allows a given RPC service and source/destination qubes without
+prompting the user.
+
+When an `allow` action is defined for a destination other than `@dispvm`, the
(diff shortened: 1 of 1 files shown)


user/how-to-guides/how-to-open-urls-in-other-qubes.md
```diff
@@ -1,174 +1,105 @@
 ---
 lang: en
 layout: doc
-permalink: /doc/how-to-open-urls-in-other-qubes/
-ref: TBD
```
andrewdavidwong: No need to worry about this. There's a script that will assign an appropriate number and add this line automatically.

user/how-to-guides/how-to-open-urls-in-other-qubes.md
```diff
@@ -1,174 +1,105 @@
...
-title: How to open URLs/files in other qubes
+permalink: /doc/how-to-open-urls-and-files-in-other-qubes/
+title: How to open URLs and files in other qubes
 ---
 
-*This page is about opening untrusted files and URLs from "secure" offline or
-firewalled qubes in "general-purpose" qubes - or any qubes you see fit. The
```
andrewdavidwong: I don't think this is really limited to the "secure/offline -> general/online" use case, so I generalized it.

user/how-to-guides/how-to-open-urls-in-other-qubes.md
```diff
@@ -1,174 +1,105 @@
...
-simplest solution is to [copy/paste](/doc/how-to-copy-and-paste-text/) URLs or
-[copy files](/doc/how-to-copy-and-move-files/) between qubes and manually open
-the resource in the destination qube. However this approach is error-prone so
-using formal RPC policies like described below is preferable.*
```
andrewdavidwong: I don't think it's quite accurate to say that this approach is "error-prone." There is nothing about the process itself that makes errors likely.

andrewdavidwong:
(Side note: Don't worry about the text wrapping. I independently decided today to try out an unwrapped text policy. You did the right thing.)

user/how-to-guides/how-to-open-urls-in-other-qubes.md
```diff
@@ -1,174 +1,105 @@
...
-
-~~~
-qvm-create --class DispVM --label red --template fedora-dvm qube_name
-~~~
-
-`qube_name` would then be started like a regular qube with the difference that
-like standard disposable qubes its private disk is wiped after each use.
-However, unlike standard disposable qubes, this qube won't "auto power off" when
-the called application (eg. firefox) is closed, so it's up to the user to power
-off (or restart) the qube when he/she deems necessary.
```
andrewdavidwong: I think this part is out-of-scope on this page. It belongs on one of the disposable doc pages instead. (If it's already be covered there, then we should [avoid duplicating documentation](https://www.qubes-os.org/doc/documentation-style-guide/#do-not-duplicate-documentation).)

andrewdavidwong:
> * I updated the RPC policies to 4.x and tested them but it would be nice to have someone else confirm I haven't made any glaring mistake.

Perhaps @marmarek can have a look at it during review.

user/how-to-guides/how-to-open-urls-in-other-qubes.md
taradiddles: > There is nothing about the process itself that makes errors likely.

I'm a heavy user of inter-vm copy/paste and I've pasted content in the wrong vm *many* times, including important passwords that I would then have to change - usually because of inattention but sometimes because a window from another vm would pop up at the same moment I was ctrl-shift-v'ing. That's what I meant with "error-prone" and "preferably" using an automated RPC policy.
(that's just a FYI - you did a very good job reformating/fixing the doc!).

andrewdavidwong: > sometimes because a window from another vm would pop up at the same moment I was ctrl-shift-v'ing. That's what I meant with "error-prone" and "preferably" using an automated RPC policy.

Ah, but that's not a problem with the global clipboard. That's a problem with your window manager settings allowing windows to pop up in ways you don't want. The exact same problem can and does occur on non-Qubes OSes with regular copy/paste (and perhaps even more commonly with mouse clicks).

PR #1321, opened 2023-05-14
Update installation guide to be compliant with Qubes OS 4.1.2

aronowski:
Installation guide has been updated so it reflects the process of installing Qubes OS 4.1.2 as well as some minor improvements, like mentioning the importance of glossary.


kennethrrosen (approved):
Very good. Screengrabs also look clean.

PR #1322, opened 2023-05-18
Update commands used to uninstall templates

FreeOni:
The 'dnf remove' command no longer removes templates and mentions "No Match" when using \<package-name-spec\>. During the template *installation* procedure using 'qubes-dom0-update' the output states the command is being redirected to qvm-template. The 'dnf remove' command lacks this redirection. Rather than implement the redirection, the documentation was instead updated.

### DIFF
--- a/user/templates/templates.md
+++ b/user/templates/templates.md
@@ -158,32 +158,6 @@ $ qvm-template list --installed
 
 In either case, if another qube is based on the template, the template will remain installed and a list of the dependent qubes will be displayed. [Switch](#switching) the dependent qubes to another template before attempting the removal again.
 
-You may see warning messages like the following:
-
-```
-warning: file /var/lib/qubes/vm-templates/fedora-XX/whitelisted-appmenus.list: remove failed: No such file or directory
-warning: file /var/lib/qubes/vm-templates/fedora-XX/vm-whitelisted-appmenus.list: remove failed: No such file or directory
-warning: file /var/lib/qubes/vm-templates/fedora-XX/root.img.part.04: remove failed: No such file or directory
-warning: file /var/lib/qubes/vm-templates/fedora-XX/root.img.part.03: remove failed: No such file or directory
-warning: file /var/lib/qubes/vm-templates/fedora-XX/root.img.part.02: remove failed: No such file or directory
-warning: file /var/lib/qubes/vm-templates/fedora-XX/root.img.part.01: remove failed: No such file or directory
-warning: file /var/lib/qubes/vm-templates/fedora-XX/root.img.part.00: remove failed: No such file or directory
-warning: file /var/lib/qubes/vm-templates/fedora-XX/netvm-whitelisted-appmenus.list: remove failed: No such file or directory
-warning: file /var/lib/qubes/vm-templates/fedora-XX/icon.png: remove failed: No such file or directory
-warning: file /var/lib/qubes/vm-templates/fedora-XX/clean-volatile.img.tar: remove failed: No such file or directory
-warning: file /var/lib/qubes/vm-templates/fedora-XX/apps.templates: remove failed: No such file or directory
-warning: file /var/lib/qubes/vm-templates/fedora-XX/apps.tempicons: remove failed: No such file or directory
-warning: file /var/lib/qubes/vm-templates/fedora-XX/apps: remove failed: No such file or directory
-warning: file /var/lib/qubes/vm-templates/fedora-XX: remove failed: No such file or directory
-```
-
-These are normal and expected. Nothing is wrong, and no action is required to
-address these warnings.
-
-If the uninstallation command doesn't work, pay close attention to
-any error message: it may tell you what qube is using the template,
-or if the template is default. In other cases, please see [VM Troubleshooting](/doc/vm-troubleshooting/).
-
 If the Applications Menu entry doesn't go away after you uninstall a template,
 execute the following type of command in dom0:
 
--- a/user/templates/templates.md
+++ b/user/templates/templates.md
@@ -144,7 +144,7 @@ Please see [How to Install Software](/doc/how-to-install-software).
 
 ## Uninstalling
 
-To remove a template, the graphical `Qube Manager` (Qubes Menu > Qubes Tools > Qube Manager) may be used. Right-click the template to be uninstalled and click "Delete qube" to begin removal.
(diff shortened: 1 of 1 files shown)


user/templates/templates.md
```diff
@@ -155,22 +155,24 @@ and also that it is not set as the default template.
...
-[above](#installing), then you must execute the following type of command in
-dom0 in order to uninstall it:
+If, on the other hand, the template came pre-installed or was installed through
+a template package in dom0 using the instructions [above](#installing), then you
+must execute the following commands in dom0 in order to uninstall it.
 
+List the installed templates:
 ```
-$ sudo dnf remove qubes-template-<DISTRO_NAME>-<RELEASE_NUMBER>
+$ qvm-template list --installed | awk -F " " '{print $1}'
```
rapenne-s, line 169:
```suggestion
$ qvm-template list --installed | awk 'NF==3 {print $1}'
```

the default field separator is a whitespace, so no need to add `-F " "`. In addition, `qvm-template list --installed` adds a `Template installed` at the end of its output when piping into a command (and without a pipe, it's the first line :thinking: ), so this could be hidden by displaying lines with 3 fields with `NF == 3`, or using an extra `grep -v ^Installed`.

FreeOni: Thank you for confirming the moving "Template installed" title. This patch:
https://github.com/QubesOS/qubes-core-admin-client/pull/245
should resolve it allowing a simpler awk command. Your suggestions helped here.

rapenne-s: nice, it would even be better to not display the header if you use `--installed` or `--available` because that's not really useful in the output, but it seems to be more work than your fix.

With that change `| awk 'NR>1 { print $1 }'` should work to discard the first line.

FreeOni: Added the pattern to skip outputting the redundant title. Thank you for the review.

rapenne-s (approved):

user/templates/templates.md
```diff
@@ -155,22 +155,24 @@ and also that it is not set as the default template.
...
-[above](#installing), then you must execute the following type of command in
-dom0 in order to uninstall it:
+If, on the other hand, the template came pre-installed or was installed through
+a template package in dom0 using the instructions [above](#installing), then you
+must execute the following commands in dom0 in order to uninstall it.
 
+List the installed templates:
 ```
-$ sudo dnf remove qubes-template-<DISTRO_NAME>-<RELEASE_NUMBER>
+$ qvm-template list --installed | awk 'NR>1 {print $1}'
```
unman, line 169: The awk statement doesnt work as you hope to suppress the "Installed" line.
In any case, I would prefer not to use it at all - we dont want users to blindly follow code, and the output from `qvm-template list --installed` is understandable as it is.

user/templates/templates.md
```diff
@@ -155,22 +155,24 @@ and also that it is not set as the default template.
...
 right-click on the template and select **Delete qube**. (If you're not sure,
 you can safely try this method first to see if it works.)
 
-If, on the other hand, the template came pre-installed or was installed by
-installing a template package in dom0, per the instructions
-[above](#installing), then you must execute the following type of command in
-dom0 in order to uninstall it:
+If, on the other hand, the template came pre-installed or was installed through
+a template package in dom0 using the instructions [above](#installing), then you
+must execute the following commands in dom0 in order to uninstall it.
```
unman, line 165: ` template package` is pretty clumsy - most users wont know that they are using packages.
Why not just say - or was installed using `qvm-template` or the GUI tool  ? ( The installation section needs to be rewritten to reflect this.)

user/templates/templates.md
```diff
@@ -155,22 +155,24 @@ and also that it is not set as the default template.
 
 The procedure for uninstalling a template depends on how it was created.
 
-If the template was originaly created by cloning another template, then you can
+If the template was originally created by cloning another template, then you can
 delete it the same way as you would any other qube. In the Qube Manager,
 right-click on the template and select **Delete qube**. (If you're not sure,
 you can safely try this method first to see if it works.)
```
unman, line 161: Do you want to mention `qvm-remove`?

unman (changes requested):

FreeOni:
Rewrote the patch around the `qvm-remove` command. Still evaluating if the rest of the “Uninstalling” section is needed now that we have moved away from directly using the `dnf` command.

user/templates/templates.md
rapenne-s: I'm not sure it would be worth mentioning `qvm-remote`. The documention gives one path that suits every user: using the GUI.

For advanced users, I wonder if a list of Qubes OS specific commands and tooling exist somewhere? This class of user may want to seek that kind of information when dealing with a problem in order to automate it or run in the console. That's why I don't think it's worth a mention here.

FreeOni:
The new patches:
- Rewrite the section around the unified `qvm-template` command.
- Remove the pre-`qvm-template` `dnf remove` warnings.
- Reduce duplicate documentation with a link.

rapenne-s:
I like it! It's clear and simple.

I'd suggest moving the Graphical UI part first and the CLI after. This would make the documentation friendlier to non-tech people using Qubes OS.

FreeOni:
GUI instructions are now given first with CLI following.

user/templates/templates.md
```diff
@@ -144,68 +144,21 @@ Please see [How to Install Software](/doc/how-to-install-software).
...
-If the template was originaly created by cloning another template, then you can
-delete it the same way as you would any other qube. In the Qube Manager,
-right-click on the template and select **Delete qube**. (If you're not sure,
-you can safely try this method first to see if it works.)
-
-If, on the other hand, the template came pre-installed or was installed by
-installing a template package in dom0, per the instructions
-[above](#installing), then you must execute the following type of command in
-dom0 in order to uninstall it:
+To remove a template, the graphical `Qube Manager` (Qubes Menu > Qubes Tools > Qube Manager) may be used. Right-click the template to be uninstalled and click "Delete qube" to begin removal.
```
rapenne-s, line 147:
```suggestion
To remove a template, the graphical `Qube Manager` (Qubes Menu > Qubes Tools > Qube Manager) may be used. Right-click the template to be uninstalled and click "Delete qube". You will be asked to type the template name in a prompt to confirm the deletion.
```

FreeOni: I feel that mentioning the "typed confirmation" dialog requires mentioning the "issues" dialog as the issues will terminate the workflow before the confirmation dialog appears. An optional fourth patch was added. Thank you.

rapenne-s (approved):

PR #1333, opened 2023-08-10
Add an example for a simple VM layout

GWeck:
Describe the VMs used by a teacher who has no time to build something complicated

Depends on https://github.com/QubesOS/qubes-attachment/pull/66


GWeck:
@fepitre What needs to be ddone to make the checks to run successfully? Thanks for any information!

andrewdavidwong:
> What needs to be done to make the checks run successfully? Thanks for any information!

Are you referring to the part that says "All checks have failed" and below it, `[red X][icon] continuous-integration/pullrequest — Pipeline: failed`?

If so, here's what you need to do:

1. Click the "Details" link to the right of the `[red X][icon] continuous-integration/pullrequest — Pipeline: failed` line.
2. Click on the `[red X] build:doc` button.
3. Read the log to see what failed.

In this case, the log tells us there were two reasons for the failure: new spellings introduced and one link to something that doesn't exist.

GWeck:
The two new spellings are dispVM and dispVMs, which should be o.k. The link points to a file which should be provided by PR #86 to qubes-attachment; so this should be o.k., too, after that PR is executed.

andrewdavidwong:
> The two new spellings are dispVM and dispVMs, which should be o.k. 

Technically, those were never correct (the initial "D" was always capitalized), and those are now deprecated terms:

https://www.qubes-os.org/doc/glossary/#disposable

GWeck:
I changed it accordingly to "disposable".

rapenne-s:
I'd like to mention split-gpg / split-ssh will be doable from a GUI in R4.2 that should be out "soon". I think the text should take this into account, because it's super easy to setup, so even less time spent configuring Qubes OS :)

unman:
Our policy to date has been to have separate sections for each
supported release - this makes it easier to find material relevant
to the user's release and to remove material when release becomes eol.
This text need not take 4.2 changes in to account.

andrewdavidwong:
As the original author of this page, I didn't think of it as being specific to a particular Qubes version when I wrote it, and I tried not to include anything that would become obsolete due to a version upgrade, so I'm not sure if organizing the content by Qubes version makes sense in the context of this page. Should this page really have anything version-specific in it?

andrewdavidwong:
Ah, at first, I thought you were suggesting that this page be divided into 4.1 and 4.2 sections, @unman, but after re-reading the comments on this PR a few more times, I now think that might not be the case, so never mind.

GWeck:
As this example is intended to show just how a beginner, without much knowledge and without much time to acquire such knowledge, might use Qubes productively, I don't think that it should contain anything complicated. When such functions (should) become available, it may make sense to include them in the example, but making the example version-dependent would clearly add a level of complexity that is beyond the understanding of the intended readers.

From my teaching experience, I often get feedback that non-technical people are not interested in any particular properties of some software. Instead, they just want something that you can install and that works reasonably. That is possible with Qubes, but you have to take these users at their hand and somehow lead them on some path through the swamp of possibilities, and this example may be of some help there.

rapenne-s:
I just wanted to share that split-GPG will be *easier* in the forthcoming release, while the last paragraph seems to say it's a "hard" step to get templates and split-gpg working before being free to use Qubes OS :)

unman:
I completely agree with your approach, and your use of this example.

GWeck:
@rapenne-s Perhaps the last paragraph sounds a bit pessimistic, but teaching virtual systems to a group of lawyers is a somewhat traumatic experience. :grin:

PR #1337, opened 2023-09-24
Update answer to "Is Qubes another Linux distribution?"

rl2939:
At the time of this writing, the answer for _Is Qubes just another Linux distribution?_ is this:

> If you really want to call it a distribution, then it's more of a "Xen distribution" than a Linux one. But Qubes is much more than just Xen packaging. It has its own VM management infrastructure, with support for template VMs, centralized VM updating, etc. It also has a very unique GUI virtualization infrastructure.

However that answer was last edited [on September 26, 2013](https://github.com/QubesOS/qubes-doc/commit/a980cf4168cd13b9b19f8548f92c12bc3840551b) and does not match up with later interpretations of what Qubes is (see below). This is important because current answer downplays the importance of Linux, which conflicts with information stated [in an 2014 intererview](https://www.linuxfoundation.org/blog/blog/qubes-the-open-source-os-built-for-security), about a year later.

(italics mine)

> **Linux.com: What is Qubes OS?**
> Joanna Rutkowska: Qubes OS is an open-source operating system designed to provide strong security for desktop computing. Qubes OS implements security by a compartmentalization approach. _It’s based on Xen and Linux_, but also has support for Windows-based AppVMs.

> **How does it use the Xen hypervisor and Linux?**
> Qubes uses Xen as a “compartmentalization provider.” We believe Xen is very powerful for this role. The Xen hypervisor is (still) relatively bloatware-free and had a decent architecture that allows us to keep it reasonably secure (e.g. it allows to keep qemu outside the TCB). It also offers support for so-called driver domains, which Qubes utilizes to sandbox networking and USB stacks.
> Having said that, it is important to remember that Qubes is largely independent of the underlying hypervisor. In the next release (R3) we’re even introducing a Hypervisor Abstraction Layer to allow for (relatively) easy swapping of Xen for other VMMs. So, in the future we might see e.g. a KVM-based variant of Qubes OS, for better hardware compatibility.
> Qubes uses Linux (here I mean both the kernel and the platform) mostly as an… “API provider” for apps and drivers. We don’t utilize the security mechanisms Linux provides, and instead we rely on Xen isolation for that.
> This statement might sound like downplaying the importance of Linux, which, however is not my intention. _There would be no Qubes OS without the flexibility of Linux._ Even though we now also have support for Windows-based AppVMs, Linux is still unbeatable when it comes to making lightweight user AppVMs and sophisticated service VMs, or fueling Dom0 — the admin domain of Qubes OS.

This pull request changes the response to make it more clear that Qubes is an Linux and Xen hybrid rather than just an (mainly) Xen one.

### DIFF
--- a/introduction/faq.md
+++ b/introduction/faq.md
@@ -170,9 +170,10 @@ Please see this [article](https://blog.invisiblethings.org/2012/09/12/how-is-qub
 
 ### Is Qubes just another Linux distribution?
 
-If you really want to call it a distribution, then it's more of a "Xen distribution" than a Linux one.
-But Qubes is much more than just Xen packaging.
-It has its own VM management infrastructure, with support for template VMs, centralized VM updating, etc.
+If you really want to call it a distribution, then it's more of a "hybrid Linux and Xen distribution" than a pure Linux one. 
+Qubes uses Linux as an "API provider" for its apps and drivers, and it uses Xen to keep the operating system secure. 
+It uses But Qubes is much more than just Linux and Xen packaging. 
+It has its own VM management infrastructure, with support for template VMs, centralized VM updating, etc. 
 It also has a very unique GUI virtualization infrastructure.
 
 ### What about safe languages and formally verified microkernels?


introduction/faq.md
```diff
@@ -170,9 +170,10 @@ Please see this [article](https://blog.invisiblethings.org/2012/09/12/how-is-qub
 
 ### Is Qubes just another Linux distribution?
 
-If you really want to call it a distribution, then it's more of a "Xen distribution" than a Linux one.
-But Qubes is much more than just Xen packaging.
-It has its own VM management infrastructure, with support for template VMs, centralized VM updating, etc.
+If you really want to call it a distribution, then it's more of a "hybrid Linux and Xen distribution" than a pure Linux one. 
+Qubes uses Linux as an "API provider" for its apps and drivers, and it uses Xen to keep the operating system secure. 
+It uses But Qubes is much more than just Linux and Xen packaging.
```
unman, line 175: Something wrong here

unman:
@marmarek , @fepitre ,@DemiMarie - any  comments?

I'm not inclined to make this change.
In the absence of any input I will close.

fepitre:
Even for a tech person it looks like more confusing than the original version. You can close it.

unman:
Good enough for me.

PR #1340, opened 2023-10-14
Added ntp package hint

gordon-shumway-net:
As also discussed in the QubesOS Forum[0], here the hint for systemd-timesyncd (or other NTP package) is added.

[0]https://forum.qubes-os.org/t/definitive-guide-to-time-setting-sync-in-qubes-os/17641/13


unman:
I'm not sure about this.
If the package is required then it should be pulled in as a requirement of the official Qubes packages, and should be dealt with there.
I'll need to check that - AFAIR was timesyncd was explictly disabled and then removed from Debian templates.

PR #1343, opened 2023-10-30
Remove deprecated external community doc links

andrewdavidwong:
See the discussion at: https://forum.qubes-os.org/t/21746/

Companion PR: https://github.com/QubesOS/qubesos.github.io/pull/236


andrewdavidwong:
I couldn't figure out how to programmatically replace every community GitHub link with a link to <https://forum.qubes-os.org/c/guides/14>, so I just removed them instead. However, I realized after submitting this PR that this will break all existing links to `https://www.qubes-os.org/doc/*` that were being redirected to those GitHub pages, so this is definitely not the best way to do things. Hopefully someone more skilled can do a better version of this, and this PR never has to be merged.

marmarek:
> I couldn't figure out how to programmatically replace every community GitHub link with a link to https://forum.qubes-os.org/c/guides/14,

Is it okay to replace all with this very link (to the category, not individual guides)? If so, I can make a PR.

unman:
It's definitely better to use specific links, and links that will
generate "good" search results - that means full titles since they
incorporate the likely search terms, rather than short, but working,
forum links.
The same goes for redirects from the old GitHub community pages - these
should redirect automatically to the full titled pages, and also
incorporate a text link in case the redirect should fail for some
reason..

marmarek:
A bit better approach here: https://github.com/QubesOS/qubes-doc/pull/1346

PR #1344, opened 2023-11-03
Switch the Firewall guide to nftables

rapenne-s:
This is a rewrite of the Firewall guide to use nftables instead of iptables

### DIFF
--- a/user/security-in-qubes/firewall.md
+++ b/user/security-in-qubes/firewall.md
@@ -266,18 +266,23 @@ You can get this information using various methods, but only the first one can b
 - in the Qubes Manager window using the column IP
 - from the Settings Window for the qube
 
-Note the IP addresses you will need.
+Note the IP addresses you will need, they will be required in the next steps.
+
 > Note: The vifx.0 interface is the one used by qubes connected to this netvm so it is _not_ an outside world interface.
 
 **2. Route packets from the outside world to the FirewallVM**
 
 For the following example, we assume that the physical interface ens6 in sys-net is on the local network 192.168.x.y with the IP 192.168.x.n, and that the IP address of sys-firewall is 10.137.1.z.
 
-In the sys-net VM's Terminal, the first step is to to define an ntables chain that will receive DNAT rules, we recommend to define a new chain for each destination qubes, this ease the rules management:
+In the sys-net VM's Terminal, the first step is to to define an ntables chain that will receive DNAT rules to relay the network traffic on a given port to the qube NetVM, we recommend to define a new chain for each destination qube to ease rules management:
 
 ```
 nft add chain qubes custom-dnat-qubeDEST '{ type nat hook prerouting priority filter +1 ; policy accept; }'
 ```
+
+> Note: the name `custom-dnat-qubeDST` is arbitrary
+
+> Note: while we use a DNAT chain for a single qube, it's totally possible to have a single DNAT chain for multiple qubes
 
 Second step, code a natting firewall rule to route traffic on the outside interface for the service to the sys-firewall VM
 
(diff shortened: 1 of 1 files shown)


andrewdavidwong:
Should this have `Closes QubesOS/qubes-issues#8641`?

marmarek:
Can we keep iptables version somewhere temporarily? Like, make a copy
and add to index with "R4.1 version" comment? With version-specific docs
it will be much easier (as those would be in separate branches), but for
now we can do a workaround like this.

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

user/security-in-qubes/firewall.md
```diff
@@ -242,270 +240,224 @@ When the qube `unstrusted` has started (after a first reboot), you can directly
...
-```
+> Note: while we use a DNAT chain for a single qube, it's totally possible to have a single DNAT chain for multiple qubes
 
-> If you want to expose the service on multiple interfaces, repeat the steps described in part 1 for each interface.
-> In Qubes R4, at the moment ([QubesOS/qubes-issues#3644](https://github.com/QubesOS/qubes-issues/issues/3644)), nftables is also used which imply that additional rules need to be set in a `qubes-firewall` nft table with a forward chain.
+Second step, code a natting firewall rule to route traffic on the outside interface for the service to the sys-firewall VM
 
-`nft add rule ip qubes-firewall forward meta iifname eth0 ip daddr 10.137.1.z tcp dport 443 ct state new counter accept`
+```
+nft add rule qubes custom-dnat-qubeDEST iifname == "ens6" ip saddr 192.168.x.y/24 tcp dport 443 ct state new,established,related counter dnat 10.137.1.z
```
marmarek, line 290: For a single specific interface it's better to use `iif` (it uses interface index internally, so it's slightly faster). Similar comment in few places below.

marmarek (changes requested):
Besides minor comment, it looks good :)
As for the question above, we can do it either in this PR, or outside when merging. @unman what's your opinion on this one?

user/security-in-qubes/firewall.md
rapenne-s: Interesting difference https://serverfault.com/questions/985158/what-is-the-different-between-iifname-and-iif-in-nftables

(TLDR: iifname allows to add rules for interfaces that doesn't exist yet)

rapenne-s:
A simple solution until we have a proper documentation with versioning could be to duplicate this page and name it "Firewall for 4.2" and rename the old page "Firewall for 4.1 and 4.0" ?

rapenne-s:
@andrewdavidwong I see I've been assigned "waiting for author", but I can't do anything at the moment about this. I guess we need to wait to have the new documentation platform with versionning support?

rapenne-s:
> > @andrewdavidwong I see I've been assigned "waiting for author", but I can't do anything at the moment about this. I guess we need to wait to have the new documentation platform with versionning support?
> 
> Can't you either do what @marmarek suggested in [#1344 (review)](https://github.com/QubesOS/qubes-doc/pull/1344#pullrequestreview-1714087993) or [what the doc guidelines currently say](https://www.qubes-os.org/doc/documentation-style-guide/#release-specific-documentation) for now until the new doc platform is up and running?

This seems a huge task to rework the whole text to separate Qubes OS 4.1 and 4.2 in different areas, I see 3 easy solutions until the new doc is up, this is getting urgent as people will upgrade to 4.2 and will be lost:

- duplicate the old page and link it as "Firewall on R4.1" on the doc index
- duplicate the old page and link it as "Firewall on R4.1" at the top of the new Firewall page covering 4.2
- concatenate the old text at the end of the new text with some separator explaining it's for older 4.1 release

marmarek:
> duplicate the old page and link it as "Firewall on R4.1" on the doc index

Sounds good for me.

rapenne-s:
> > * duplicate the old page and link it as "Firewall on R4.1" on the doc index
> 
> This is what @marmarek suggested in [#1344 (comment)](https://github.com/QubesOS/qubes-doc/pull/1344#issuecomment-1793872386), isn't it?
> 
> > * concatenate the old text at the end of the new text with some separator explaining it's for older 4.1 release
> 
> And this is [what the doc guidelines currently say](https://www.qubes-os.org/doc/documentation-style-guide/#release-specific-documentation) to do, isn't it?

Oh ok I understand, in https://github.com/QubesOS/qubes-doc/pull/1344#issuecomment-1803316008 where I suggested to do so, I understood your reply and the linked documentation saying that it shouldn't be done this way.

I won't be much available until the new year, if someone wants to do the changes to make it happen faster, please do.

unman:
I've taken the liberty of pushing the new page.
I've adopted Marek's suggestion, rather than the canonical subdivison of one page - I tried this, but the wholesale changes made it unwieldy and confusing, so I went for the two page option, linked at top.

rapenne-s:
thank you very much @unman

PR #1345, opened 2023-11-09
R4.2 Clean Install: Change Qubes Sources to 4.2 in Imported Templates

deeplow:
Tell users to upgrade the software sources when doing a 4.2 clean install. Otherwise they may miss on critical integrations for 4.2 like VMExec (for the updater to work) and most importantly, after 4.1 is EOL, it'll silently stop resolving Qubes package upgrades even though the template is not yet EOL.

## User story
I have a (rarely used) minimal template which I had been migrating ever since 4.0 through clean installs which was still targeting Qubes 4.0 repos :grimacing:. And the updates seemed to be working so I never assumed something was wrong.

I only detected this issue when after migrating to 4.2 updates stopped failed to udpate this template since `/etc/qubes-rpc/qubes.VMExec` was missing.

These instructions are also missing in the [4.1 clean install](https://www.qubes-os.org/doc/upgrade/4.1/#clean-installation) instructions.

## Unsolved issue: Missing Keyring (Debian only)
One thing I didn't manage to solve yet is the keyrings on Debian. On Fedora this was easy since the 4.2 signing keyring was already available and explicit in the `.repo` file. But for Debian not only was it not explicit in the `.list` file (unlike a 4.2 template), but the keyring wasn't available as far as I can tell in `/usr/share/keyrings/qubes-archive-keyring-4.2.gpg`.

This gives the users a chicken and egg situation. If the user doesn't add the new 4.2 key, they'll get an error:
```
GPG error: https://deb.qubes-os.org/r4.2/vm bullseye InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 5BB71C441BCB1OFE
```

However, the user won't get that key from any keyring unless they explicitly import it, which would add more complexity to the instructions. `qubes-core-agent` is the package providing the keyring file in 4.2 templates. Maybe this keyring could be backported to 4.1 template updates so the user doesn't run into this issue? :thinking:

## Ideally this would be solved by software

In an ideal situation the user would not need to do anything special. I see two ways of going about this:

### Option A

If the 4.1 -> 4.2 upgrade tool already does this, then tell the user doing a clean install to run this specific stage of the upgrade tool (in case that's possible)

### Option B

Integrate this template upgrade thing into the updater:

1. the updater would check if a template has 4.1 (or older) qubes repo sources
2. update the repo configuration accordingly

**Note**: to make this work the upgrade tool would probably need to make use of the old `qubes.VMShell` since in older templates `qubes.VMExec` isn't available.

### DIFF
--- a/user/downloading-installing-upgrading/upgrade/4_2.md
+++ b/user/downloading-installing-upgrading/upgrade/4_2.md
@@ -22,37 +22,26 @@ data.
 If you would prefer to perform a clean installation rather than upgrading
 in-place:
 
-1. Create a
+1. (optional) Run the updater to ensure all of your templates are in their latest version.
+2. Install the `qubes-dist-upgrade` tool. This is the inplace upgrade tool, which is not what we're doing. However it will needed in order to prepare the templates for the 4.2 version. You install it with the following command in the dom0 terminal:
+
+       sudo qubes-dom0-update -y qubes-dist-upgrade
+
+3. Change your templates to use the 4.2 repositories instead of the 4.1 ones. You do this with the following command in the dom0 terminal:
+
+       qubes-dist-upgrade --template-standalone-upgrade
+
+    **Note**: This step is critical to ensure the templates will receive updates once Qubes 4.1 reaches end-of-life (EOL) and was missing in previous clean installation instructions.
+
+4. Create a
    [backup](/doc/how-to-back-up-restore-and-migrate/#creating-a-backup) of your
    current installation.
-2. [Download](/downloads/) the latest 4.2 release.
-3. Follow the [installation guide](/doc/installation-guide/) to install Qubes
+5. [Download](/downloads/) the latest 4.2 release.
+6. Follow the [installation guide](/doc/installation-guide/) to install Qubes
    4.1.
-4. [Restore from your
+7. [Restore from your
    backup](/doc/how-to-back-up-restore-and-migrate/#restoring-from-a-backup) on
    your new 4.2 installation.
-5. In each old the old templates update the Qubes software sources from 4.1 to Qubes 4.2.
-   > **Note**: If the templates were first installed on a version older than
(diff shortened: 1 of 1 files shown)


deeplow:
Thanks for the quick feedback. I'll open an issue now.

marmarek:
> `qubes-core-agent` is the package providing the keyring file in 4.2 templates. Maybe this keyring could be backported to 4.1 template updates

This should be the case already, package for 4.1 (but not 4.0) includes keys for 4.2.

As for the overall issue, in 4.2 templates started advertised for which qubes version they are (more or less based on the version of qubes-core-agent package). This information isn't used anywhere yet - I guess we should add some warning based on that.

deeplow:
> This should be the case already, package for 4.1 (but not 4.0) includes keys for 4.2.

Good to know. Then in my case that was not working because this template came from 4.0. I have now added a note that instructions don't apply to 4.0 or older templates.

deeplow:
> I don't see an open issue for this. Do you want to open one, or do you want me to?

@andrewdavidwong done: https://github.com/QubesOS/qubes-issues/issues/8701

unman:
@deeplow  ping

marmarek:
The R4.2 keyrings are already included in R4.1 packages. You just need to have all R4.1 updates installed.

As for `qubes-dist-upgrade`, that's an interesting idea. We'd need to add some safeguards to not allow upgrading dom0 when already upgraded (in theory it shouldn't break anything, but may have some side effects, like leaving extra backup files or such).

marmarek:
> Actually, now that I think about it. Why not just tell users before the migration to run the `qubes-dist-upgrade --template-standalone-upgrade` while they're in 4.1? I think that may be even better than the changes in this PR.

If you are doing in-place upgrade from R4.1, this step is already part of the process. The issue is when you restore a backup from R4.1 (possibly an earlier one, from before you planned an upgrade).

PR #1347, opened 2023-11-11
Change links to HTTPS

marmarek:
Change links that has HTTPS equivalent to HTTPS.

### DIFF
--- a/introduction/support.md
+++ b/introduction/support.md
@@ -176,7 +176,7 @@ information in your message. A great way to provide your hardware details is by
 [generating and submitting a Hardware Compatibility List (HCL)
 report](/doc/how-to-use-the-hcl/#generating-and-submitting-new-reports), then
 linking to it in your message. [Ask questions the smart
-way.](http://www.catb.org/esr/faqs/smart-questions.html)
+way.](https://www.catb.org/esr/faqs/smart-questions.html)
 
 ### Be patient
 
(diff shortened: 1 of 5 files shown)


introduction/support.md
```diff
@@ -176,7 +176,7 @@ information in your message. A great way to provide your hardware details is by
 [generating and submitting a Hardware Compatibility List (HCL)
 report](/doc/how-to-use-the-hcl/#generating-and-submitting-new-reports), then
 linking to it in your message. [Ask questions the smart
-way.](http://www.catb.org/esr/faqs/smart-questions.html)
+way.](https://www.catb.org/esr/faqs/smart-questions.html)
```
zpc0, line 179: It seems `www.catb.org` doesn't support HTTPS.

introduction/support.md
```diff
@@ -309,7 +309,7 @@ account is in no way required, expected, or encouraged. Many discussants
 [mailing lists](https://en.wikipedia.org/wiki/Electronic_mailing_list),
 interacting with them solely through plain text email with
 [MUAs](https://en.wikipedia.org/wiki/Email_client) like
-[Thunderbird](https://www.thunderbird.net/) and [Mutt](http://www.mutt.org/).
+[Thunderbird](https://www.thunderbird.net/) and [Mutt](https://www.mutt.org/).
```
zpc0, line 312: It seems `www.mutt.org` doesn't support HTTPS. :cry:

introduction/support.md
gonzalo-bulnes: As of December 2023, the site does support HTTPS, but the certificate is invalid.

It is issued for `animadance.org` which well, is what it is... That being said, I don't think linking to the HTTPS version of the site and triggering the warning is making anything worse. It is a healthy warning, and the HTTP version probably warrants a similar warning anyway. I'd personally proceed with the change, YMMV.

gonzalo-bulnes (approved):
@andrewdavidwong @zpc0 I left some thoughts next to @zpc0's comments. I'll approve the PR because I would personally proceed with the change, but I let you(s) make the decision.

- The links are valid (no typos)
- The failing check was unrelated and is already fixed in `main`

gonzalo-bulnes:
I can't update the branch in @marmarek's fork @andrewdavidwong, so that's as far as I could go. 🙂

PR #1352, opened 2023-12-07
Use the 4.1 policy file location and format.

skyvine:
The current document still refers to /etc/qubes-rpc/policy and uses the old 3-column format. This commit updates it to use /etc/qubes/policy.d and and 5+-column format.

Additionally, correct a technical inaccuracy in the final example (the policy which the reader is instructed to create denies requests from source_vm2 with the +testfile1 argument, but the prose states that it will be accepted).

I wrote this before seeing the note about separate subsections for different versions in the style guide. However, as the QubesOS downloads page indicates that 4.0.4 is no longer supported, I think it makes sense to replace the information on this page.

I tested all of the examples on my machine, with the caveat that I am running the 4.2rc, not 4.1.

I signed this commit with my key. Note that I allow my keys to expire and rotate them on a regular basis (including my primary key), typically annually. If having an expired key on a commit will cause problems for your workflow in the future let me know and we can figure out a solution.

### DIFF
--- a/developer/services/qrexec.md
+++ b/developer/services/qrexec.md
@@ -86,25 +86,23 @@ Additionally, disposable VMs are tightly integrated -- RPC to a DisposableVM is 
 
 ### Policy files
 
-The dom0 directory `/etc/qubes-rpc/policy/` contains a file for each available RPC action that a VM might call.
+The dom0 directory `/etc/qubes/policy.d/` contains a file for each available RPC action that a VM might call.
 Together the contents of these files make up the RPC access policy database.
 Policies are defined in lines with the following format:
 
 ```
-srcvm destvm (allow|deny|ask[,default_target=default_target_VM])[,user=user_to_run_as][,target=VM_to_redirect_to]
+service-name|* +argument|* source destination action  [options]
 ```
 
-You can specify srcvm and destvm by name or by one of the reserved keywords such as `@anyvm`, `@dispvm`, or `dom0`.
-(Of these three, only `@anyvm` keyword makes sense in the srcvm field.
+You can specify the source and destination by name or by one of the reserved keywords such as `*`, `@dispvm`, or `dom0`.
+(Of these three, only `*` keyword makes sense in the source field.
 Service calls from dom0 are currently always allowed, and `@dispvm` means "new VM created for this particular request," so it is never a source of request.)
 Other methods using *tags* and *types* are also available (and discussed below).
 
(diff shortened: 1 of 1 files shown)


skyvine:
Thanks for checking. It looks like my issue had to do with trying to access the build log through a TOR connection, which seems like a reasonable thing to check in hindsight.

gonzalo-bulnes:
@skyvine The build was fixed. Updating this branch should get it the green tick / successful build report.

(Assuming this remote is called `origin`: `git checkout main && git pull origin main && git checkout update-qrexec-policy-for-4.1-refactoring && git rebase -S main` and a gentle push with `--force-with-lease` should do it.)

developer/services/qrexec.md
```diff
@@ -86,25 +86,23 @@ Additionally, disposable VMs are tightly integrated -- RPC to a DisposableVM is
 
 ### Policy files
 
-The dom0 directory `/etc/qubes-rpc/policy/` contains a file for each available RPC action that a VM might call.
+The dom0 directory `/etc/qubes/policy.d/` contains a file for each available RPC action that a VM might call.
```
gonzalo-bulnes, line 89: **Nitpicking**: The files in `/etc/qubes/policy.d/` are no longer split by RPC action. Rather they are organised by something more akin to _topics_ and _priority_. 

Since you're updating this sentence it might be a good opportunity to convey that as well? To me, saying "contains a file for each policy" is probably enough, since you're already updating the examples further down.

#### TL;DR:

```suggestion
The dom0 directory `/etc/qubes/policy.d/` contains a file for each available policy.
```

#### Reference docs:

> The biggest difference between the old and new policy formats is that, under the new format, the entire policy is a single entity. It is not divided into separate, per-service fragments. ([source](https://www.qubes-os.org/news/2020/06/22/new-qrexec-policy-system/#policy-files))

#### Example:

(For context, I know you've seen this @skyvine.)

```
# dom0

ls /etc/qubes-rpc/
# admin.backup.Cancel
# admin.backup.Execute
# admin.backup.Info
# admin.deviceclass.List
# ... (long list)
```

```
# dom0

ls /etc/qubes/policy.d

# 30-user.policy
# 35-compat.policy
# 80-whonix.policy
# ... (way shorter list)
```

gonzalo-bulnes (approved):
I left a comment regarding the first sentence, but I wouldn't consider that a blocker.

- The policy files new syntax is correct.
- The new copy is clear.
- The correction in the example at the end was indeed warranted :+1: 
- The build failure is unrelated (and has already been fixed in `main`)

PR #1355, opened 2023-12-19
update old reference to 4.1

ra--:


gonzalo-bulnes (approved):
This seems like a typo indeed. :+1:

PR #1365, opened 2024-01-03
remove outdated write-up against sudo passwords

adrelanos:
Maybe not perfect but better than spreading highly outdated, since revised viewpoints.

Fixes https://github.com/QubesOS/qubes-issues/issues/8823

### DIFF
--- a/user/security-in-qubes/vm-sudo.md
+++ b/user/security-in-qubes/vm-sudo.md
@@ -12,53 +12,7 @@ title: Passwordless root access in qubes
 
 Background (`/etc/sudoers.d/qubes` in VM):
 
-```
-user ALL=(ALL) NOPASSWD: ALL
-
-# WTF?! Have you lost your mind?!
-#
-# In Qubes VMs there is no point in isolating the root account from
-# the user account. This is because all the user data is already
-# accessible from the user account, so there is no direct benefit for
-# the attacker if she could escalate to root (there is even no benefit
-# in trying to install some persistent rootkits, as the VM's root
-# filesystem modifications are lost upon each start of a VM).
-#
-# One might argue that some hypothetical attacks against the
-# hypervisor or the few daemons/backends in Dom0 (so VM escape
-# attacks) most likely would require root access in the VM to trigger
-# the attack.
-#
-# That's true, but mere existence of such a bug in the hypervisor or
-# Dom0 that could be exploited by a malicious VM, no matter whether
-# requiring user, root, or even kernel access in the VM, would be
-# FATAL. In such situation (if there was such a bug in Xen) there
-# really is no comforting that: "oh, but the mitigating factor was
-# that the attacker needed root in VM!" We're not M$, and we're not
-# gonna BS our users that there are mitigating factors in that case,
-# and for sure, root/user isolation is not a mitigating factor.
-#
-# Because, really, if somebody could find and exploit a bug in the Xen
-# hypervisor -- as of 2016, there have been only three publicly disclosed
-# exploitable bugs in the Xen hypervisor from a VM -- then it would be
-# highly unlikely if that person couldn't also found a user-to-root
-# escalation in VM (which as we know from history of UNIX/Linux
-# happens all the time).
-#
-# At the same time allowing for easy user-to-root escalation in a VM
-# is simply convenient for users, especially for update installation.
-#
-# Currently this still doesn't work as expected, because some idotic
-# piece of software called PolKit uses own set of policies. We're
-# planning to address this in Beta 2. (Why PolKit is an idiocy? Do a
-# simple experiment: start 'xinput test' in one xterm, running as
-# user, then open some app that uses PolKit and asks for root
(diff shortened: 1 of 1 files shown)


marmarek:
I don't think it's appropriate to remove this description, maybe just rephrase to not suggest it's directly from some file in a VM that isn't there anymore. After all, password-less root access in VM is still the default and it should be documented why it's the case.

adrelanos:
But the new state is more correct than the current state which is a huge reasoning pro passwordless sudo which has now been revised?

I believe that this might have [created](https://forums.whonix.org/t/benefiting-from-security-misc-inside-qubes-whonix/18124) the impression that Qubes isn't interested in sudo hardening, which I think is no longer true.

So rather than being completely anti-sudo password I think it's better to just link to the ticket so at least highly technical users can find out.

> is still the default and it should be documented why it's the case.

It's the case because nobody implemented https://github.com/QubesOS/qubes-issues/issues/2695 yet?

marmarek:
> So rather than being completely anti-sudo password I think it's better to just link to the ticket so at least highly technical users can find out.

Well, all those points about isolation (or rather lack of it) a user _with shell/arbitrary exec access_ from root (if the user can actually get root access, via some extra authorization) are still very much true. If you can use `sudo` with some extra steps, then malicious code on your user account can simply wait until you perform those extra steps, whatever they are.

So, most of this description is still accurate. The thing that changed is we made the password-less root access opt-out, if you have a case where you more or less never use sudo at all. In fact, uninstalling passwordless-root package prevents using sudo for arbitrary commands at all (unless you configure some alternative authorization).

> > is still the default and it should be documented why it's the case.
> 
> It's the case because nobody implemented [QubesOS/qubes-issues#2695](https://github.com/QubesOS/qubes-issues/issues/2695) yet?

There is `qvm-console-dispvm` that can be used to access root account even if sudo is disabled. So, some alternative method exists already, if one really wants to go this route.

But again, the current state is the passwordless-root enabled by default, and it's wrong to remove documentation explaining why it is the case. It might be appropriate if that default changes at some point.

adrelanos:
> If you can use `sudo` with some extra steps, then malicious code on your user account can simply wait until you perform those extra steps, whatever they are.

I would suggest to have VMs that by user volunteer policy (or one day even a Qubes dom0 VM setting) can never get sudo. Useful for VMs only used for browsers.

> So, most of this description is still accurate. The thing that changed is we made the password-less root access opt-out, if you have a case where you more or less never use sudo at all. In fact, uninstalling passwordless-root package prevents using sudo for arbitrary commands at all (unless you configure some alternative authorization).

Good.

> > > is still the default and it should be documented why it's the case.
> > 
> > 
> > It's the case because nobody implemented [QubesOS/qubes-issues#2695](https://github.com/QubesOS/qubes-issues/issues/2695) yet?
> 
> There is `qvm-console-dispvm` that can be used to access root account even if sudo is disabled. So, some alternative method exists already, if one really wants to go this route.

Nice. Also needs documentation.

> But again, the current state is the passwordless-root enabled by default, and it's wrong to remove documentation explaining why it is the case. It might be appropriate if that default changes at some point.

Ok.

SaswatPadhi:
> But again, the current state is the passwordless-root enabled by default

It doesn't seem to be the default for any of the  minimal VMs provided by Qubes.

runephilosof-abtion:
@andrewdavidwong You added the "waiting for author" label, so I guess you want @adrelanos to respond before reviewing again?

If the outdated reasoning for adding passwordless in the first place must stay, I would prefer to prepend it with:
> This is the original reasoning for implementing passwordless sudo, see issue https://github.com/QubesOS/qubes-issues/issues/2695 for revising the policy.

But I would prefer to just have it deleted as in this PR.

> the current state is the passwordless-root enabled by default, and it's wrong to remove documentation explaining why it is the case

The proposal links to a place where that explanation can be found. So because of that I see nothing wrong with this PR. It is a better state than the current.

unman:
There's confusion here.
For most Qubes users, in default set-up, passwordless root is enabled.
As far as I am aware, the thinking behind this has NOT changed, and no one has given any reason to think that it has.
Given that the explanation in the qubes has been (unfortunately) removed in 4.2 the PR should be revised accordingly.

The fact that passwordless root is not enabled in minimal templates is well documented.
The possibility of using dom0 authorisation is linked from this page.

adrelanos:
meta: https://github.com/QubesOS/qubes-issues/issues/3629#issuecomment-1935589503

unman:
This PR is based on a mistaken view - that the original page contains "highly outdated, since revised viewpoints"

Passwordless root remains the default, and as @marmarek has said, most of the description is still accurate.
I am moving details of the implementation to developer docs (as per discussion in PR #1296), and updating this page.

tlaurion:
@andrewdavidwong ELI5?

unman:
*None* of the documentation is ELI5
I thought the "rant" was already pitched quite low.
Can you suggest what you like to see?

PR #1374, opened 2024-02-06
fix hex digit number in the key upload instruction

slayoo:


unman (commented):
Since you are selecting a key from your own keyring, I cant see that use of the short ID would be incorrect.        
Even if there were collision all you would do would be to push a number of keys, **including your own**, to the keyserver.     
No?

slayoo:
@unman thanks for feedback
My point was just to make the narrative (8 digits) consistent with the example shell command snippet (16 digits)

unman:
PipelineRetry

PR #1375, opened 2024-02-06
remove repeated code (shell session) snippet

slayoo:


unman (commented):
Since you are selecting a key from your own keyring, I cant see that use of the short ID would be incorrect.        
Even if there were collision all you would do would be to push a number of keys, **including your own**, to the keyserver.
What's wrong here?

slayoo:
@unman thanks for checking it
this PR is not about the content, just about the code snippet being given twice:
```diff
'''shell_session
$ gpg --send-keys --keyserver hkps://keyserver.ubuntu.com 6E2F4E7AF50A5827
gpg: sending key 6E2F4E7AF50A5827 to hkps://keyserver.ubuntu.com
'''

- '''
- $ gpg --send-keys --keyserver hkps://keyserver.ubuntu.com 6E2F4E7AF50A5827
- gpg: sending key 6E2F4E7AF50A5827 to hkps://keyserver.ubuntu.com
- '''
```

unman:
PipelineRetry

unman:
PipelineRetry

PR #1372, opened 2024-02-02
Generalize YubiKey guide by adding TOTP MFA

deeplow:
Not every user will have a hardware token. This adds instructions for setting up multi-factor with TOTP. This was inspired / based on [work](https://forum.qubes-os.org/t/otp-for-xscreensaver-guide/23988) by @kennethrrosen.

Everything about the YubiKey guide is kept but moved to a lower heading level to acommodate for the two MFA options: YubiKey or TOTP.

**NOTE** I am no expert on MFA on Linux, so (as expected) I'd recommend especially careful review.

References that helped me build this guide:
 - https://forum.qubes-os.org/t/otp-for-xscreensaver-guide/23988
 - https://michaelpesa.com/posts/creating-custom-authselect-profiles/
 - https://www.redhat.com/sysadmin/mfa-linux
 - [Authselect man page](https://www.mankier.com/8/authselect)

PR #1380, opened 2024-03-02
Include NitroKey 3 in MFA doc

UndeadDevel:
This PR adds documentation for the recently [added](https://github.com/QubesOS/qubes-app-yubikey/pull/22) NK3 support to the `qubes-app-yubikey` package. I've also integrated the changes of #1372 (except the Google Authenticator part), because that PR changes the name and structure of the file and it makes sense to adapt my changes to it. Do note that the referenced PR's inserted block from line 28 to 123 should thus be merged also from #1372 to include Google Authenticator instructions.

Regarding NitroKey and `pipx`: I've waited now for almost two months and their binaries are still not signed, so I've simply linked to their installation instructions and put a prominent warning against `pipx` in this doc PR. Their main developer did confirm in their forum that detached signatures will be added in one of the next releases, however, and they also have the intention to package `pynitrokey` eventually, which I also reference in this doc.


unman:
@UndeadDevel  Can you  review this after the merging of #1372?

UndeadDevel:
Conflicts are resolved; ready to merge.

NitroKey still hasn't started providing signatures for their artifacts smh...it's been >4 months and several pynitrokey versions now since they promised that, but, alas, the doc has a big fat warning and should be robust to change as long as they update their docs, which mine link to, once they finally provide signed artifacts or even proper packaging.

Checks still failing for the same reason as before (see above)...someone should probably fix that.

PR #1381, opened 2024-03-03
Mention existence of Qubes Builder V2

deeplow:
Qubes Builder V1 is still supported at least until Qubes 4.1 reaches EOL. However the docs don't mention at all the existance of version 2 of the builder. The ideal solution would be to document builder v2, while that doesn't happen, at least having a warning is better than nothing.


unman:
I have just now confirmed with Marmarek the status of v1 and v2, so
this is timely.
Thanks.

PR #1389, opened 2024-03-26
Update gsod.md

mfc:
2023->2024


mfc:
as we need to update proposed project language as well since we did the instructional video series in 2023 GSoD

PR #1396, opened 2024-04-26
Fix a few typos

ldv-alt:


ldv-alt:
The test failure seems to be unrelated.

PR #1398, opened 2024-04-28
Fix action/options policy separator in split-gpg doc

lubellier:
The split-gpg doc gives the below example:
```
@anyvm  @anyvm  ask,default_target=work-gpg
```
which failed because the policy format is 
```
service-name|* +argument|* source destination action  [options]
```
based on the [qrexec reference doc](https://www.qubes-os.org/doc/qrexec/#policy-files). So this PR replaces the `,` separator by a space.


lubellier:
@unman You didn't merge this PR, is something wrong with my PR? I think I followed the steps from [how-to-edit-the-documentation](https://www.qubes-os.org/doc/how-to-edit-the-documentation/).

PR #1412, opened 2024-07-12
 Firewall - use iifgroup instead of iif ==

apparatius:
The disappearance of the eth0 interface when you restart the net qube of the sys-firewall or set it’s net qube to none is causing `iif == "eth0"` to become `iif 2` and the rules won't work anymore. It’s better to use `iifgroup 1` instead of `iif == "eth0"`. Related discussion:
https://forum.qubes-os.org/t/iptables-not-available-in-sys-net-in-qubes-os-4-2-1/26706/26

### DIFF
--- a/user/security-in-qubes/firewall.md
+++ b/user/security-in-qubes/firewall.md
@@ -297,13 +297,13 @@ nft add chain qubes custom-dnat-qubeDEST '{ type nat hook prerouting priority fi
 Second step, code a natting firewall rule to route traffic on the outside interface for the service to the sys-firewall VM
 
 ```
-nft add rule qubes custom-dnat-qubeDEST iif == "ens6" ip saddr 192.168.x.y/24 tcp dport 443 ct state new,established,related counter dnat 10.137.1.z
+nft add rule qubes custom-dnat-qubeDEST iifgroup 1 ip saddr 192.168.x.y/24 tcp dport 443 ct state new,established,related counter dnat 10.137.1.z
 ```
 
 Third step, code the appropriate new filtering firewall rule to allow new connections for the service
 
 ```
-nft add rule qubes custom-forward iif == "ens6" ip saddr 192.168.x.y/24 ip daddr 10.137.1.z tcp dport 443 ct state new,established,related counter accept
+nft add rule qubes custom-forward iifgroup 1 ip saddr 192.168.x.y/24 ip daddr 10.137.1.z tcp dport 443 ct state new,established,related counter accept
 ```
 
 > Note: If you do not wish to limit the IP addresses connecting to the service, remove `ip saddr 192.168.x.y/24` from the rules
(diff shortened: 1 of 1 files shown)


rapenne-s:
This is actually a good idea. eth0 interface is by default in group 1 on Qubes OS, while vif interfaces are in group2, so this will work fine

rapenne-s:
This is not so clear on a fedora sys-net, physical interfaces are in the group "default", while vif interface is correctly in group 2.

I'm not convinced this PR can work as this, as I understand it modifies the rules for sys-net, while you may want to use `iifgroup 1` in sys-firewall. The sys-net part should be dropped, you can't replace `iif "ens6"` by `iifgroup 1` because (as I can see now on a fedora 41 sys-net), physical interfaces and loopback are in group `default` (which seems to be 0)

user/security-in-qubes/firewall.md
```diff
@@ -351,10 +351,10 @@ Content of `/rw/config/qubes-firewall-user-script` in `sys-net`:
 if nft add chain qubes custom-dnat-qubeDEST '{ type nat hook prerouting priority filter +1 ; policy accept; }'
 then
   # create the dnat rule
-  nft add rule qubes custom-dnat-qubeDEST iif == "ens6" saddr 192.168.x.y/24 tcp dport 443 ct state new,established,related counter dnat 10.137.1.z
```
rapenne-s, line 354: this can't work, physical interfaces are in group default

user/security-in-qubes/firewall.md
```diff
@@ -351,10 +351,10 @@ Content of `/rw/config/qubes-firewall-user-script` in `sys-net`:
...
+  nft add rule qubes custom-dnat-qubeDEST iifgroup 1 saddr 192.168.x.y/24 tcp dport 443 ct state new,established,related counter dnat 10.137.1.z
 
   # allow forwarded traffic
-  nft add rule qubes custom-forward iif == "ens6" ip saddr 192.168.x.y/24 ip daddr 10.137.1.z tcp dport 443 ct state new,established,related counter accept
```
rapenne-s, line 357: this can't work, physical interfaces are in group default

rapenne-s:
Why did this PR get merged? I don't think it works, as I stated in my previous comment and inline comments of requested changes

unman:
Look at the history.
Usually I fix and commit.
Here I committed and fixed.
Do you think the current page is not working?

rapenne-s:
> Look at the history. Usually I fix and commit. Here I committed and fixed. Do you think the current page is not working?

Yes, I studied nftables syntax and verified what the use of iifgroup would do in a fedora qube, in my opinion it does not work as expected / explained in this PR.

unman:
Have you looked at the page?
iifgroup is consistent in a fedora sys-firewall across netvm changes,
no?

rapenne-s:
> Have you looked at the page? iifgroup is consistent in a fedora sys-firewall across netvm changes, no?

from https://wiki.nftables.org/wiki-nftables/index.php/Matching_packet_metainformation , iifgroup matches a dev group. By default, physical interfaces on qubes os are in group "default", which stands for dev group 0

```
sys-net $ sudo ip l
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
2: ens6: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP mode DEFAULT group default qlen 1000
    link/ether XX brd ff:ff:ff:ff:ff:ff permaddr XX
    altname enp0s6
3: wls7f0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state DOWN mode DORMANT group default qlen 1000
    link/ether XX brd ff:ff:ff:ff:ff:ff permaddr XXX
    altname wlp0s7f0
4: vif5.0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP mode DEFAULT group 2 qlen 1000
    link/ether fe:ff:ff:ff:ff:ff brd ff:ff:ff:ff:ff:ff
```

marmarek:
@rapenne-s unman is trying to tell you to look at the current version of the page, not just this PR. There are further changes applied after this PR. Especially, I see some `iifgroup` are changed (back?) to `iifname`.

PR #1414, opened 2024-07-21
Improve system update documentation

DemiMarie:
This is a very important task, but it is often not done due to bad tools.

Marking as draft because support for System76 firmware update has not been merged yet.


rapenne-s:
PipelineRetry

PR #1417, opened 2024-08-10
Fix a dom0 package downgrade/reinstall commands

apparatius:
Related forum topic:
https://forum.qubes-os.org/t/documentation-error-in-how-to-install-software-in-dom0/28190


rapenne-s:
PipelineRetry

PR #1422, opened 2024-08-30
Fix headings

RandyTheOtter:
Fix headings before working on the issue 9306


rapenne-s:
Thank you for your contribution :)

Sorry for the delay, it should get better since Qubes OS now has a documentation maintainer team now :tada:

PR #1433, opened 2024-10-19
Update docs about /rw/config/rc.local* scripts

marmarek:
Specify that files needs to be executable.
Add info about `/rw/config/rc.local.d/*.rc`.
Add info about `/rw/config/rc.local-early.d/*.rc` and
`/rw/config/rc.local-early`
(https://github.com/QubesOS/qubes-core-agent-linux/pull/386)

### DIFF
--- a/user/advanced-topics/config-files.md
+++ b/user/advanced-topics/config-files.md
@@ -19,7 +19,7 @@ That way, they can be used to customize a single VM instead of all VMs based on 
 The scripts here all run as root.
 
 - `/rw/config/rc.local` - script runs at VM startup.
-    Good place to change some service settings, replace config files with its copy stored in `/rw/config`, etc.
+    Good place to change some service settings, replace config files with its copy stored in `/rw/config`, etc. The script need to have the executable permission set to be executed.
     Example usage:
 
     ~~~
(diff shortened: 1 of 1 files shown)


user/advanced-topics/config-files.md
```diff
@@ -19,7 +19,7 @@ That way, they can be used to customize a single VM instead of all VMs based on
 The scripts here all run as root.
 
 - `/rw/config/rc.local` - script runs at VM startup.
-    Good place to change some service settings, replace config files with its copy stored in `/rw/config`, etc.
+    Good place to change some service settings, replace config files with its copy stored in `/rw/config`, etc. The script need to have the executable permission set to be executed.
```
MarnixCroes, line 22:
```suggestion
    Good place to change some service settings, replace config files with its copy stored in `/rw/config`, etc. The script needs to have the executable permission set to be executed.
```

maiska: I would suggest replacing _needs to_ wiht _must_. It reads more easily. F.ex.: _The script must have executable permissions set in order to be executed._ Or _The script needs to have executable permissions in order to run._ These are merely suggestions tho.

maiska (changes requested):
It suffices if _needs_ is incorporated

unman:
Does rc.local not ship with executable permission enabled? If not, this
is a regression.

PR #1435, opened 2024-10-30
Restore backup: add info about refresh applications

MarnixCroes:
This is nowhere mentioned.

Someone has find the answer on a forum, so this is a good addition/clarification imo.

### DIFF
--- a/user/how-to-guides/how-to-back-up-restore-and-migrate.md
+++ b/user/how-to-guides/how-to-back-up-restore-and-migrate.md
@@ -160,6 +160,9 @@ out.
 6. When you are ready, click **Next**. Qubes will proceed to restore from your
 backup. Once the progress bar has completed, you may click **Finish**.
 
+In case that applications are not shown, i.e. "No applications found", open the
+settings of the qube -> select `Applications` -> click `Refresh applications`.
+
 **Note:** When restoring from a dom0 backup, a new directory will be created in
 the current dom0 home directory, and the contents from the backup will be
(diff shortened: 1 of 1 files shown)


rapenne-s:
PipelineRetry

user/how-to-guides/how-to-back-up-restore-and-migrate.md
```diff
@@ -160,6 +160,9 @@ out.
 6. When you are ready, click **Next**. Qubes will proceed to restore from your
 backup. Once the progress bar has completed, you may click **Finish**.
 
+In case that applications are not shown, i.e. "No applications found", open the
+settings of the qube -> select `Applications` -> click `Refresh applications`.
+
 **Note:** When restoring from a dom0 backup, a new directory will be created in
```
maiska, line 166: Hey @MarnixCroes thank you for your suggested contribution!
I would suggest elaborating it further by pointing out that in the case of a restored application qube refreshing the application lists will open the template qube on which it is based (template qube should also be restored in that case, if missing the default qube will be assigned). 
The updated list of the installed software can be seen on the left and adjusted accordingly to your needs.
What do you think?

MarnixCroes: thank you for your review,
I've applied your suggestion

maiska: merged and corrected  user's needs :)

PR #1442, opened 2024-12-09
Syntax changed for enabling repositories due to dnf5/Fedora41

ghijg:
Syntax changed for enabling repositories with dnf5 which is used by default from Fedora 41 on.

https://dnf5.readthedocs.io/en/latest/changes_from_dnf4.7.html


rapenne-s:
Thanks for the contribution, I think for now we should have separate instructions for Fedora 40 (not EoL) and Fedora 41, then once Fedora 42 is out, we could only keep the new instructions, what do you think?

Fedora 41 has its share of issues, I think a lot of people stayed on Fedora 40

PR #1432, opened 2024-10-10
system-requirements: fix link to section

MarnixCroes:
the section is `Important Notes`, not `Updates`

fixed the link

PR #1443, opened 2024-12-20
Secondary storage: work around 4k incompatibility, drop unnecessary `luksFormat` arguments

rustybird:
Feel free not to take the second or third commit, it just looked a bit crowded to me.


rustybird:
The unrelated HTML-Proofer failures are already fixed in #1432

PR #1448, opened 2025-01-28
Update installation-guide.md : disconnect all USB devices prior of stage2 install

tlaurion:
Otherwise detected USB HID devices (keyboard) will be permitted to interact with dom0 as security baseline

Closes https://github.com/linuxboot/heads-wiki/issues/171


marmarek:
This wants two important clarifications:
1. If your only/primary keyboard/mouse is on USB, of course don't disconnect it.
2. The presence of USB keyboard affects _only_ presented default configuration. Alternatively to disconnecting all of that, you can simply uncheck "allow usb keyboard" checkbox (if you don't want to use USB keyboard). This one: https://openqa.qubes-os.org/tests/126419#step/firstboot/9. Below the option you can see list of devices that will be affected.

maiska:
Hello @tlaurion, would you be open to incorporating the clarification suggested by @marmarek ? I think it would enhance the documentation. Thanks for your work!

PR #1449, opened 2025-02-11
dnf5 command syntax update

MattiasHansson:
Fedora 41 uses dnf5 which has updated syntax on new dnf

dnf config-manager --set-enabled
changed to 
dnf config-manager setopt

First time doing this. So hope it's correct.


rapenne-s:
hi, there is already a P5 in #1442

PR #1452, opened 2025-02-19
Remove mention that Microsoft provided ISOs are untrustworthy

TommyTran732:
This is no longer true, as they do give the hashes when downloading via their website now:

### DIFF
--- a/user/templates/windows/windows-qubes-4-1.md
+++ b/user/templates/windows/windows-qubes-4-1.md
@@ -53,9 +53,9 @@ However, if you are an expert or want to do it manually you may continue below.
 
 Have the Windows ISO image (preferrably the 64-bit version) downloaded in some qube.
 
-Windows ISOs can be downloaded directly from Microsoft (eg. [here](https://www.microsoft.com/en-us/software-download/windows10ISO) for Win10), or selected and downloaded via the [Windows Media Creation Tool](https://go.microsoft.com/fwlink/?LinkId=691209). You should, however, regard the downloaded image to be untrustworthy, since there is no reliable way to check that the download was not somehow compromised (see the discussion in issue [Simplify Qubes Windows Tools Installation for R4.1 #7240](https://github.com/QubesOS/qubes-issues/issues/7240)).
+Windows ISOs can be downloaded directly from Microsoft (eg. [here](https://www.microsoft.com/en-us/software-download/windows10ISO) for Win10), or selected and downloaded via the [Windows Media Creation Tool](https://go.microsoft.com/fwlink/?LinkId=691209).
 
-Unofficial “debloated” ISOs from projects like reviOS 18 or ameliorated 10 can be found on the net, although obviously you should consider them even “unsafer” than MS provided ISOs. Alternatively, one could download an official ISO and run scripts/apply patches before installation. Some of the “tweaks” might end up being too much depending on the qube’s planned usage though (eg. no appx functionality in ameliorated windows - so the installation of Windows Store apps is impossible, even with powershell).
+Unofficial “debloated” ISOs from projects like reviOS 18 or ameliorated 10 can be found on the net, although obviously you should consider them unsafe ISOs. Alternatively, one could download an official ISO and run scripts/apply patches before installation. Some of the “tweaks” might end up being too much depending on the qube’s planned usage though (eg. no appx functionality in ameliorated windows - so the installation of Windows Store apps is impossible, even with powershell).
 
 **Create Windows VM**
 


ben-grande:
There still isn't any reliable way to check that the download was not somehow compromised. Checksum does not provide authenticity. A compromised server can provide modified hashes with modified binaries and you cannot attest it was made by Microsoft or not.

TommyTran732:
> There still isn't any reliable way to check that the download was not somehow compromised. Checksum does not provide authenticity. A compromised server can provide modified hashes with modified binaries and you cannot attest it was made by Microsoft or not.

By that logic, how can you check the authenticity of Linux isos?

The webserver for Linux websites can also provide modified hashes then link you malicious mirrors with malicious ISOs

tlaurion:
> > There still isn't any reliable way to check that the download was not somehow compromised. Checksum does not provide authenticity. A compromised server can provide modified hashes with modified binaries and you cannot attest it was made by Microsoft or not.
> 
> By that logic, how can you check the authenticity of Linux isos?
> 
> The webserver for Linux websites can also provide modified hashes then link you malicious mirrors with malicious ISOs

Distros typically provide hash file with different checksums, and that hash file is also typically detached signed with a kept private gpg key that can be validated with public key counterpart. Some Distros still provide detached signature of ISO, which can validate both integrity and authenticity at the same time.

The key here is detached signed digest (hash) files, but the problem there is the complexity of verifying those by lambda users which typically won't verify the hash file detached signature, because getting the public key and doing the whole verification process is not convenient.

Tails, for example, provides a Javascript helper on their download page, so that end user can, once ISO downloaded, use the helper to verify the iso file downloaded directly from the web page. 

Otherwise, torrents also help doing the whole process, but with the torrent file containing hashes of blocks that are validated as they are downloaded from a swarm.

Mirrors of iso files are synced automatically, typically once a day if not more.

Tampering an iso and distributing it is possible, but hard. The idea here is to distrust the infrastructure, but to do so, there is reliance on end users doing the right thing and apply due process.

In my opinion, the easiest still today is torrent files. But opinions may vary. If torrent contains both digest (hash) file and detached signature file and torrent verifying integrity, you get everything done at once for you. Then dd that iso to thumb drive and unless thumb drive flash is dying, you should be good to go.

Otherwise, ideal is for bootloader to have distro public keys, but that woukd be a big contract. So therefore. The exercice is left to the end user of doing do diligence and verify both integrity and authenticity of iso, hash and signature.

TommyTran732:
> Distros typically provide hash file with different checksums, and that hash file is also typically detached signed with a kept private gpg key that can be validated with public key counterpart. 

Nothing really stopping a malicious web server from also saying "Oh hey btw this is our [malicious] hash file signed against our [malicious] gpg key".

For example: https://fedoramagazine.org/verify-fedora-iso-file/
The GPG public key is hosted at https://getfedora.org/static/fedora.gpg, the same webserver linking to the ISOs and mirrors. A compromise of getfedora.org will defeat all of the verification steps. The checksum verification can only protect against malicious third party mirrors, but not against a compromise of getfedora.org.

> Tails, for example, provides a Javascript helper on their download page, so that end user can, once ISO downloaded, use the helper to verify the iso file downloaded directly from the web page.

Same situation here. The checksum verification can only protect against third party mirrors, not tails.net. It's a distinction without a real difference.

> In my opinion, the easiest still today is torrent files.

This is yet another example of the same thing. If the user gets the torrent file/magnet link from distro.tld, then there is nothing stopping distro.tld from telling them to download a malicious torrent.

> Otherwise, ideal is for bootloader to have distro public keys, but that woukd be a big contract. 

This is not remotely ideal, because it's just the same situation signed shim with a different paint. 

---

The whole thing only makes sense if the gpg signature is distributed from one source, while the link ISOs are provided from a different source. And even then, if there is a mismatch, the user will be left confused as to whether the source for the gpg key is just outdated or if it is a real compromise.

Anyhow, the point I am trying to make is that this is not what's happening with relevant distros. And if we are not going out of our way to say "Fedora ISOs" are unsafe, then holding this against Microsoft is a double standard.

TommyTran732:
Oh, and I forgot to mention - by this standard, even the Qubes ISO is unsafe:

The GPG key, the signed hashes, and the link to the ISOs are all from qubes-os.org. If qubes-os.org is compromised, it can just change everything. So how much more trustworthy is it compared to the Microsoft ISO download?

marmarek:
> For example: https://fedoramagazine.org/verify-fedora-iso-file/
> The GPG public key is hosted at https://getfedora.org/static/fedora.gpg, the same webserver linking to the ISOs and mirrors. A compromise of getfedora.org will defeat all of the verification steps. The checksum verification can only protect against malicious third party mirrors, but not against a compromise of getfedora.org.

There are serveral ways to verify GPG key, you surely should _not_ trust it based on downloading it from the very same website, for the reason you explained. See second part of https://www.qubes-os.org/security/verifying-signatures/#how-to-import-and-authenticate-the-qubes-master-signing-key. Remember, the GPG key (or another signing key) is something that is usually valid for a long time, so it's feasible to have several places to verify it. You can't realistically do that with hashes, as you'd need to re-distribute them securely every time new release is posted.

unman:
I do not see that this will reach a conclusion, so I am minded to close this PR.
I'll be happy to accept an alternative that addresses the issue of checking authenticity.

PR #1455, opened 2025-03-11
Adding more headers and ideas to troubleshoot appmenus

parulin:
The "App menu shortcut troubleshooting" page seemed to be somewhat outdated and lacking in information to facilitate troubleshooting. This is what I did:

* trying to use consistently "qube" instead of "vm"
* trying to get consistent placeholders as uppercase between brackets
* the first part now has a title and distinguishes the GUI from the CLI, when needed, it is referred to in the following part, instead of repeating it.
* when a shortcut is missing, add a way to check it from the qube, to get a better understanding of the problem ("Unavailable desktop entry")
* the example of a desktop entry is simplified to its bare minimum
* "Fixing shortcuts" is replaced by a more accurate title. Not sure if this part is still useful.
* mention that `my-old-vm` becomes `my_dold_dvm` in some files names
* removed an off-topic comment on Windows HVM and a missing (old ?) script
* new part about how to troubleshoot a shortcut that doesn't do anything

Todo:

* update the screenshots
* is the "custom entry" still useful ?
* "Behind the scenes" could be developed

### DIFF
--- a/user/troubleshooting/app-menu-shortcut-troubleshooting.md
+++ b/user/troubleshooting/app-menu-shortcut-troubleshooting.md
@@ -14,22 +14,19 @@ title: App menu shortcut troubleshooting
 For ease of use Qubes aggregates shortcuts to applications that are installed in app qubes and shows them in one application menu (aka "app menu" or "start menu") in dom0.
 Clicking on such shortcut runs the assigned application in its app qube.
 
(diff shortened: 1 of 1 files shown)


rapenne-s:
PipelineRetry

rapenne-s:
thanks

parulin:
And thanks for reviewing it!

PR #1457, opened 2025-03-18
firewall: reword DNS resolution note

coyotebush:
This paragraph excessively discouraged me from using the firewall, because two elements led to misconceptions:
1. In the context of the GUI, "applying" sounds like it refers to the "Apply" button, suggesting DNS resolution happens only when using the GUI and only IP addresses are persisted.
2. "load balancing" sounds like it could also refer to DNS names that return a (stable) set of multiple IP addresses for clients to choose from, or IP addresses that route incoming traffic to multiple processes.

Meanwhile, the note about "complex web sites" is true but neither relevant to DNS resolution nor clearly helpful to users.


rapenne-s:
Thanks for your PR, the new wording is really good.

PR #1465, opened 2025-04-09
Add CLI instructions for configuring a Yubikey in Challenge/Response Mode

jmynes:
The GUI instructions for configuring a Yubikey in Challenge/Response mode require a now EOL tool, which is not found in the Debian Bookworm repository:
https://developers.yubico.com/yubikey-personalization-gui/

Yubikey recommends one of two replacement tools:
- GUI: [yubioath-flutter](https://github.com/Yubico/yubioath-flutter/)
  - Seemingly, this does not support setting Challenge/Response
- CLI: [yubikey-manager](https://github.com/Yubico/yubikey-manager/)
  - I did not test this tool, yet

I used their older CLI tool, [ykpersonalize](https://github.com/Yubico/yubikey-personalization), to achieve the same result as the existing documentation below; however, I now realize while writing this, that _this_ tool is also deprecated, despite still being available in Debian Bookworm.

Further testing is recommended, I'll update when I find out more

### DIFF
--- a/user/security-in-qubes/mfa.md
+++ b/user/security-in-qubes/mfa.md
@@ -200,16 +200,27 @@ website](https://docs.nitrokey.com/software/nitropy/all-platforms/installation).
 
 3. Configure your YubiKey / NitroKey3:
    
-   **YubiKey**
+   **YubiKey (via CLI)**
    
    Configure your YubiKey for challenge-response `HMAC-SHA1` mode. This can be
    done on any qube, e.g. a disposable (you need to [attach the
 YubiKey](https://www.qubes-os.org/doc/how-to-use-usb-devices/) to this app qube
 though) or directly on the sys-usb vm.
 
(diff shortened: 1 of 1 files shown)


mkowalski18:
If using Debian 12 (bookworm), the command `sudo apt-get install yubikey-manager` will install ykman(1), which is the supported way to configure YubiKeys under Linux now.

From mfa.md, I don't immediately recognize what is trying to be achieved there - it's been years since YubiKey Personalization GUI was a supported tool.

If I understood what was trying to be achieved I'd suggest the alternative command for ykman(1).  However you can find the documentation for it here: https://docs.yubico.com/software/yubikey/tools/ykman/Base_Commands.html

maiska:
Heads up: this will take a little while to review :)

unman:
It seems that this PR does not use current tools and the aim is unclear. I'm going to close this PR, and open a new issue to update the page with current tools.

PR #1466, opened 2025-04-25
Describe installation changes for Windows 11

GWeck:
Microsoft has put further restrictions on Windows 11 installation without TPM or Microsoft account


GWeck:
Spelling check: 24H2 is unknown, but that's o.k.

PR #1467, opened 2025-04-27
Fix broken API documentation link 

PixelPil0t1:
Removed the non-working API documentation link and replaced it with a working alternative that I found. If there is a better or more accurate link, please suggest it.

### DIFF
--- a/developer/releases/4_0/release-notes.md
+++ b/developer/releases/4_0/release-notes.md
@@ -9,7 +9,7 @@ title: Qubes R4.0 release notes
 New features since 3.2
 ----------------------
 
-* Core management scripts rewrite with better structure and extensibility, [API documentation](https://dev.qubes-os.org/en/latest/index.html)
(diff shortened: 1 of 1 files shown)


developer/releases/4_0/release-notes.md
```diff
@@ -9,7 +9,7 @@ title: Qubes R4.0 release notes
 New features since 3.2
 ----------------------
 
-* Core management scripts rewrite with better structure and extensibility, [API documentation](https://dev.qubes-os.org/projects/qubes-core-admin/en/latest/)
+* Core management scripts rewrite with better structure and extensibility, [API documentation](https://dev.qubes-os.org/en/latest/index.html)
```
marmarek, line 12: The page you link to has the link you are looking for. It looks like it's `/core-admin/`, not `/qubes-core-admin/`.

PixelPil0t1: Here is the corrected link that should work:
https://dev.qubes-os.org/projects/core-admin/en/latest/qubes.html

Please let me know if this fits better, or if you would prefer a different reference.

PR #1469, opened 2025-05-05
Additional hint on moving user data

GWeck:
Make sure that the private disk is empty and called `Q:`


unman:
@GWeck - is this required? I thought on Forum a user reported it was fine without? (Though I may have not heard right)

GWeck:
On the Forum, a user had the [problem ](https://forum.qubes-os.org/t/windows-appvm-from-windows-templatevm-cant-acess-the-user-profile/33720)that it worked for a template, but not for an AppVM based on it.  This might be caused by a directory `Q:\Users\`, already existing but having the wrong protection. I checked this, and it caused the directory movement to fail. The same thing happened if the private drive had a letter other than `Q:`. Both of these are taken care of by QWT installation but not when using `relocate_dir.exe` alone. So I am trying to clarify this situation.

There may be other problems around, and I try to hunt them down.

unman:
Thanks for this. I'll leave this open until you reach a firm conclusion I can test.

PR #1476, opened 2025-05-21
Updated debug guidance for clarity and simplicity.

seamustuohy:
This update to the salt debugging documentation provides walks the user through the specific steps that need to be followed to get debugging output when running qubesctl.


seamustuohy:
Realized that this is just documenting adding logging the hard way.

PR #1483, opened 2025-07-24
Update guidelines for duplicate issues

andrewdavidwong:
Per @marmarek's request in https://github.com/QubesOS/qubes-issues/issues/2229#issuecomment-3110186417, a new bug in some feature should not prompt us to reopen the old feature request after the initial implementation of that feature has been completed successfully. Instead, we should open a new bug report. Problems associated with the initial implementation of a feature can still merit reopening that feature request.

Since the section on duplicates now refers to issue types, it now makes more sense to put the section on issue types before the section on duplicates.


maiska:
will close this PR mannually, since I cherry-pickerd the single comment because it is more efficient.
Relevant changes  in the rst branch can be seen here https://github.com/QubesOS/qubes-doc/commit/1eea1244d3fb2da24814fc30277243c5e09debec

PR #1485, opened 2025-07-24
Change documentation for Qubes R4.2

GWeck:
Describe usage of old QWT version for W7 and new, experimental version for W10 and W11


GWeck:
That's o.k., I just wished to avoid losing the new texts. I'll wait till then. With this pull request, I'll just wait till the migration has completed.

GWeck:
I started to transfer the texts to the [rst branch](https://github.com/QubesOS/qubes-doc/tree/rst). It's not so bad - the first (and worst) chapter is already transferred, and it looks quite good. After the transfer of all of them, I'll start a pull request there, so that you can check if it integrates well into the new structure.

PR #1490, opened 2025-07-27
HTML theme: style external links and add logos

parulin:
PNG files are copied from [qubes-attachment](https://github.com/QubesOS/qubes-attachment) and [qubesos.github.io](https://github.com/QubesOS/qubesos.github.io) repos.


maiska:
that should be fast, will try to review this this week :)

PR #1491, opened 2025-07-27
Introduction: minor fixes

parulin:
Minimal change, following comments of #1484


maiska:
@parulin same request as the previous PR :) - would you be so kind to reopen/renew after the rst-commit-add force-push several days ago :)

PR #1494, opened 2025-07-27
Playground

parulin:
Correct references and markup on:

* `how-to-install-software.rst`
* `app-menu-shortcut-troubleshooting.rst`

Proposal to use `rst_epilog` in `conf.py` to provide global substitution (with debian as an example).


maiska:
same here, personally will have time to review it properly in a week's time

PR #1495, opened 2025-07-27
Use extlinks extension

parulin:
Provides an easy way to link to Github issues, Forum topics, Qubes website and Wikipedia articles.


maiska:
That is interesting! Will have time to review it in a week though!

PR #1497, opened 2025-07-27
Remove all links pointing to Qubes-Community docs

parulin:
Most of the links have been replaced by their forum counterpart. I wasn't able to find a few links like "Template: Ubuntu", so I removed them, because the link is pointless anyway.

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -43,7 +43,7 @@ Core documentation for Qubes users.
 
    user/hardware/system-requirements
    user/hardware/certified-hardware/certified-hardware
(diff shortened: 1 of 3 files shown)


maiska:
The index is already prepared with direct links to the forums. In this QA phase this is still a content's change thus it is saved for later.

parulin:
> The index is already prepared with direct links to the forums.

I thought that this PR will be obsolete but the current `index.rst` on the main branch is still pointing to github links instead of the forums.

rapenne-s:
This is merged, I had to resolve a conflict so I guess it's why this PR is not closed automatically.

PR #1505, opened 2025-08-04
How to use disposables: qvm-run -> qvm-run-vm

rustybird:
https://forum.qubes-os.org/t/how-to-make-qvm-run-available-in-debian-templates/35252/3


rustybird:
Sure: #1519

PR #1506, opened 2025-08-06
Update How to enter fullscreen mode

parulin:
**This is a content change: do not review before the end of the transition to RTD!**

Update of the page, following this topic: https://forum.qubes-os.org/t/i-cant-fullscreen-video-qubes-os-fix/35283

I removed all references to `guid.conf`, put a first paragraph allowing users to immediately enter fullscreen mode without bothering with security concerns and presented a way to do it with a GUI and a CLI each time.


parulin:
Thanks for the review.

PR #1507, opened 2025-08-09
Fix formattings and other issues

tokideveloper:
This PR fixes all issues I found in issue [#8180](https://github.com/QubesOS/qubes-issues/issues/8180).

This PR could overlap with PRs by @parulin and maybe others.

### DIFF
--- a/developer/services/admin-api.rst
+++ b/developer/services/admin-api.rst
@@ -152,11 +152,11 @@ it easy to set the policy using current mechanism.
      - ``dom0``
      - `-`
      - `-`
-     - ``<property-name> <full-value-as-in-property.Get>``
+     - ``<property-name> <full-value-as-in-property.Get>\n``
      - Get all the properties in one call. Each property is returned on a separate line and use the same value encoding as property.Get method, with an exception that newlines are encoded as literal ``\n`` and literal ``\`` are encoded as ``\\``.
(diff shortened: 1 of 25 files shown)


maiska:
@tokideveloper would you update yout branch to the state of current rst branch or should I?

tokideveloper:
> @tokideveloper would you update yout branch to the state of current rst branch or should I?

Done.

developer/services/admin-api.rst
```diff
@@ -355,9 +349,9 @@ it easy to set the policy using current mechanism.
    * - ``admin.vm.device.<class>.Assign``
      - vm
      - device
-     - assignement-serialization
+     - assignment-serialization
      - `-`
-     - ``device`` is in form ``<backend-name>+<device-ident>`` ``assignment-serialization`` is specified in the section Device Serialization.
+     - ``device`` is in form ``<backend-name>+<device-ident>``; ``assignment-serialization`` is specified in the section Device Serialization.
```
maiska, line 354: Actually in the markdown version there is no ';' but a new line. I think we shoud leave with the markdown version. Would you be so nice to add newlines to such cells where applicable? (I can also do it afterwards. 
I think that ';' as a character may be misleading, though not likely. Thank you for the review of this importart doc!!

PR #1518, opened 2025-08-23
Code block prompts full

maiska:
Fixes to code block pygemnt's highlighting, add __pychache__ to gitignore + add prompts where needed and seen

### DIFF
--- a/user/security-in-qubes/split-gpg-2.rst
+++ b/user/security-in-qubes/split-gpg-2.rst
@@ -23,9 +23,9 @@ Configuration
 
 Create/Edit ``/etc/qubes/policy.d/30-user-gpg2.policy`` in dom0, and add a line like this:
 
-.. code:: console
+.. code:: text
 
-      $ qubes.Gpg2 + gpg-client-vm @default allow target=gpg-server-vm
+      qubes.Gpg2 + gpg-client-vm @default allow target=gpg-server-vm
 
 
 
@@ -33,8 +33,8 @@ Import/Generate your secret keys in the server domain. For example:
 
 .. code:: console
 
-      $ gpg-server-vm$ gpg --import /path/to/my/secret-keys-export
-      $ gpg-server-vm$ gpg --import-ownertrust /path/to/my/ownertrust-export
+      gpg-server-vm$ gpg --import /path/to/my/secret-keys-export
+      gpg-server-vm$ gpg --import-ownertrust /path/to/my/ownertrust-export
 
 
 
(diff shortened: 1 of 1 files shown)


user/security-in-qubes/split-gpg-2.rst
```diff
@@ -64,7 +64,7 @@ Output should be:
 
 .. code:: console
 
-      split-gpg2-client on
+      $ split-gpg2-client on
```
parulin, line 67: That's an output.

user/security-in-qubes/split-gpg-2.rst
```diff
@@ -73,20 +73,20 @@ Export the **public** part of your keys and import them in the client domain. Al
...
+      $ gpg-client-vm$ gpg --import-ownertrust ~/QubesIncoming/gpg-server-vm/ownertrust-export
 
 
 
 This should be enough to have it running:
 
 .. code:: console
 
-      gpg-client-vm$ gpg -K
+      $ gpg-client-vm$ gpg -K
```
parulin, line 89: Same remark about the prompt.

rapenne-s (approved):

rapenne-s:
thanks @maiska for the PR and kudos to @parulin for the good code review

maiska:
> I wonder if having `$` was necessary, it makes copy/pasting harder especially on multiple lines snippets, but do we want people to copy / paste code? 😅
> 
> Usually, root prompt starts with `#` whereas a non root prompt starts with `$`, I saw some `#` replaced by `$`, I'm not sure if it's a style we want to have here.

there is an extention which I had in mind https://pypi.org/project/sphinx-copybutton/ where the last update is from 2023, but it is still in use as of my search. there is also another one https://pypi.org/project/sphinx-prompt/ which can be utilized. We can postpone this PR for after merge to main and do it properly as you pointed out.

maiska:
the above suggested enhancements will be addressed as later todo's

parulin:
> I wonder if having `$` was necessary, it makes copy/pasting harder especially on multiple lines snippets, but do we want people to copy / paste code? 😅

On Mullvad browser, even with Javascript disabled, I can't copy the prompt part on multilines. If you want to check, look for a prompt starting with `$ curl` here:

https://qubes-doc--1518.org.readthedocs.build/en/1518/developer/code/source-code.html

rapenne-s:
> > I wonder if having `$` was necessary, it makes copy/pasting harder especially on multiple lines snippets, but do we want people to copy / paste code? 😅
> 
> On Mullvad browser, even with Javascript disabled, I can't copy the prompt part on multilines. If you want to check, look for a prompt starting with `$ curl` here:
> 
> https://qubes-doc--1518.org.readthedocs.build/en/1518/developer/code/source-code.html

Thanks for checking this, this is good then for copy/paste!

PR #1527, opened 2025-09-11
fixes for side-by-side preview

saces:
a fixed typo and some additions for the esbanio preview to work

https://forum.qubes-os.org/t/editing-the-qubes-doc-with-side-by-side-preview/35867

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,3 +30,5 @@ dev = [
     "sphinx-rtd-theme>=3.0.2,<4",
 ]
 
+[tool.esbonio.sphinx]
+pythonCommand = ["${venv:.venv}"]


rapenne-s:
This was merged, but I'm not a fan of having `.venv` in the .gitignore as there are no standard for the venv path as far as I know. This might work for OP, but I use venv, the contributing guide mentions a .q_env outside the git directory, everyone might have their own way.

maiska:
> This was merged, but I'm not a fan of having `.venv` in the .gitignore as there are no standard for the venv path as far as I know. This might work for OP, but I use venv, the contributing guide mentions a .q_env outside the git directory, everyone might have their own way.

good point: i will change q_venv to .venv, it is consistent with all the documentation out there, and having it in .gitignore as a fall back, does not hurt anyone imho :)

PR #1530, opened 2025-09-12
change sphinx output format to dirhtml

saces:
this produces prettier urls without '.html' at the end

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -10,7 +10,7 @@ submodules:
   recursive: true
 
 sphinx:
-  builder: html
+  builder: dirhtml
   configuration: conf.py
   fail_on_warning: false
 


parulin:
> > I think it is too late now? All the current links with `.html` have to be redirected to the "clean links" and if it takes too much time, all the links containing anchors will be lost again.
> 
> no, i don't think so. there should not be anchors in the redirects, and if it is sufficiently tested and run by the link checker in rst itself we should be good. i do not see a lot of overhead here, but it will be approached at a slow pace and testing is needed

Sorry, I wasn't clear. I had the forum in mind, or any external source: any link containing an anchor is lost.

I.e.: https://www.qubes-os.org/doc/glossary/#qube will redirect to https://doc.qubes-os.org/en/latest/user/reference/glossary.html, not https://doc.qubes-os.org/en/latest/user/reference/glossary.html#qube

So, moving again from https://doc.qubes-os.org/en/latest/user/reference/glossary.html to https://doc.qubes-os.org/en/latest/user/reference/glossary/ will have the same effect.

marmarek:
Look what I found in RTD settings:
image

rapenne-s:
> Look what I found in RTD settings: image

Did you enable it? Do we want to make a change about the .html at the end of URLs?

marmarek:
> We could still implement `.html` pages that would redirect to their regular counterparts

You mean a page with section names, that redirect to appropriate section on the new URL? Yeah, I agree it might work, but probably not practical to do (I don't see any ready-made solution for that). OTOH, if just redirecting plain pages (and ignoring that links to individual pages will break now), there is a one-click option for that in RTD settings.

PR #1535, opened 2025-09-14
Refactor glossary: clarifications, new section and more consistent markup

parulin:
This PR is trying to clarify the glossary by using the field markup for recurring fields like "Previously known as". Some examples are provided.

Some inaccuracies are corrected and some sentences are reorganized to try to make more sense.

A new section "Confusions between qube types" is trying to provide another way to understand the terminology.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -132,6 +132,8 @@ gettext_uuid = True
 
 # Define a block of reusable reStructuredText (reST) snippets, warnings etc. that Sphinx automatically appends to every source file before it is parsed
 rst_epilog = """
+.. |fedora-version| replace:: 42
 .. |debian-codename| replace:: bookworm
 .. |debian-version| replace:: 12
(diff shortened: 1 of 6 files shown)


unman:
I think the "Confusions" table is misleading, and some of the entries inconsistent with the above text. The entries here are more nuanced than is presented - eg disposables (named or not) are app qubes, and therefore qubes; an HVM may  be a disposable template; a service qube may be a standalone or HVM; a standalone may be a template.
Either this needs more detail or it should be dropped.

unman:
Is this still a draft?
Should it be dropped in favor of the more recent offering on glossary?

parulin:
I started to work on this a long time ago. It was obviously a draft and I shared it as such, but I didn't want this to waste your time. Sorry. I'm closing it because the past and current PRs are better.

PR #1537, opened 2025-09-15
Windows toctree

parulin:
At first I wanted to convert the list in `user/templates/windows/windows.rst` to a real toctree in order to fix some warnings. I ended up converting it to a real index page.

Nice side effect there is a better breadcrumbs on the Windows pages.


rapenne-s:
There is a conflict on conf.py now, I guess it's since the merge of a windows tree change, could you take a look?

PR #1539, opened 2025-09-15
Hints on disposables and correct nft syntax error

GWeck:


parulin:
Your second version looks better! Maybe you can [create a reference](https://doc.qubes-os.org/en/latest/developer/general/rst-documentation-style-guide.html#cross-referencing) with this syntax:

``` rst
.. _where-to-put-firewall-rules-disposables:
```

And link that each time the docs mentions to "you should write/store"? That way will avoid duplicate content but gives a hint to the user, like your first commit?

GWeck:
I added the references.

By the way, the really relevant part of this change is the minor change in the script, because without that, it will silently fail.

rapenne-s:
Good catch, thanks

PR #1547, opened 2025-09-24
Add style to external links in the menu

parulin:
Should fix QubesOS/qubes-issues#10260

This is a draft because I need help to figure out where I should put the CSS file. I just copied the Readthedocs conf.py


parulin:
Following [another discussion](https://github.com/QubesOS/qubes-doc/pull/1610#issuecomment-3889085661), I think it is better to publish this PR without waiting for the correct place for the CSS file.

After playing with the templates, I think it is okay. It might even be better to remove `attachment/doc` from `html_static_path`. All it does is copying useless files to the build dir.

PR #1549, opened 2025-09-28
Remove outdated commands to run grub2-mkconfig

parulin:
Following a [forum
topic](https://forum.qubes-os.org/t/how-to-change-boot-options-for-usb-boot/36387/2), it seems that the documentation still referenced some outdated way of running `grub2-mkconfig` command.


rapenne-s:
merged! Thanks!

PR #1551, opened 2025-10-03
Update UpdatesProxy to new policy format

deeplow:


unman:
This is fine as is. Thanks for the PR. 

If there is a PR that affects only 4.2, make it against 4.2

If it is NOT for 4.2, make it against main (We'll have to work out what
happens with docs in main for features that dont make it in to the
latest release, if we are ever in that happy position.)

If it is for BOTH 4.2 AND main, you can either PR against both, or PR
against main with note in commit message - something like "affects X.X
also".

rapenne-s:
thanks!

merged and in main and backported to r4.2 branch

PR #1553, opened 2025-10-03
Add a Howto on setting wallpapers

marmarta:
A small HowTo on setting a wallpaper, inspired by a mastodon exchange.

I'm not sure if my rst is correct here, this is my first contribution to new docs.


parulin (commented):
You asked for review so I ended up with two kind of comments:

1. the hard wrap and the code-block ones (useful?)
2. the minor ones (all the others, feel free to ignore them!)

Otherwise, I think that this could be a really helpful how-to, nice catch!

marmarta:
Thank you so much @parulin for the review, I'll push the fixes in a second

parulin:
You're welcome! But the code-block is still the same and doesn't work. Take a look at the preview: https://qubes-doc--1553.org.readthedocs.build/en/1553/user/how-to-guides/how-to-set-a-wallpaper.html

PR #1598, opened 2025-12-24
Change icons and title when build on RTD from a PR

parulin:
By checking some environment variable from Read The Docs in conf.py, we can determine if the build is made from a PR build. In that case, the HTML title is slightly different and instead of normal icons, we use red-tinted ones.

That way, putting some links to the PR builds overview will be clearing distinct from the official build.


maiska:
I am not sure that red coloring is the best way to approach this. It adds some uneasy feeling to the person viewing the docs.
I think it suffices that when a build is made from a PR - it is highly visible in the url, you cannot miss it :) and there is the option on the side, to show a diff :)

maiska:
OK I decided to close this PR. A long time has passed.

I would not like to change the branding of Qubes OS (blue Q) since the branding is here more important  imho.
Danger can be intepreted as in 
1. the danger is the Qubes OS official documentaiton (a cue of red and untrustworthy as in color coded qubes themselves)
2. or danger: this is just a PR

A banner is always added with a PR + visual diff RTD addon is on (https://docs.readthedocs.com/platform/stable/visual-diff.html), so one could also see the changes from the PR and the latest build.


If you think it is still insufficient, why not add a custom css making use of the visual diff addon and changing the colors of the diff itself to something more contrastful? What do you think

parulin:
Let's say it's not my priority so I will keep it like that :)

PR #1662, opened 2026-03-26
Update fedora and debian upgrade docs

### LINKED ISSUE #9462: Documentation about template upgrades is very alarming
[How to file a helpful issue](https://www.qubes-os.org/doc/issue-tracking/)

### The problem you're addressing (if any)

The documentation about template upgrades is very alarming and offputting to users.

### The solution you'd like

Better documentation that does not scare users.

### The value to a user, and who that user might be

Users will not be scared to upgrade their templates or to use them.

### Completion criteria checklist

(This section is for developer use only. Please do not modify it.)

nihalxkumar:
change scary alerts and simplify language.

Closes: https://github.com/QubesOS/qubes-issues/issues/9462

### DIFF
--- a/user/templates/fedora/fedora-upgrade.rst
+++ b/user/templates/fedora/fedora-upgrade.rst
@@ -4,13 +4,10 @@ How to upgrade a Fedora template in-place
 
 .. warning::
 
-      This page is intended for advanced users.
+    Most users should simply :doc:`install a fresh template <user/templates/fedora/fedora:installing>` instead of performing an in-place upgrade.
+    Use the in-place upgrade method below only if you have made customizations or installed additional packages in the current template that you want to preserve.
 
-.. DANGER::
-
-      **Warning:** This page is intended for advanced users only. Most users seeking to upgrade should instead :ref:`install a new Fedora template <user/templates/fedora/fedora:installing>`      . Learn more about the two options :ref:`here <user/templates/fedora/fedora:upgrading>`      .
-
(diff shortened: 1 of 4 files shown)


user/templates/fedora/fedora-upgrade.rst
```diff
@@ -4,13 +4,10 @@ How to upgrade a Fedora template in-place
 
 .. warning::
 
-      This page is intended for advanced users.
```
maiska, line 7: Hey, I have a suggestion, since it is the convetion to use the warning 'This page is intended for advanced users.', do not remove it but add a note instead of danger with the text below, The same for above :)

user/templates/fedora/fedora.rst
```diff
@@ -53,8 +53,8 @@ Upgrading
...
-
-- **Advanced:** :doc:`Perform an in-place upgrade of an existing Fedora template. </user/templates/fedora/fedora-upgrade>` This option will preserve any modifications you’ve made to the template, **but it may be more complicated for less experienced users.**
-
+- **Recommended for most users:** :ref:`Install a fresh template <user/templates/fedora/fedora:installing>`.
+  After installing the new template, redo any custom modifications you need and :ref:`switch all qubes that were using the old template to the new one <user/templates/templates:switching>`. You may want to note down your customizations (or check ``dnf history`` in the old template) so you can easily recreate them.
 
+- **Alternative (in-place upgrade):** :doc:`Upgrade the existing template in-place </user/templates/fedora/fedora-upgrade>`.
```
maiska, line 59: In these cases (also debian) I would add Advance to Alternative as in
```
Advanced alternative (in-place upgrade):
```

PR #1671, opened 2026-04-11
Clarify downloading installer under R4.3

GWeck:


user/templates/windows/qubes-windows-tools.rst
```diff
@@ -226,7 +226,15 @@ Installing the Qubes Windows Tools on Windows 7, 10, and 11, both as a Standalon
 
    - For use with **Windows 10 and 11**, you have to install the new version 4.2.2 of Qubes Windows Tools. As this new QWT version is only provided for Qubes R4.3, it has to be installed by downloading the .rpm file from the repository (version `4.2.2-1 <https://yum.qubes-os.org/r4.3/current-testing/dom0/fc41/rpm/qubes-windows-tools-4.2.2-1.fc41.noarch.rpm>`__ ) if you want to use it in Qubes R4.2.
 
-   After downloading, copy the file to `dom0` as described in `How to copy from dom0 <https://www.qubes-os.org/doc/how-to-copy-from-dom0/#copying-to-dom0>`__ and install it via ``sudo dnf install PATH_TO_RPMFILE``.
+     In Qubes R4.2, after downloading, copy the file to `dom0` as described in `How to copy from dom0 <https://www.qubes-os.org/doc/how-to-copy-from-dom0/#copying-to-dom0>`__ and install it via ``sudo dnf install PATH_TO_RPMFILE``.
```
maiska: Since this will go into r4.3 and main i suggest omitting *all* the info about r4.2 (it will still be there in the r4.2 branch). So that there is a clear separation. And perhaps add a sentence and referal to r4.2 branch documentation? What do you think?

GWeck: Removing the information for R4.3 will surely make the text easier to understand.

So, I shortened it considerably, also removing the text for Windows 7 direct installation. The package is still in the repositories for R4.2, but no longer for R4.3. So, for R4.3 and Windows 7, only the manual download path is still valid.

Where I am not so sure is the complete removal of the text for R4.2 and the new QWT version. Will users of R4.2 always be directed to the old documentation, or could it happen that they find themselves in the R4.3 version, where they will be lost at this point?

maiska: The documentation about the r4.3 is the default one, when one visits the doc.qubes-os.org. And it might on occasions slip the attention of the reader. If there is a clear separation between version of QWT docs we can add a banner at the start of the file with referal to the respective other version as a note - for 4.3 -> to 4.2 and vice versa. 

I am a little confused - why removing the information for R4.3. ("Removing the information for R4.3 will surely make the text easier to understand.")  Did you mean perhaps R4.2?

GWeck: > Removing the information for R4.3

You're right, it should be R4.2. Sorry for the confusion.

I updated the note at the beginning and added a short note here, specifying the download location. I hope that helps.

maiska: Hey, I think more work is needed to clearly have separate documentations for r4.3 and r4.2. and it will take considerably more time than this small but vital change.
I merged [here](https://github.com/QubesOS/qubes-doc/commit/ecb5f368fff07b3ba4bd9608e935b29e6dfdcbc4) and [there](https://github.com/QubesOS/qubes-doc/commit/8079f9c9be1a2155a51f681414edc2c4497e8863) :) and will close this PR manually

GWeck:
Thank you!

PR #1674, opened 2026-04-13
Some glossary improvements

marmarta:
Rewrote some entries to be more grammatical / human readable.
Added VM icons.
Rewrote some related paragraphs to be simpler and more readable.


user/reference/glossary.rst
```diff
@@ -30,43 +22,68 @@ Compartment nomenclature
...
 
    app qube
-      Any :term:`qube` that does not have a root filesystem of its own. Every app qube is based on a :term:`template` from which it borrows the root filesystem.
+      .. image:: https://raw.githubusercontent.com/QubesOS/qubes-artwork/refs/heads/main/icons/scalable/apps/appvm-red.svg
+         :width: 24px
+         :align: left
+
+      A :term:`qube` that is based on a template. Normally, *app qubes* are used to run user applications and store user files.
+
+      An *app qube* does not have a root filesystem of its own. The qube borrows its root filesystem from its  :term:`template`, and only owns its own home directory and user files (in Linux-based qubes, they are the /home and /user/local directories).
```
maiska: Would you highlight the paths with :file:? I could also do this when merging!

user/reference/glossary.rst
```diff
@@ -30,43 +22,68 @@ Compartment nomenclature
...
+         :width: 24px
+         :align: left
+
+      A :term:`disposable` qube is a stateless :term:`qube` - it does not save any data between reboots. Every time it is started, it has the same fresh filesystem.  These qubes can be used for a variety of situations, from experimentation to creating a more secure network connection. See :doc:`/user/how-to-guides/how-to-use-disposables`.
+
+      Typically, disposable qubes have a temporary name (**dispXYZ**, where XYZ is a random number). They cease to exist after the qube is shut down, and closing the first application that was opened in the disposable will trigger the qube to shut down.
+
+      Multiple instances of a disposable qube based on a single disposable template can run at the same time.
+
+      See also :term:`named disposable`).
```
maiska: `)` too much :)  sam eas above, can alter when merging

user/how-to-guides/how-to-use-disposables.rst
```diff
@@ -31,7 +31,7 @@ Disposable template
...
+:term:`Disposable template <disposable template>` is not a disposable in itself, but a special template that can create different disposable types: "normal" (unnamed) :term:`disposables <disposable>` and :term:`named disposables <named disposable>`. When there is need to customize the files of a disposable, it must be done on the disposable template or the template itself to be reflected in every new disposable based on that template. By default, Qubes OS creates the ``default-dvm`` disposable template, we will use it as an example on this page, but you can create as many disposable templates as you'd prefer. :doc:`Disposable customization </user/advanced-topics/disposable-customization>` is outside of the scope of this page.
```
ben-grande: Consider that the "normal" installation has disposable sys-net and sys-usb, these are named disposables. I don't like removing "unnamed" from the term in the glossary as it leaves room for ambiguity. There are two types of disposables, "named" and "unnamed", and having these attributes in the name makes it explicit of what it is, even though not being enough to fully comprehend its capabilities.

I think that removing the term also makes it difficult to search, one would search for "disposable", "normal" and "named" would appear, without this change, it is possible to search specifically for "unnamed disposable".

user/reference/glossary.rst
```diff
@@ -78,67 +95,65 @@ Qube's types
 
       - Previously known as: ``TemplateVM``.
 
-
-Qube's types variations
```
ben-grande, line 82: This cannot be merged with "Qube types", because it is not the same thing.

- "Qube types": one-to-one correlation with klass
- "variations": not klass bound, but role bound

Using different headers/sections also helps guarantee that new changes to this doc, such as adding a new qube type, such as remote qube, will be added close to the klass correlation, rather than anywhere else on that big section (which was the case in the past before I added headers for people think of terms in groups).

user/reference/glossary.rst
```diff
@@ -78,67 +95,65 @@ Qube's types
...
-      - Previously known as: ``DisposableVM Template``, ``DVM Template``, ``DVM``. It is advised against the use of the ``DVM`` terms as it can be interpreted by some users as an abbreviation of ``DispVM``, which a ``DVM`` is not.
+      - Previously known as: ``DisposableVM Template``, ``DVM Template``, ``DVM``.
 
    named disposable
-      A type of :term:`disposable` given a permanent name that continues to exist even after it is shut down and can be restarted again.
+      .. image:: https://raw.githubusercontent.com/QubesOS/qubes-artwork/refs/heads/main/icons/scalable/apps/dispvm-red.svg
+         :width: 24px
+         :align: left
+
+      A type of :term:`disposable` given a permanent name. This qube continues to exist even after it is shut down and can be started again. In all other respects, it behaves like a disposable qube.
```
ben-grande: Named disposable doesn't shut down when the application that was first initiated in them, closes, so it breaks this rule of "In all other respects". Maybe add it also?

user/reference/glossary.rst
```diff
@@ -78,67 +95,65 @@ Qube's types
...
--------------
+Other terms
+-----------
 
 
 .. glossary::
 
    dom0
-      :term:`domain` zero. A type of :term:`admin qube`. Also known as the **host** domain, dom0 is the initial qube started by the Xen hypervisor on boot. Dom0 runs the Xen management toolstack and has special privileges relative to other domains, such as direct access to most hardware.
+      A type of :term:`admin qube`. Also known as the **host** :term:`domain`, dom0 is the initial qube started by the Xen hypervisor on boot. Dom0 runs the Xen management toolstack and has special privileges relative to other domains, such as direct access to most hardware.
```
ben-grande, line 156: Do you think that referring to "domain zero" being the first qube to be started by Xen makes it easier to understand? I actually like that it explains that "dom" stands for domain and "zero" being the first.

user/reference/glossary.rst
```diff
@@ -2,26 +2,18 @@
...
-
 .. glossary::
 
    Qubes OS
-      A security-oriented operating system (OS). The main principle of Qubes OS is security by compartmentalization (or isolation), in which activities are compartmentalized (or isolated) in separate :term:`qube`.
+      A security-oriented operating system (OS). The main principle of Qubes OS is security by compartmentalization (isolation), in which activities are compartmentalized (isolated) in separate :term:`qubes <qube>`.
 
       - **Important:** The official name is "Qubes OS" (note the capitalization and the space between "Qubes" and "OS"). In casual conversation, this is often shortened to "Qubes". Only in technical contexts where spaces are not permitted (e.g., in usernames) may the space be omitted, as in ``@QubesOS``.
 
-Compartment nomenclature
```
ben-grande, line 17: I'm afraid that removing the headers will make a mess of the glossary. I added them because the terms were scattered across this page and I like to keep related concepts close to each other. Now, they are close, but after many people start changing this page, it can become messy if there is nothing mandating order.

Do you think there are too many sections?

unman:
Personally I would prefer "unnamed disposable" to die the death - it's
always jarred with me, because those disposables are named, of course. I
think the term first surfaced on the Forum, and had little traction.
It seems to me that it's more used now, and that you ***@***.***) use
it heavily.

user/how-to-guides/how-to-use-disposables.rst
marmarta: However, unnamed is a pretty silly name for something that very clearly has a name. I don't like the term, I think that for most intents and purposes it makes sense to think of disposables and a special case of named disposables.

user/reference/glossary.rst
marmarta: Ok but "qube's types variations" doesn't mean anything in English. I don't want really ungrammatical things in the docs. I'm also not sure what this is supposed to be, because here you have some properties, some features, some general concepts...

ben-grande: There is not a term for this yet. I had no better idea at the time than "variations". Maybe "roles"?

marmarta:
> Personally I would prefer "unnamed disposable" to die the death - it's always jarred with me, because those disposables are named, of course. I think the term first surfaced on the Forum, and had little traction. It seems to me that it's more used now, and that you ***@***.***) use it heavily.

I'm glad someone agrees with me on the weirdness of this term :D I really don't think it's that clarifying, compared to the larger type of disposable and the subtype of named disposable.

user/reference/glossary.rst
marmarta: I don't think this is a helpful header in the docs though. This sounds like a developer docs kind of terminology, not user docs.

user/reference/glossary.rst
marmarta: ok

user/reference/glossary.rst
marmarta: I don't really see how it clarifies anything about qubes architecture though?

user/reference/glossary.rst
marmarta: I don't think those sections are very helpful, because glossaries tend to be used like dictionaries: not to be read through top-to-bottom, but to clarify an unknown term. I don't like those headers in particular because I don't think they are clear and explanatory and they definitely use too complicated words (yes, I'm also to blame for using C2 words in a B1 context, but I welcome corrections on that, really, and nomenclature is just too heavy a word here).

user/how-to-guides/how-to-use-disposables.rst
ben-grande: I agree the name is silly. Maybe we could have a better term than "named" and "unnamed"? The difference is that "unnamed" has `auto_cleanup` property enabled, which some tools (e.g. `qvm-run`) check to call `cleanup()` if the first connection ceases.

user/reference/glossary.rst
maiska: I would agree there of user friendlness or easyness of reading.
I think it is important that we do not add too much information to the cognitive load of getting into Qubes OS for new users.
For advanced users there are the advanced docs, and the devs, well they have code :) Joke aside, there are the dev docs. 
 It is imho not incorrect to refer to all of the variations/roles/types as one, since at the end they are lvms :)

user/reference/glossary.rst
```diff
@@ -78,67 +95,65 @@ Qube's types
...
+         :width: 24px
+         :align: left
+
+      A type of :term:`disposable` given a permanent name. This qube continues to exist even after it is shut down and can be started again. Unlike normal disposable qubes, it is also not tied to the program it was started with and will not shut down when the program closes. In all other respects, it behaves like a disposable qube.
 
       - Only one instance of a named disposable can run at a time.
 
       - Technical note: Named disposables are useful for certain :term:`service qube`\ s, where the combination of persistent device assignment and ephemeral qube state is desirable.
 
-   unnamed disposable
```
maiska, line 108: The unnamed disposable refered in other parts of the docuemnation should be trimed down to just disposable as per this glossary imho :) 
An example would be:
how-to-use-disposables.rst: How to create unnamed disposables , l 66
or also the header Unnamed disposables

user/reference/glossary.rst
```diff
@@ -30,43 +22,68 @@ Compartment nomenclature
...
-------------
+Qube types and properties
+-------------------------
 
 
 .. glossary::
 
    admin qube
-      A type of :term:`qube` used for administering Qubes OS.
+      .. image:: https://raw.githubusercontent.com/QubesOS/qubes-artwork/refs/heads/main/icons/scalable/apps/adminvm-black.svg
```
parulin: Some comments (about all the images):

* ["Do` not attempt to link to images hosted on other websites."](https://doc.qubes-os.org/en/r4.3/developer/general/rst-documentation-style-guide.html#images)
* If you want an empty alt, you have to be explicit, like this:
  ```rst
  .. image:: path
     :alt:
     :width: [...]
  ```
* I don't have a solution for this, but putting the image on the left creates an offset, a bit inconsistent with the other definitions.

marmarta: Right, but I do not want to create a situation where changing icons requires manually scouring different repos to figure out where also they need to be changed. This is not a link to a random external website, this is our repository.

marmarta: unless there is a nicer way to do it, but I think that manual copying and praying that when in 4 years someone does adjustments they remember to also change the docs is not a good approach...

user/how-to-guides/how-to-use-disposables.rst
ben-grande: I checked some synonyms for "persistent" and "disposable" and tried to find something that could be a single word attribute (or combined words) that is not high level english. What about:

- "reusable disposable" for named disposable
- "single-use disposable" for unnamed disposable

It is very hard to explain in a term (in a phrase it is easy) that all disposables have disposable volume, but not disposable qube name. I think that using "reusable" and "single-use" is decent enough.

marmarta: I don't think it makes sense to treat those as two equivalent classes; I think that "general disposable, which is by default the dispXYZ kind" and a "special named subtype" makes more sense.

marmarta:
@parulin - your way of dealing with icons is truly better, hats off :)

user/reference/glossary.rst
maiska: I think it is also against our own guideline where we say - please preferably do not use urls :)
Def there is a way to get around this, be it with custom directive, which say have a second attribute a url and a timestamp, and can be checked against that while building, or some other more slick approach.
I think it is OK for now, but will make a todo out of this :)

user/reference/glossary.rst
maiska: @marmarta  @ben-grande thoughts about the above?

ben-grande: > The unnamed disposable refered in other parts of the docuemnation should be trimed down to just disposable as per this glossary

Yes.

user/reference/glossary.rst
parulin: > Def there is a way to get around this, be it with custom directive, which say have a second attribute a url and a timestamp

Tell me if I'm wrong but we can't: we don't trust infrastructure.

maiska: You are not wrong and it is not what I am suggesting here. 
Or something else: why would you think that the infrastructure is trusted at all? :)
I am merely low key brainstorming what can be a convenient solution for such a thing.
f.ex.  a custome image directive where a url can be placed as an attribute, but not using it as a path, cz not good.
And a timestamp or not for when the image was introduced/last edited.
Then it can be checked against the last update on the uri of the image if there are new edits, if it is still there etc and spill out a warning when building the docs.

ben-grande: Can the URL be `file://`, then it can load from package qubes-artwork.

marmarta: This wouldn't work online though?

user/reference/glossary.rst
marmarta: done

user/reference/glossary.rst
marmarta: done

user/reference/glossary.rst
marmarta: done

user/reference/glossary.rst
marmarta: alright, I just put the images in this repo, you have convinced me everyone :)

unman (commented):
Resuming review in new session

unman (commented):
Resuming review in new session

unman (commented):
Ready to go I think with some minor changes - I prefer file system throughout, some minor spelling.

PR #1675, opened 2026-04-15
Add warning re unofficial and external guides to Menu and SideBar

unman:


parulin:
Instead of using custom HTML and CSS code, you can reuse the current theme by:

1. removing the inline style
2. adding a CSS class to the div (a bit hacky?):
   ```js
   warningDiv.classList.add('rst-content')
   ```
3. using this HTML as innerHTML:
   ```html
   <div class="admonition warning">
       <p class="admonition-title">Warning</p>
       <p>This is unofficial content provided by community members or third parties. These guides have not been reviewed by the Qubes team. Use them at your own risk.</p>
   </div>
   ```
4. adding this custom CSS to `_static/css/custom.css`:
   ```css
   .wy-nav-side .admonition {
       color: #404040;
   }
   ```

The output is more consistent and might be easier to understand.

_static/custom.js
```diff
@@ -0,0 +1,25 @@
+document.addEventListener('DOMContentLoaded', function() {
+  const captions = document.querySelectorAll('p.caption');
+  
+  captions.forEach(caption => {
+    if (caption.textContent.toLowerCase().includes('external')) {
```
parulin: Using `.startsWith()` instead of `.includes()` might prevent some trouble in future. Like a caption being "How to use external devices"?

unman:
I tried that, of course, but the CSS didnt render properly for me.
I'll try this although I dont feel it's easier to understand.

unman:
On Fri, Apr 17, 2026 at 05:08:48AM -0700, parulin wrote:
> @parulin commented on this pull request.
> Using `.startsWith()` instead of `.includes()` might prevent some trouble in future. Like a caption being "How to use external devices"?

I wondered about that - but we control the captions and I thought
.includes gave us more flexibility.

parulin:
> I wondered about that - but we control the captions and I thought .includes gave us more flexibility.

Right, and the pattern is quite obvious.

PR #1678, opened 2026-04-17
Use r4.3 folder instead of 4.3

parulin:
Fix a little inconsistency from 7e8977e5


user/how-to-guides/how-to-use-devices.rst
```diff
@@ -37,7 +37,7 @@ When clicking on the tray icon, several device-classes separated by lines are di
 
 On most laptops, integrated hardware such as cameras and fingerprint-readers are implemented as USB-devices and can be found here.
 
-.. image::  /attachment/doc/4.3/qubes-devices-widget.png
+.. image::  /attachment/doc/r4.3/qubes-devices-widget.png
    :alt:
    :width: 100%
```
maiska, line 43: Hey, thank you.
Good point. There are different paths across the board.
Would you care adjusting that too (see f.ex. line 29 of this file), and removing the directory doc/4.3 altogether?

parulin: I don't think I can do anything about `doc/4.3`, because of how git works?

maiska: ok i'll take care of all of that

PR #1684, opened 2026-04-20
Fix errors and warnings recently added to the docs

parulin:
Sorry about that!


unman:
I believe this has already been resolved.

parulin:
Not only has it been resolved, but this PR was totally wrong!

Except for one thing, maybe. Is it intentional to remove the text telling the user about default MAC randomization? The commit message is "removed bogus entry from index" (see: @9edd5d251face) but the issue there is the missing custom reference label, @maiska?

PR #1693, opened 2026-04-28
Add section in installing software guide on running scripts after package operations. Closes QubesOS/qubes-issues#10456

unman:


user/how-to-guides/how-to-install-software.rst
```diff
@@ -169,6 +169,15 @@ Standalones
 
 The process for installing and updating software in :term:`standalones <standalone>` is the same as described above for templates, except no qubes are based on standalones, so there are no other qubes to restart.
 
+
+Running scripts after package installation or removal
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+**In Fedora templates**, after a successful `dnf install`  or  `dnf remove` a hook runs :file:`/etc/qubes-rpc/qubes.PostInstall`  which executes any shell scripts in :file:`/etc/qubes/post-install.d/`.
```
parulin, line 176: It would be better to use double backticks around the `dnf` commands.

unman:
On Tue, Apr 28, 2026 at 06:58:04AM -0700, parulin wrote:
> It would be better to use double backticks around the `dnf` commands.

Why? Is it clearer to the user? More aesthetically pleasing?
I initially used double backticks, but thought that single worked better.

parulin:
With single backticks, it is an "interpreted text" while with two backticks it is just a "literal text". The first one is like a "default role", something we might consider in the future?

https://docutils.sourceforge.io/docs/user/rst/quickref.html#inline-markup

Sphinx recommends the second one for inline markup:

https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#inline-markup

I'm not 100% sure on this but it makes me think that using the double version right away will avoid some trouble in the future if we decide to use the default role for something specific.

unman:
But here the words are not code samples, and if anything they *are*
program identifiers - if I understand that.
In context it is just better.

parulin:
Right, I agree with you.

unman:
@maiska Any view on this?

PR #1709, opened 2026-05-17
Add warning about unofficial and external guides to Menu and SideBar Enhance warning about using external documentation

unman:


unman:
I am uncertain whether this should be a Warning or a Danger notice.
Open to thoughts?

PR #1712, opened 2026-05-18
Remove duplicated template versions

parulin:
The current version of the templates in `user/templates` now refers to the Supported releases page, to avoid forgetting to update it.

The unused `|debian-version|` substitution is removed.

### DIFF
--- a/user/templates/debian/debian.rst
+++ b/user/templates/debian/debian.rst
@@ -3,7 +3,7 @@ Debian templates
 ================
 
 
-The Debian :term:`template` is an officially :ref:`supported <user/downloading-installing-upgrading/supported-releases:templates>` template in Qubes OS. The Current version is Debian |debian-version| (“|debian-codename|”). It is available in 3 versions - ``debian-13-xfce``, the default Debian template with the Xfce desktop environment; ``debian-13``, an alternative Debian template with the GNOME desktop environment; and ``debian-13-minimal``. This page is about the “full” templates. For the minimal version, please see the :doc:`Minimal templates </user/templates/minimal-templates>` page.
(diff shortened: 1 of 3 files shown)


unman (commented):
I recall that Nina was quite opposed to making users click through for essential information. So I am against this for that reason.
No problem with referring on to explanatory pages, glossary etc, but do not remove essential information.

parulin:
Ok, the starting point of this was an obsolete reference to Fedora 42 that has been missed in f1167175 I wanted to avoid this, but this is the wrong way of doing it, so I will close it.

Before that, what do you think about defining the supported versions in `conf.py` and putting them as rst substitutions into `rst_epilog`? Then, when needed, we can use something like `|fedora-supported-versions|` on any page.

Another option is to document somewhere which pages should be manually changed when bumping the supported versions.

user/templates/debian/debian.rst
```diff
@@ -3,7 +3,7 @@ Debian templates
...
+The Debian :term:`template` is an officially :ref:`supported <user/downloading-installing-upgrading/supported-releases:templates>` template in Qubes OS. The current version is listed in the :ref:`Templates section of Supported releases <user/downloading-installing-upgrading/supported-releases:Templates>`. It is available in 3 versions - ``debian-13-xfce``, the default Debian template with the Xfce desktop environment; ``debian-13``, an alternative Debian template with the GNOME desktop environment; and ``debian-13-minimal``. This page is about the “full” templates. For the minimal version, please see the :doc:`Minimal templates </user/templates/minimal-templates>` page.
```
maiska, line 6: I would second unman.  Why not both: citing the current verison immediately AND reffering further?

maiska (changes requested):

user/templates/debian/debian.rst
parulin: I'm okay with that, before updating this PR, I wanted some advice about this:

> Before that, what do you think about defining the supported versions in conf.py and putting them as rst substitutions into rst_epilog? Then, when needed, we can use something like |fedora-supported-versions| on any page.
> 
> Another option is to document somewhere which pages should be manually changed when bumping the supported versions.

unman:
Before that, what do you think about defining the supported versions in  conf.py  and putting them as rst           
  substitutions into  rst_epilog ? Then, when needed, we can use something like  |fedora-supported-versions|  on any  
  page.                                                                                                               
                                                                                                                      
  Another option is to document somewhere which pages should be manually changed when bumping the supported versions.

I dont want to use conf.py for this purpose.
As to documenting, I think it's unnecessary -
On bumping supported versions, the **required** change is in supported-releases. There *may* be changes in the fedora and debian pages. There *may* also be changes required throughout the documenitation as templates become eol. A simple grep should be sufficient to find instances of use of eol templates.

parulin:
Okay, let's close it then.

PR #1714, opened 2026-05-20
Replace code directive by the sphinx code-block in examples

parulin:
The `code::` directive is a directive compatible with docutils, allowing less options than the `code-block::` or `sourcecode::` Sphinx directives.

See this comment: https://github.com/sphinx-doc/sphinx/issues/14411#issuecomment-4440190135

This PR replaces `code::` by `code-block::` only the documentation style guide and the how-to.

### DIFF
--- a/developer/general/rst-documentation-style-guide.rst
+++ b/developer/general/rst-documentation-style-guide.rst
@@ -28,10 +28,10 @@ provided by `Docutils <https://www.docutils.org/>`__ and extended by `Sphinx <ht
 Directives are used to insert non-paragraph content, such as images, tables, and code blocks.
 Example directives are:
 
-.. code:: rst
+.. code-block:: rst
 
    .. image::
-   .. code::
+   .. code-block::
    .. figure::
 
 Directives start with ``..``, followed by directive name, arguments, options, and indented content.
@@ -42,7 +42,7 @@ Images
 To include images (without a caption), use the ``image`` directive.
 You need to specify the path to the image and an alt text.
 
-.. code:: rst
+.. code-block:: rst
 
   .. image:: path/to/image.png
      :alt: Alternative text
@@ -64,7 +64,7 @@ Lists
 
 Nested lists must be separated from the parent list items by blank lines:
 
-.. code:: rst
+.. code-block:: rst
 
   - Item 1
   - Item 2
@@ -76,7 +76,7 @@ Nested lists must be separated from the parent list items by blank lines:
 
 Numbered lists can be autonumbered using the ``#`` sign.
 
-.. code:: rst
+.. code-block:: rst
 
   #. Item 1
   #. Item 2
@@ -96,9 +96,9 @@ or `code-block <https://www.sphinx-doc.org/en/master/usage/restructuredtext/dire
 
 By specifying the language, you enable pygments, which show syntax color coding for that code sample (see `here <https://pygments.org/languages/>`__ for a list of supported languages).
 
-.. code:: rst
+.. code-block:: rst
 
-   .. code:: language
+   .. code-block:: language
 
       code
 
@@ -108,15 +108,15 @@ Use ``[...]`` for anything omitted.
 
 For inlining small code snippets you can use the `code role <https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-code>`__ as in
 
-.. code:: rst
+.. code-block:: rst
 
    :code:`:term:`qube``
 
 You can add line numbers to code examples with the ``:linenos:`` parameter.
 
-.. code:: rst
+.. code-block:: rst
 
-    .. code:: python
+    .. code-block:: python
       :linenos:
 
        def hello_world():
@@ -125,50 +125,50 @@ You can add line numbers to code examples with the ``:linenos:`` parameter.
 
 You can have certain lines with the ``:emphasize-lines:`` parameter.
 
-.. code:: rst
+.. code-block:: rst
 
- .. code:: python
+ .. code-block:: python
    :emphasize-lines: 1,3,4
 
 
 
 For Python use ``python``.
 
-.. code:: rst
+.. code-block:: rst
 
-    .. code:: python
+    .. code-block:: python
 
       string_var = 'python'
 
 For Bash use ``bash``.
 
-.. code:: rst
+.. code-block:: rst
 
-    .. code:: bash
+    .. code-block:: bash
 
        #!/bin/bash
 
 For a terminal session use ``console``.
 
-.. code:: rst
+.. code-block:: rst
 
-    .. code:: console
+    .. code-block:: console
 
        $ echo "Hello"
 
 For text output use ``output``.
 
-.. code:: rst
+.. code-block:: rst
 
-    .. code:: output
+    .. code-block:: output
 
        some output
 
 For text use ``text``.
 
-.. code:: rst
+.. code-block:: rst
 
-    .. code:: text
+    .. code-block:: text
 
        some text
 
@@ -180,7 +180,7 @@ We adhere to the list tables directive by docutils as described `here <https://d
 
 A simple example would be:
 
-    .. code:: rst
+    .. code-block:: rst
 
         .. list-table::
            :widths: 15 10
@@ -204,7 +204,7 @@ They are typically styled as colored text boxes, usually accompanied by icons pr
 
 Here are examples of several types of alerts:
 
-.. code:: rst
+.. code-block:: rst
 
      .. hint::
         If you would like to avoid prefixing commands with [...]
@@ -235,7 +235,7 @@ These render as:
 
 You can also provide a `custom title to the admonition <https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-admonition>`_:
 
-.. code:: rst
+.. code-block:: rst
 
    .. admonition:: Did you know?
 
@@ -252,7 +252,7 @@ Glossary
 
 Most of the `glossaries <https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#glossary>`__ are present in :file:`/user/reference/glossary.rst`, like:
 
-.. code:: rst
+.. code-block:: rst
 
    .. glossary::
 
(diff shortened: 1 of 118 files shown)


unman (commented):
I am somewhat opposed to global changes like this. The results are difficult to review.
In this case I'd like to FIRST discuss the potential benefits to users and THEN move to implement if agreed.
What are the concrete advantages to users of this global change?

parulin:
I was lazy, tried to batch replace everything to see if it worked and getting no warning, I made this PR. Now, I'm reconsidering it and the few advantages I can find are limited.

I still think it would be good to favor `code-block` in the rst how-to and style guide, because some options don't work with `code` (i.e.: the examples of `:linenos:` and `:emphasize-lines:` are not functional). To me the main advantage is the `emphasize-lines` option. There are other options but I'm not sure they are helpful. There is no need to replace the `code` directive by the `code-block` one in the other parts of the documentation.

We can also remove the mention to `emphasize-lines` and correct the `linenos` example to work with `code` directive.

What do you think?

unman:
I thought that the code options were being  brought in line with the
codeblock options. Sure there is a PR for this.

parulin:
There are currently two different PRs on the Sphinx side: one keeping things as they are but changing the docs, the other one making `code` a real alias for `code-block`. So making such a global change in qubes-doc is obviously wrong right now.

I don't think we can expect them to be shipped in a new Sphinx version very soon, so I would like to just fix the 2 examples where the `code` directive is wrong in the meantime. That would be a very small PR. Is that okay?

developer/general/rst-documentation-style-guide.rst
```diff
@@ -28,10 +28,10 @@ provided by `Docutils <https://www.docutils.org/>`__ and extended by `Sphinx <ht
 Directives are used to insert non-paragraph content, such as images, tables, and code blocks.
 Example directives are:
 
-.. code:: rst
+.. code-block:: rst
 
    .. image::
-   .. code::
```
maiska, line 34: I would not like to impose `code-block` overall, `code` is still a valid directive and should not be swept away . So please if you want to use code-block that is fine, but mention also the alternative, meaning do not delete code or elaborate

maiska (changes requested):

developer/general/rst-documentation-style-guide.rst
parulin: I think I already made my point and I don't mind if you disagree. I just want to fix the examples in the style guide because they are wrong.

developer/general/rst-documentation-style-guide.rst
```diff
@@ -91,9 +92,13 @@ Item 3 will start at 1.
 Code blocks
 """""""""""
 
-When writing code blocks, use syntax highlighting within the `code <https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code>`__
```
maiska, line 94: Please refer back to the sphix doc. It mixes things up. If you want to add the docutils reference fine, but do not dismiss a valid reference of the sphinx docs :)

developer/general/rst-documentation-style-guide.rst
maiska: I beg your pardon? It seems correct to me :) 
https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code
& 
https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block

:)

parulin: Take a look at the sphinx issue mentioned in the description of this PR and/or test the examples provided in the docs. There was a PR I made that you fixed using this trick too :)

developer/general/rst-documentation-style-guide.rst
parulin: Same here, I disagree

developer/general/rst-documentation-style-guide.rst
maiska: yes, ok, i haven't read that. They should update their docs tho

developer/general/rst-documentation-style-guide.rst
maiska: Well I also disagree.
A little explanation can go a long way.
So would you like to add a summarization of the comment of the PR you linked:

```
IMO, this is a documentation problem:

While the sourcecode directive is an alias for ["code-block"](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block), code is a compatibility syntax for the upstream [Docutils "code" directive](https://docutils.sourceforge.io/docs/ref/rst/directives.html#code) with a different set of options:

instead of linenos and lineno-start, code uses number-lines,
emphasize-lines, caption, and dedent are not supported,
force is only supported by Sphinx, not by Docutils,
Authors that want to use the Sphinx-only options should use "code-block" or "sourcecode".
Authors that want their source to be compatible with Docutils should use "code" and only options supported by Docutils.
```
with respective referals. Or should i?

parulin: Okay, now it's not that I disagree... but I'm a bit lost about what you want here :)

I tried to do that in the warning, at line 98. I don't know what to do...

maiska: O the warning :) 

My bad, i concentrated on code vs code-block... OK, you did good!
I would like a little more elaboration as a discrimination - what functions with code and what with code-block to ease up user/contributors experience so that they should not delve right  into small gimmicks.
If you are willing to do so, if not I can enhance after merging, no prob

parulin: It seems to me that this PR has been misunderstood... So: what functions with `code`? Only one option which is also supported by code-block as two distinct options. And code-block adds emphasize-lines, caption, and dedent

So there is no use case where code is better than code-block. That's why I proposed to remove it from at least the style guide, to avoid issues for contributors confusing the two directives and their options.

maiska: yeah, let us wait till they fix this on the sphinx doc side and after that use `sourcecode` as an alternative without any warnings and elaboration.

parulin: The sphinx doc issue has been open since April but it was already there at least during the migration... Even if it was "fixed" both directives would work exactly like they do right now. The style guide tells the contributor to use either `code` or `code-block`, so I'm not sure if I understand your plan?

My proposition is to get rid of `code` or any other alias, so the contributor doesn't have to think about it and just uses `code-block`. It doesn't need something to be fixed on the sphinx side.

maiska: I saw that you wrote 2 days ago on the issue suggesting that they push the doc changes or am i mistaken?

Of course the doc should link to the sphinx documentaiton when the doc says: here this is a directive. 
And when people open the link and see three different versions and there is no explanation why only this and this -  is subpar imho.
So the correct approach is to explain it in the doc or wait for it to be fixed properly on the sphinx side, so one referal will suffice without the chain of explantion, than watching the PR when it gets merged, than removing the warning etc etc.

Sorry rn i am handling a little too much, so my attention span for such important but minor issues is a little restricted.

parulin: I will try to make a recap for us not to spend too much time on this. But 1st, let's say the Sphinx documentation is fixed, it will mention code-block/sourcecode on one side and `code` on the other. So, the contributor has 3 choices.

We are talking about the style *guide*, so I assume that, like links, headings, etc. the guide provides conventions, it's not a rST tutorial, there is not always an explanation. 

So for me, the best way to clarify the situation is to tell the contributor to just use `code-block`, without any explanation. This PR would become much easier to review (just remove or replace `code` by `code-block` in the "Code blocks" section.

If one still wants to use code or sourcecode directive after that, it's up to them, and if it fails (because of the current sphinx doc issue), it's up to them!

maiska: For me it will be refering to the sphinx documentation, saying you can either use this or that and in a consensus after that in the same document use one code-block for example, if you insist, idc, but not ruling out sourcecode (or code for that matter), and not punishing the future use of any of these since they are valid, if they are properly used :)

parulin: The thing is, I don't understand what kind of result you have in mind?

I have another proposal to just fix this, without spending more time: remove everything above the warning (warning included) and just make the example work with `code-block` instead of `code`? Just this:

```patch
diff --git i/developer/general/rst-documentation-style-guide.rst w/developer/general/rst-documentation-style-guide.rst
index 13edb358e..f634a8364 100644
--- i/developer/general/rst-documentation-style-guide.rst
+++ w/developer/general/rst-documentation-style-guide.rst
@@ -116,7 +116,7 @@ You can add line numbers to code examples with the ``:linenos:`` parameter.
 
 .. code:: rst
 
-    .. code:: python
+    .. code-block:: python
       :linenos:
 
        def hello_world():
@@ -127,7 +127,7 @@ You can have certain lines with the ``:emphasize-lines:`` parameter.
 
 .. code:: rst
 
- .. code:: python
+ .. code-block:: python
    :emphasize-lines: 1,3,4
```

parulin: Note that the second one is missing from the current PR (my mistake!)...

maiska: OK, I do not know how to express it otherwise ... :))

It is already time consuming writing all these texts for such a minor issue that actually can be solved pretty straightforward.
Part of this is on me, since I did not really pay attention at the start. If not yet clear, i would try:

1. Qubes OS documentaiton relies on sphinx. When there is a guide about that, it should refer back to the sphinx documentation for different reasons, f.ex. verification, as well as further reading/reference, etc. I think there is no misunderstanding about that. 
2. I would not like to constrain or restrict the usage of sphinx directives for the Qubes OS documenation (as I understand is what your prefered way of approaching this is, meaning only have code-block). Since there is a non-clarity on the sphinx side - one approach for Qubes OS doc would be to take all the info and point to the issue, and discuss why docutils and which sphinx directive etc etc., the other approach is to wait a little bit till the sphinx people fix that and fix then with a minor adjustment - point to code-block + sourcecode with an honorable mention of code with a caution, without having to deal with the extra info and github issues and so on. Here I assume that the fix will be merged soonish on the sphinx side :) 
3. I think the problem is Nr 2.  because for the code(-block) specific examples  with highlighting - yes - straight away they can be fixed, just the above is the hurdle atm

PR #1716, opened 2026-06-01
admin-api: document target=dom0 for tag-scoped qrexec policy rules

alex-schose:
`admin.*` calls are handled by `qubesd` in dom0. When a qrexec policy rule uses a `@tag:` selector to scope which VMs a caller may act on, omitting `target=dom0` routes the call to the tagged VM instead of dom0 — starting that VM in the process, even for read-only calls such as `admin.vm.firewall.Get` — and the call then fails, because `admin.*` is served by `qubesd` in dom0, not the target VM.

The default Admin API policy header already reminds maintainers to add `target=dom0` to allow/ask entries, but doesn't explain why, or what happens if it's omitted. This adds a short "Writing qrexec policy for Admin API calls" section that:

- describes the side effect (the tagged VM is started; the call fails)
- gives the correct pattern with `target=dom0`
- notes it applies to any `admin.vm.*` method whose destination is `vm`

Found while writing tag-scoped Admin API policies for an MCP-server isolation project (qubes-mcp); raised on qubes-devel: https://groups.google.com/g/qubes-devel/c/4NuSqL64DVE

### DIFF
--- a/developer/services/admin-api.rst
+++ b/developer/services/admin-api.rst
@@ -846,6 +846,32 @@ Example device serialization:
 
    1-1.1.1 manufacturer='unknown' self_identity='0000:0000::?******' serial='unknown' ident='1-1.1.1' product='Qubes' vendor='ITL' name='Some untrusted garbage' devclass='bus' backend_domain='vm' interfaces=' ******u03**01' _additional_info='' _date='06.12.23' parent_ident='1-1.1' parent_devclass='None'
 
+Writing qrexec policy for Admin API calls
+=========================================
+
+The default Admin API policy advises adding ``target=dom0`` to every
+allow/ask entry, but does not document the reason.
+
+All ``admin.*`` calls are handled by ``qubesd`` in dom0. When writing qrexec
+policy rules for these calls, a ``@tag:`` selector in the destination column
+must be paired with ``target=dom0``. For example:
+``admin.vm.firewall.Get * source-vm @tag:some-tag allow target=dom0``.
+
+Without ``target=dom0``, the policy matches the rule on the caller-named
+destination VM (validated against the tag constraint) but routes the call to
+that VM rather than to dom0. If the VM is not running, qrexec starts it as
+part of the routing attempt. This is an unintended side effect that occurs
+even with read-only calls such as ``admin.vm.firewall.Get``. The call does
+not succeed, because ``admin.*`` calls are served by ``qubesd`` in dom0, not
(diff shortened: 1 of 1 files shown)


developer/services/admin-api.rst
```diff
@@ -846,6 +846,32 @@ Example device serialization:
 
    1-1.1.1 manufacturer='unknown' self_identity='0000:0000::?******' serial='unknown' ident='1-1.1.1' product='Qubes' vendor='ITL' name='Some untrusted garbage' devclass='bus' backend_domain='vm' interfaces=' ******u03**01' _additional_info='' _date='06.12.23' parent_ident='1-1.1' parent_devclass='None'
 
+Writing qrexec policy for Admin API calls
+=========================================
+
+The default Admin API policy advises adding ``target=dom0`` to every
+allow/ask entry, but does not document the reason.
```
HW42: I think this paragraph doesn't add anything useful and can just be removed. (If the users looked at the policy they know already, if not it doesn't matter what the comment there says)

HW42 (changes requested):
The whole text is currently written in a way that suggest that `@tag:` rules are special here. The `target=dom0` is always needed not just for `@tag:` targets.

alex-schose:
Thanks, both fair.

- The section no longer frames `target=dom0` as `@tag:`-specific - it now opens by stating every `admin.*` allow/ask rule needs `target=dom0` regardless of the destination selector (VM name, `@anyvm`, or `@tag:`), matching the policy header's "any (allow/ask) entry".
- Dropped the opening paragraph about the default policy's comment.

Updated on the same branch.

alex-schose:
Following up on my 2026-06-09 comment — both review points were addressed in 9b33d9d6 the same day: the note now states that `target=dom0` is required for every `admin.*` allow/ask rule regardless of the destination selector, rather than framing it as `@tag:`-specific, and the paragraph about the default policy's comment is gone.

@HW42 the review is still standing as changes-requested against the earlier commit (1148f8a8), so the PR may be reading as unactioned in triage. Would you mind another look when you have a moment? Happy to rework the wording further if it's still not right.

unman (commented):
I would prefer NOT to use VM in place of 'qube'

unman (commented):
I see no reason to remove the reference to the default policy. I think it worthwhile to explicitly refer to it - a user who has not yet checked it will find it informative in the context of this page.

PR #1636, opened 2026-02-25
Fix: avoid indefinite Sphinx build freeze from unbounded git log call

suhr25:
**Summary**

This PR fixes a blocking issue in `_ext/last_edition.py` where a `git log` subprocess is executed without a timeout inside the `html_page_context` hook.

Since `html_page_context` runs once per page during Sphinx builds, the following code path is executed repeatedly:

```python
last_datetime_raw = subprocess.check_output(
    (
        "git",
        "--git-dir",
        git_dir,
        "log",
        "-1",
        "--pretty=format:%ci",
        "--",
        f"{pagename}{context['page_source_suffix']}",
    )
).decode()
```

Without a timeout, any delay/hang in `git` (repo lock, concurrent git operations, slow filesystem, etc.) can block the **entire** Sphinx build indefinitely. This is especially noticeable with `sphinx-autobuild`, where one stuck build prevents subsequent rebuilds.

**Fix**

The subprocess call is now bounded with a timeout and handles timeout failures gracefully:

```python
last_datetime_raw = subprocess.check_output(
    (...),
    timeout=5,
).decode()
except (subprocess.CalledProcessError, subprocess.TimeoutExpired):
    return
```

**Key improvements**

* Prevents indefinite blocking of Sphinx builds
* Ensures builds continue even if `git` is slow/unavailable
* Gracefully skips `last_edition` metadata for affected pages
* Minimal, localized change; normal fast-path remains unchanged

**Verification**

* Ran `sphinx-autobuild` locally and confirmed rebuilds continue under simulated `git` delays (e.g., temporary repo lock / slow response)
* Confirmed normal behavior is unchanged when `git log` executes quickly

PR #1724, opened 2026-06-20
Optional and resilient last_edition extension

parulin:
**Note:** it's not my work but I would like to keep #1636 opened

This PR originally fixed a blocking issue in `_ext/last_edition.py` where a `git log` subprocess is executed without a timeout inside the `html_page_context` hook.

Now, the `last_edition` extension is only loaded when a `QUBES_DOC_FULL_BUILD` environment variable is set.

The use of environment variable is now documented (including the onion meta tag).

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -37,8 +37,10 @@ extensions = [
   'sphinx_reredirects', # Manage redirects in the documentation
   'sphinxext.opengraph', # Add Open Graph meta tags for social media sharing
   'youtube_frame', # Embed YouTube videos
-  'last_edition',
 ]
+
+if os.environ.get('READTHEDOCS') or os.environ.get('QUBES_DOC_FULL_BUILD'):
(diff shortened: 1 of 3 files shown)


parulin:
I'm converting it to draft: the use of `--tag` with `sphinx-build` might more appropriate!

conf.py
```diff
@@ -37,9 +37,13 @@
...
   'sphinxext.opengraph', # Add Open Graph meta tags for social media sharing
   'youtube_frame', # Embed YouTube videos
-  'last_edition',
 ]
 
+if os.environ.get("READTHEDOCS") or os.environ.get(
+    "QUBES_DOC_FULL_BUILD", ""
+).lower() in ("true", "on", "1"):
+    extensions.append("last_edition")
+
```
parulin, line 46: I'm considering removing everything else except that part.

PR #1725, opened 2026-06-23
Add "How to shutdown a qube when idle" example section

parulin:


user/how-to-guides/how-to-shutdown-when-idle.rst
```diff
@@ -0,0 +1,16 @@
+================================
```
maiska: Hi, I think a new how to page is overhead here.
Perhaps you can add a section in this how to: 
https://doc.qubes-os.org/en/r4.3/user/how-to-guides/how-to-enable-a-service.html

maiska (changes requested):

user/how-to-guides/how-to-shutdown-when-idle.rst
parulin: > I think a new how to page is overhead here.

Thanks for the feedback, I agree but I wasn't able to figure out where to put this. Is it related to the navigation menu or in general? (to see if it fits in this issue: https://github.com/QubesOS/qubes-issues/issues/10264)

parulin: Since this how-to might not be a frequent task, another solution could be to "hide" it in a sublevel, with similar how-tos? I.e:

* How to organize your qubes
* [...]
* How to enable a qube service
* More how-tos (sublevel hidden from menu):
  * How to enter fullscreen mode
  * How to shutdown a qube when idle
  * How to use PCI devices
  * How to use optical discs
  * How to reinstall a template

maiska: Yes, I have the same problem - where to place it.

I think this can be cross referenced since there are 2 main topics - services and qube customization 

Perhaps this can be a new how to (or a new Level in How-Tos) around qubes in general - how can I customize my qube with examples. I can create a Qubes OS issue  which will be an enhancement and nice to have (meaning not urgent :) )

About the hierarchy: I get your point, is a good thing to do, but nice to have atm, but good to keep in mind actually for better organization :)

PR #1728, opened 2026-07-09
Clarify terminology around qube types, admin qubes

RandyTheOtter:
This pr aims to make the glossary better reflect the reality of terminology related to qube types, and to clarify the confusion around the terms "admin qube" and "AdminVM" and klass "AdminVM".

Changes:
- Add `admin client qube` term coined by @rustybird, see [forum post](https://forum.qubes-os.org/t/ssh-to-dom0/41427/28)  
- Replace "previously used" notes with more detailed explanation of possible synonyms; "previously used" is simply not true as many listed terms are still in use.

### DIFF
--- a/user/reference/glossary.rst
+++ b/user/reference/glossary.rst
@@ -33,12 +33,29 @@ Qube types and properties
 
 .. glossary::
 
+   admin client qube
+      .. image:: /attachment/doc/adminvm-black.svg
+         :width: 24px
+         :align: left
+
+      A :term:`qube` used for administering Qubes OS, running admin client software. In the default install, the only admin client qube is dom0, running `qubes-core-admin-client <https://github.com/QubesOS/qubes-core-admin-client>`__ software.
+
+      “admin client qube” is the preferred term, but it can also be referred to as "AdminVM" or "management VM" in historic blog and forum posts.
+
+      .. note:: Be careful to avoid confusion with these terms. "AdminVM" when referring to admin client qubes is not related to the ``AdminVM`` qube :term:`class`. Similarly, "management VM" and "management DVM" or "default-mgmt-dvm" are not the same, see :term:`management qube`.
+
    admin qube
       .. image:: /attachment/doc/adminvm-black.svg
          :width: 24px
          :align: left
 
-      A type of :term:`qube` used for administering Qubes OS. In the default install, the only admin qube is :term:`dom0`.
+      A :term:`qube` used for administering Qubes OS, running admin server software. In the default install, the only admin qube is dom0, running `qubes-core-admin <https://github.com/QubesOS/qubes-core-admin>`__ and `qrexec <https://github.com/QubesOS/qubes-core-qrexec>`__ daemons.
+
+      "admin qube" is the preferred term, but it can also be referred to as:
+
+      - ``AdminVM`` in historic blog and forum posts, but also in modern technical context when referring to qube :term:`class` used by ``qvm`` tools.
+      - ``@adminvm`` in qrexec policies.
+      - ``dom0`` because in current implementation of Qubes OS this is the only admin qube, and it is hard-coded to be that.
 
    app qube
       .. image:: /attachment/doc/appvm-red.svg
@@ -49,8 +66,10 @@ Qube types and properties
 
       An *app qube* does not have a root file system of its own. The qube borrows its root file system from its  :term:`template`, and only owns its own home directory and user files. (In Linux-based qubes, these are the :file:`/home` and :file:`/usr/local` directories.)
 
-      - Previously known as: ``AppVM``, ``TemplateBasedVM``.
+      "app qube" is the preferred term, but it can also be referred to as:
 
+      - ``AppVM`` in historic blog and forum posts and documentation, but also in modern technical context when referring to qube :term:`class` used by qvm tools, qrexec policies, configuration management, etc.
+      - ``TemplateBasedVM`` in historic blog and forum posts.
 
    disposable
       .. image:: /attachment/doc/dispvm-red.svg
@@ -67,7 +86,11 @@ Qube types and properties
 
       See also :term:`named disposable`.
 
-      - Previously known as: ``DisposableVM``, ``DispVM``.
+      "disposable" is the preferred term, but it can also be referred to as:
+      
+      - ``DispVM`` in historic blog and forum posts and documentation, but also in modern technical context when referring to qube :term:`class` used by qvm tools, qrexec policies, configuration management, etc.
+      - ``@dispvm`` for new disposables in certain qrexec policies and admin API.
+      - ``DisposableVM`` in historic blog and forum posts.
 
    standalone
       .. image:: /attachment/doc/standalonevm-red.svg
@@ -78,7 +101,7 @@ Qube types and properties
 
       See :doc:`/user/advanced-topics/standalones-and-hvms`.
 
-      - Previously known as: ``StandaloneVM``.
+      "standalone" is the preferred term, but it can also be referred to as ``StandaloneVM`` in historic blog and forum posts and documentation, but also in modern technical context when referring to qube :term:`class` used by qvm tools, qrexec policies, configuration management, etc.
 
    template
       .. image:: /attachment/doc/templatevm-red.svg
@@ -95,7 +118,7 @@ Qube types and properties
 
       - Regular templates cannot function as :term:`disposable template`. (Disposable templates must be app qubes).
 
-      - Previously known as: ``TemplateVM``.
+      "template" is the preferred term, but it can also be referred to as ``TemplateVM`` in historic blog and forum posts and documentation, but also in modern technical context when referring to qube :term:`class` used by qvm tools, qrexec policies, configuration management, etc.
 
    disposable template
       .. image:: /attachment/doc/templatevm-red.svg
@@ -112,7 +135,7 @@ Qube types and properties
 
       - Unlike :term:`disposables <disposable>`, disposable templates behave like normal :term:`app qubes <app qube>` in terms of persistence - their contents survive reboots. Thus, you can configure your disposable template to have e.g. browser extensions or configuration that will be present in every disposable qube based on it.
 
-      - Previously known as: ``DisposableVM Template``, ``DVM Template``, ``DVM``.
+      "disposable template" is the preferred term, but it can also be referred to as ``DisposableVM Template``, ``DVM Template`` and ``DVM`` in historic blog and forum posts and documentation.
 
    named disposable
       .. image:: /attachment/doc/dispvm-red.svg
@@ -126,7 +149,10 @@ Qube types and properties
       - Technical note: Named disposables are useful for certain :term:`service qube`\ s, where the combination of persistent device assignment and ephemeral qube state is desirable.
 
    management qube
-      A :term:`qube` used for automated management of a Qubes OS installation via :doc:`/user/advanced-topics/salt`.
+      A :term:`qube` used for automated management of a Qubes OS installation via :doc:`/user/advanced-topics/salt`. Sometimes referred to as "default-mgmt-dvm".
+
+   class
+      Property describing qube type from the system's perspective, sometimes spelled ``klass`` due to a technicality. Can be either of ``AdminVM``, ``StandaloneVM``, ``TemplateVM``, ``AppVM``, and ``DispVM``.
 
    net qube
       A property of any qube that tells the system which (if any) :term:`service qube` to use to connect to the network.
--- a/user/reference/glossary.rst
+++ b/user/reference/glossary.rst
@@ -43,7 +43,6 @@ Qube types and properties
       "admin qube" is the preferred term, but it can also be referred to as:
 
       - ``AdminVM`` in outdated sources, but also in modern technical context when referring to qube :term:`class`.
-      - ``@adminvm`` in qrexec policies.
       - ``management VM`` in outdated sources.
 
       .. note:: Be careful to avoid confusion with these terms. "AdminVM" when referring to admin qubes does not necessarily imply ``AdminVM`` qube :term:`class` - qubes of other classes can perform some administrative functions too, see :doc:`admin API </developer/services/admin-api>`.
--- a/user/reference/glossary.rst
+++ b/user/reference/glossary.rst
@@ -33,29 +33,20 @@ Qube types and properties
 
 .. glossary::
 
-   admin client qube
-      .. image:: /attachment/doc/adminvm-black.svg
-         :width: 24px
-         :align: left
-
-      A :term:`qube` used for administering Qubes OS, running admin client software. In the default install, the only admin client qube is dom0, running `qubes-core-admin-client <https://github.com/QubesOS/qubes-core-admin-client>`__ software.
-
-      “admin client qube” is the preferred term, but it can also be referred to as "AdminVM" or "management VM" in historic blog and forum posts.
-
-      .. note:: Be careful to avoid confusion with these terms. "AdminVM" when referring to admin client qubes is not related to the ``AdminVM`` qube :term:`class`. Similarly, "management VM" and "management DVM" or "default-mgmt-dvm" are not the same, see :term:`management qube`.
-
    admin qube
       .. image:: /attachment/doc/adminvm-black.svg
          :width: 24px
          :align: left
 
-      A :term:`qube` used for administering Qubes OS, running admin server software. In the default install, the only admin qube is dom0, running `qubes-core-admin <https://github.com/QubesOS/qubes-core-admin>`__ and `qrexec <https://github.com/QubesOS/qubes-core-qrexec>`__ daemons.
+      A :term:`qube` used for administering Qubes OS, running admin software. In the default install, the only admin qube is dom0, running :doc:`qubes-core-admin <core-admin:index>` and :doc:`qrexec </developer/services/qrexec>` daemons. Dom0 is the only admin qube by default, but more can be added by leveraging :doc:`admin API </developer/services/admin-api>`.
 
       "admin qube" is the preferred term, but it can also be referred to as:
 
-      - ``AdminVM`` in historic blog and forum posts, but also in modern technical context when referring to qube :term:`class` used by ``qvm`` tools.
+      - ``AdminVM`` in outdated sources, but also in modern technical context when referring to qube :term:`class`.
       - ``@adminvm`` in qrexec policies.
-      - ``dom0`` because in current implementation of Qubes OS this is the only admin qube, and it is hard-coded to be that.
+      - ``management VM`` in outdated sources.
+
+      .. note:: Be careful to avoid confusion with these terms. "AdminVM" when referring to admin qubes does not necessarily imply ``AdminVM`` qube :term:`class` - qubes of other classes can perform some administrative functions too, see :doc:`admin API </developer/services/admin-api>`.
 
    app qube
       .. image:: /attachment/doc/appvm-red.svg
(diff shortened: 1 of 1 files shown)


ben-grande:
Please send this patch to qubes-devel mailing list so we can discuss with more visibility.

user/reference/glossary.rst
```diff
@@ -33,12 +33,29 @@ Qube types and properties
...
 .. glossary::
 
+   admin client qube
+      .. image:: /attachment/doc/adminvm-black.svg
+         :width: 24px
+         :align: left
+
+      A :term:`qube` used for administering Qubes OS, running admin client software. In the default install, the only admin client qube is dom0, running `qubes-core-admin-client <https://github.com/QubesOS/qubes-core-admin-client>`__ software.
+
+      “admin client qube” is the preferred term, but it can also be referred to as "AdminVM" or "management VM" in historic blog and forum posts.
```
ben-grande: Please remove this. I think this will cause a lot of confusion.

user/reference/glossary.rst
```diff
@@ -33,12 +33,29 @@ Qube types and properties
...
+
+      .. note:: Be careful to avoid confusion with these terms. "AdminVM" when referring to admin client qubes is not related to the ``AdminVM`` qube :term:`class`. Similarly, "management VM" and "management DVM" or "default-mgmt-dvm" are not the same, see :term:`management qube`.
```
ben-grande: Same comment as above, to avoid confusion, I wouldn't mention it.

user/reference/glossary.rst
```diff
@@ -33,12 +33,29 @@ Qube types and properties
 
 .. glossary::
 
+   admin client qube
+      .. image:: /attachment/doc/adminvm-black.svg
+         :width: 24px
+         :align: left
+
+      A :term:`qube` used for administering Qubes OS, running admin client software. In the default install, the only admin client qube is dom0, running `qubes-core-admin-client <https://github.com/QubesOS/qubes-core-admin-client>`__ software.
```
ben-grande: Link to the RTD docs: https://dev.qubes-os.org/projects/core-admin-client/en/latest/. Same thing on other places.

user/reference/glossary.rst
```diff
@@ -33,12 +33,29 @@ Qube types and properties
...
+
+      .. note:: Be careful to avoid confusion with these terms. "AdminVM" when referring to admin client qubes is not related to the ``AdminVM`` qube :term:`class`. Similarly, "management VM" and "management DVM" or "default-mgmt-dvm" are not the same, see :term:`management qube`.
+
    admin qube
       .. image:: /attachment/doc/adminvm-black.svg
          :width: 24px
          :align: left
 
-      A type of :term:`qube` used for administering Qubes OS. In the default install, the only admin qube is :term:`dom0`.
+      A :term:`qube` used for administering Qubes OS, running admin server software. In the default install, the only admin qube is dom0, running `qubes-core-admin <https://github.com/QubesOS/qubes-core-admin>`__ and `qrexec <https://github.com/QubesOS/qubes-core-qrexec>`__ daemons.
```
ben-grande: https://github.com/QubesOS/qubes-doc/pull/1728/changes#r3568978680

user/reference/glossary.rst
```diff
@@ -33,12 +33,29 @@ Qube types and properties
...
+
+      "admin qube" is the preferred term, but it can also be referred to as:
```
ben-grande: Contrary to "admin client qube" above, I think this explanation is much better because it tells "what it is", not "what it isn't".

user/reference/glossary.rst
```diff
@@ -33,12 +33,29 @@ Qube types and properties
...
+
+      - ``AdminVM`` in historic blog and forum posts, but also in modern technical context when referring to qube :term:`class` used by ``qvm`` tools.
```
ben-grande: I would not mention "forum posts", it is not a reliable source. I would change this line to:

```rst
``AdminVM`` when referring to qube :term:`class`
```

I dropped also "historic blog" because they will never be edited also.

user/reference/glossary.rst
```diff
@@ -49,8 +66,10 @@ Qube types and properties
 
       An *app qube* does not have a root file system of its own. The qube borrows its root file system from its  :term:`template`, and only owns its own home directory and user files. (In Linux-based qubes, these are the :file:`/home` and :file:`/usr/local` directories.)
 
-      - Previously known as: ``AppVM``, ``TemplateBasedVM``.
+      "app qube" is the preferred term, but it can also be referred to as:
 
+      - ``AppVM`` in historic blog and forum posts and documentation, but also in modern technical context when referring to qube :term:`class` used by qvm tools, qrexec policies, configuration management, etc.
+      - ``TemplateBasedVM`` in historic blog and forum posts.
```
ben-grande: It should never be referred this way. I assume you included "historic blog" because you found a mention there, still, this should be mentioned here. I'd say we should totally forget this word.

user/reference/glossary.rst
```diff
@@ -49,8 +66,10 @@ Qube types and properties
 
       An *app qube* does not have a root file system of its own. The qube borrows its root file system from its  :term:`template`, and only owns its own home directory and user files. (In Linux-based qubes, these are the :file:`/home` and :file:`/usr/local` directories.)
 
-      - Previously known as: ``AppVM``, ``TemplateBasedVM``.
+      "app qube" is the preferred term, but it can also be referred to as:
 
+      - ``AppVM`` in historic blog and forum posts and documentation, but also in modern technical context when referring to qube :term:`class` used by qvm tools, qrexec policies, configuration management, etc.
```
ben-grande: Same comment as https://github.com/QubesOS/qubes-doc/pull/1728/changes#r3569003631, I would just keep the mention of "qube class" to keep the information here short and concise.

user/reference/glossary.rst
```diff
@@ -67,7 +86,11 @@ Qube types and properties
 
       See also :term:`named disposable`.
 
-      - Previously known as: ``DisposableVM``, ``DispVM``.
+      "disposable" is the preferred term, but it can also be referred to as:
+      
+      - ``DispVM`` in historic blog and forum posts and documentation, but also in modern technical context when referring to qube :term:`class` used by qvm tools, qrexec policies, configuration management, etc.
+      - ``@dispvm`` for new disposables in certain qrexec policies and admin API.
+      - ``DisposableVM`` in historic blog and forum posts.
```
ben-grande: Same as https://github.com/QubesOS/qubes-doc/pull/1728/changes#r3569012551, drop it.

user/reference/glossary.rst
```diff
@@ -67,7 +86,11 @@ Qube types and properties
 
       See also :term:`named disposable`.
 
-      - Previously known as: ``DisposableVM``, ``DispVM``.
+      "disposable" is the preferred term, but it can also be referred to as:
+      
+      - ``DispVM`` in historic blog and forum posts and documentation, but also in modern technical context when referring to qube :term:`class` used by qvm tools, qrexec policies, configuration management, etc.
+      - ``@dispvm`` for new disposables in certain qrexec policies and admin API.
```
ben-grande: This is only for `qrexec`, not for the Admin API.

user/reference/glossary.rst
```diff
@@ -67,7 +86,11 @@ Qube types and properties
 
       See also :term:`named disposable`.
 
-      - Previously known as: ``DisposableVM``, ``DispVM``.
+      "disposable" is the preferred term, but it can also be referred to as:
+      
+      - ``DispVM`` in historic blog and forum posts and documentation, but also in modern technical context when referring to qube :term:`class` used by qvm tools, qrexec policies, configuration management, etc.
```
ben-grande: Same as https://github.com/QubesOS/qubes-doc/pull/1728/changes#r3569019629.

user/reference/glossary.rst
```diff
@@ -78,7 +101,7 @@ Qube types and properties
 
       See :doc:`/user/advanced-topics/standalones-and-hvms`.
 
-      - Previously known as: ``StandaloneVM``.
+      "standalone" is the preferred term, but it can also be referred to as ``StandaloneVM`` in historic blog and forum posts and documentation, but also in modern technical context when referring to qube :term:`class` used by qvm tools, qrexec policies, configuration management, etc.
```
ben-grande: https://github.com/QubesOS/qubes-doc/pull/1728/changes#r3569019629

user/reference/glossary.rst
```diff
@@ -95,7 +118,7 @@ Qube types and properties
 
       - Regular templates cannot function as :term:`disposable template`. (Disposable templates must be app qubes).
 
-      - Previously known as: ``TemplateVM``.
+      "template" is the preferred term, but it can also be referred to as ``TemplateVM`` in historic blog and forum posts and documentation, but also in modern technical context when referring to qube :term:`class` used by qvm tools, qrexec policies, configuration management, etc.
```
ben-grande: https://github.com/QubesOS/qubes-doc/pull/1728/changes#r3569019629

user/reference/glossary.rst
```diff
@@ -112,7 +135,7 @@ Qube types and properties
 
       - Unlike :term:`disposables <disposable>`, disposable templates behave like normal :term:`app qubes <app qube>` in terms of persistence - their contents survive reboots. Thus, you can configure your disposable template to have e.g. browser extensions or configuration that will be present in every disposable qube based on it.
 
-      - Previously known as: ``DisposableVM Template``, ``DVM Template``, ``DVM``.
+      "disposable template" is the preferred term, but it can also be referred to as ``DisposableVM Template``, ``DVM Template`` and ``DVM`` in historic blog and forum posts and documentation.
```
ben-grande: I think that the "Previously known as" is much better. I don't have to mention where it is used, just that it shouldn't be used.

user/reference/glossary.rst
```diff
@@ -126,7 +149,10 @@ Qube types and properties
       - Technical note: Named disposables are useful for certain :term:`service qube`\ s, where the combination of persistent device assignment and ephemeral qube state is desirable.
 
    management qube
-      A :term:`qube` used for automated management of a Qubes OS installation via :doc:`/user/advanced-topics/salt`.
+      A :term:`qube` used for automated management of a Qubes OS installation via :doc:`/user/advanced-topics/salt`. Sometimes referred to as "default-mgmt-dvm".
+
+   class
+      Property describing qube type from the system's perspective, sometimes spelled ``klass`` due to a technicality. Can be either of ``AdminVM``, ``StandaloneVM``, ``TemplateVM``, ``AppVM``, and ``DispVM``.
```
ben-grande: It is missing `RemoteVM`. Check out: `qvm-create --help-classes`.

user/reference/glossary.rst
```diff
@@ -33,12 +33,29 @@ Qube types and properties
 
 .. glossary::
 
+   admin client qube
```
ben-grande: I'd put this section below "admin" qube section, so we start describing the server, than the client of said server.

user/reference/glossary.rst
ben-grande: Ok, I might prefer the following:

```rst
"app qube" is the preferred term for pretty text, but it can also be referred to as ``AppVM`` when referring to qube :term:`class`.

- Previously known as: ``TemplateBasedVM``.
```

This way we don't say that `AppVM` is outdated, we don't add large text explaining when it can be used, and we don't explain when `TemplateBasedVM` could be used (it's outdated, doesn't matter).

ben-grande (changes requested):

ben-grande:
> Please send this patch to qubes-devel mailing list so we can discuss with more visibility.

By reading the forum, I think the issue unman had was "admin client qube", this is what should be discussed on the qubes-devel mailing list. We are already using "admin qube", so I'd like to say that I find reasonable to use "admin client qube".

unman:
On Mon, Jul 13, 2026 at 01:28:15AM -0700, ben-grande wrote:
> By reading the forum, I think the issue unman had was "admin client qube", this is what should be discussed on the qubes-devel mailing list. We are already using "admin qube", so I'd like to say that I find reasonable to use "admin client qube".
I think I have made it clear that I do **not**. Discussion on the
mailing list will not, I think, change my mind on this.

rustybird:
Yeah [I agree](https://forum.qubes-os.org/t/terminology-admin-qube-adminvm/42462/30) the term "admin client qube" is a no-go:

> TBH if I had seen it coming that the term “admin client qube” would turn out so controversial, I would never have suggested it. Probably the worst outcome would be only adding it as one more of several terms people continue using to refer to the same thing.

user/reference/glossary.rst
RandyTheOtter: I still think this note should stay as, unlike others, AdminVM function does not imply AdminVM class. I will remove "management VM" part and try to make it more clear.

user/reference/glossary.rst
RandyTheOtter: How did you come to the last conclusion?

One of the reasons for these notes to exist is to clarify terms used in the old 
posts, similar to "archaic" entries in a dictionary. Otherwise, if we allow ourselves to edit the old posts using new terminology, these notes are no longer needed.

I replace "historic" with "outdated" to make deprecation of the terms more obvious.

user/reference/glossary.rst
RandyTheOtter: It is included in the current glossary. It is used in QSBs.

user/reference/glossary.rst
RandyTheOtter: It is important to inform that "DispVM" in outdated sources does not refer to the class.

user/reference/glossary.rst
RandyTheOtter: I follow pattern similar to the app qube for now, mainly because this term is still present in some docs. Tell me if you want it removed anyway.

user/reference/glossary.rst
RandyTheOtter: Yeah, this one might be better as is, I changed it to align with the others.

user/reference/glossary.rst
RandyTheOtter: Resolved as "admin client qube" is removed.

user/reference/glossary.rst
ben-grande: > AdminVM function does not imply AdminVM class

It does, if it was used another way, it is wrong.

user/reference/glossary.rst
ben-grande: Because the paragraph is not teaching when to use it or not. It is telling where you can possibly find the term. I'd very much prefer the former. I think "outdated" is better, yes.

user/reference/glossary.rst
```diff
@@ -38,7 +38,15 @@ Qube types and properties
          :width: 24px
          :align: left
 
-      A type of :term:`qube` used for administering Qubes OS. In the default install, the only admin qube is :term:`dom0`.
+      A :term:`qube` used for administering Qubes OS, running admin software. In the default install, the only admin qube is dom0, running :doc:`qubes-core-admin <core-admin:index>` and :doc:`qrexec </developer/services/qrexec>` daemons. Dom0 is the only admin qube by default, but more can be added by leveraging :doc:`admin API </developer/services/admin-api>`.
+
+      "admin qube" is the preferred term, but it can also be referred to as:
+
+      - ``AdminVM`` in outdated sources, but also in modern technical context when referring to qube :term:`class`.
+      - ``@adminvm`` in qrexec policies.
```
rustybird: `@adminvm` only means `dom0` though. It can never refer to a non-dom0 qube.

https://github.com/QubesOS/qubes-core-qrexec/blob/main/qrexec/policy/parser.py

unman:
The glossary and documentation is primarily user facing, not technical.
It's a mistake to try to extend it. I dont think there is need for
reference to policy at all.

Also, i find the wording here confusing - 
It's not the "preferred term", it's what we use. In older posts and news
you can find ManagementVM and AdminVM - that is how we used to refer to
it. Now we use "admin qube"

user/reference/glossary.rst
RandyTheOtter: It is rather that we don't use "AdminVM" to describe function of a qube in modern terminology. When used in such a way, however, "AdminVM" does not imply class.

I don't think that "wrong" is a good word for old documents. It is correct, but meanings can be over the place. That is what this note aims to help with.

New iteration of the note states that "AdminVM" can only refer to qube role in outdated sources.

RandyTheOtter:
> The glossary and documentation is primarily user facing, not technical. It's a mistake to try to extend it. I dont think there is need for reference to policy at all.

Policies are documented in the user section:
https://github.com/QubesOS/qubes-doc/blob/main/user/how-to-guides/how-to-edit-a-policy.rst

> Also, i find the wording here confusing - It's not the "preferred term", it's what we use. In older posts and news you can find ManagementVM and AdminVM - that is how we used to refer to it. Now we use "admin qube"

I have considered "modern" as an alternative, but to me it seems like it doesn't communicate the intent as well. I changed it to "Previously known as" for now.

More importantly, "what we use" is not the best target for a glossary. Unlike most other documents, it is documenting the language, not the system. You may argue that users should refer to an older iteration of a glossary for outdated terms, but that involves browsing documentation repository, an activity not fit for an end user. On the other hand, it is not uncommon for dictionaries to include archaic entries.

Thus documenting outdated terminology well and stating what term is preferred makes sense.

user/reference/glossary.rst
```diff
@@ -67,7 +72,11 @@ Qube types and properties
 
       See also :term:`named disposable`.
 
-      - Previously known as: ``DisposableVM``, ``DispVM``.
+      "disposable" is the preferred term, but it can also be referred to as:
+      
+      - ``DispVM`` in outdated sources, but also in modern technical context when referring to qube :term:`class`.
+      - ``@dispvm`` for new disposables in certain qrexec policies.
+      - ``DisposableVM`` in outdated sources.
```
ben-grande, line 79: I'd like to suggest the following:

```rst
It can be referred by different terms depending on context

- pretty name: "disposable"
- qube :term:`class`: ``DispVM``

Previously known as:

- pretty name: ``DisposableVM``
```

- I don't think we should include the Qrexec term here at all, because well, it is missing `@type:DispVM` for example, and I think that would be too much repetition. The Qrexec syntax should be explained in the Qrexec page and not in the glossary.
- I shortened the text a lot and I think this is enough and this formula should be repeated on the other terms.

user/reference/glossary.rst
```diff
@@ -78,7 +87,7 @@ Qube types and properties
 
       See :doc:`/user/advanced-topics/standalones-and-hvms`.
 
-      - Previously known as: ``StandaloneVM``.
+      "standalone" is the preferred term, but it can also be referred to as ``StandaloneVM`` in outdated sources and in modern technical context when referring to qube :term:`class`.
```
ben-grande, line 90: Same as https://github.com/QubesOS/qubes-doc/pull/1728/changes#r3594238569.

```rst
It can be referred by different terms depending on context

- pretty name: "standalone"
- qube :term:`class`: ``StandaloneVM``
```

user/reference/glossary.rst
```diff
@@ -95,7 +104,7 @@ Qube types and properties
 
       - Regular templates cannot function as :term:`disposable template`. (Disposable templates must be app qubes).
 
-      - Previously known as: ``TemplateVM``.
+      "template" is the preferred term, but it can also be referred to as ``TemplateVM`` in outdated sources and in modern technical context when referring to qube :term:`class`.
```
ben-grande, line 107: Same as https://github.com/QubesOS/qubes-doc/pull/1728/changes#r3594238569.

```rst
It can be referred by different terms depending on context

- pretty name: "template"
- qube :term:`class`: ``TemplateVM``
```

user/reference/glossary.rst
```diff
@@ -126,7 +135,10 @@ Qube types and properties
       - Technical note: Named disposables are useful for certain :term:`service qube`\ s, where the combination of persistent device assignment and ephemeral qube state is desirable.
 
    management qube
-      A :term:`qube` used for automated management of a Qubes OS installation via :doc:`/user/advanced-topics/salt`.
+      A :term:`qube` used for automated management of a Qubes OS installation via :doc:`/user/advanced-topics/salt`. Sometimes referred to as "default-mgmt-dvm".
```
ben-grande, line 138: Suggestion:

```rst
By default, this qube is ``default-mgmt-dvm``.
```

PR #1732, opened 2026-07-15
Ease the use of intersphinx: use local inventories and remove prefixes

parulin:
* **Add a way to look for local intersphinx inventory**
  
  When editing references between qubes-doc and qubes-core-* projects, it's better to use inventory from local builds.

* **Allow intersphinx references without prefixes**
  
  We only use qubes projects, so it removes the need to prefix cross-references.

**TODO:** document environment variables


parulin:
It would be better to just use `sphinx-build --define` and totally drop this PR (but see: https://github.com/sphinx-doc/sphinx/issues/14539) or to use something like `--tag intersphinx-core-admin-target=...`

parulin:
This is a wrong approach: if someone wants to do this, they just have to write another configuration file and build the local docs with that.

PR #1733, opened 2026-07-15
Update references to qvm-service and qvm-features

parulin:
Now that https://github.com/QubesOS/qubes-core-admin-client/pull/479 is merged, I updated some of the references to `qvm-service` and `qvm-features` tools. I also made some edits to the "Custom persist section" of `bind-dirs.rst`

Cherry-picked from #1732: "Allow intersphinx references without prefixes". We only use qubes projects, so it removes the need to prefix cross-references.

### DIFF
--- a/developer/releases/4_3/release-notes.rst
+++ b/developer/releases/4_3/release-notes.rst
@@ -104,7 +104,7 @@ UI/UX
 - Centralized Tray Notifications
   (`#889 <https://github.com/QubesOS/qubes-issues/issues/889>`__).
 
-- Options to launch root terminal or `console terminal <https://dev.qubes-os.org/projects/core-admin-client/en/latest/manpages/qvm-features.html#expert-mode>`__ from :program:`Qubes Domains widget` (`#9788 <https://github.com/QubesOS/qubes-issues/issues/9788>`__).
+- Options to launch root terminal or :option:`console terminal <qvm-features expert-mode>` from :program:`Qubes Domains widget` (`#9788 <https://github.com/QubesOS/qubes-issues/issues/9788>`__).
 
   Holding the shift key changes the :guilabel:`Run Terminal` command into a :guilabel:`Run Root Terminal` action.
 
@@ -182,14 +182,14 @@ Security features
 - Preventing unsafe practice of 3rd party template installation with rpm/dnf
   (`#9943 <https://github.com/QubesOS/qubes-issues/issues/9943>`__).
 
-- Ability to `prohibit start <https://dev.qubes-os.org/projects/core-admin-client/en/latest/manpages/qvm-features.html#prohibit-start>`__ of specific qubes
+- Ability to :option:`prohibit start <qvm-features prohibit-start>` of specific qubes
   (`#9622 <https://github.com/QubesOS/qubes-issues/issues/9622>`__).
 
 - UUID support for qubes and support for addressing them by UUID in policies
   (`#8862 <https://github.com/QubesOS/qubes-issues/issues/8862>`__ &
   `#8510 <https://github.com/QubesOS/qubes-issues/issues/8510>`__).
 
-- `Custom persist feature <https://dev.qubes-os.org/projects/core-admin-client/en/latest/manpages/qvm-features.html#custom-persist>`__ to avoid unwanted data to persist as much as possible
+- :option:`Custom persist feature <qvm-features custom-persist.*>` to avoid unwanted data to persist as much as possible
   (`#1006 <https://github.com/QubesOS/qubes-issues/issues/1006>`__).
 
 Anonymity improvements
@@ -276,7 +276,7 @@ Other
 - Automatically clean up `QubesIncoming` directory if empty
   (`#8307 <https://github.com/QubesOS/qubes-issues/issues/8307>`__).
 
-- `vm-config.* <https://dev.qubes-os.org/projects/core-admin-client/en/latest/manpages/qvm-features.html#vm-config>`__ features to pass external configuration to inside the qube (`#9837 <https://github.com/QubesOS/qubes-issues/issues/9837>`__).
+- :option:`vm-config.* <qvm-features vm-config.*>` features to pass external configuration to inside the qube (`#9837 <https://github.com/QubesOS/qubes-issues/issues/9837>`__).
 
 - Admin API for reading/writing denied device-interface list
   (`#9674 <https://github.com/QubesOS/qubes-issues/issues/9674>`__).
--- a/developer/releases/4_3/release-notes.rst
+++ b/developer/releases/4_3/release-notes.rst
@@ -33,7 +33,7 @@ Major features and improvements since Qubes 4.2
 - Device “self-identity oriented” assignment (a.k.a New Devices API)
   (`#9325 <https://github.com/QubesOS/qubes-issues/issues/9325>`__).
 
-  **See:** :option:`core-admin-client:qvm-device --assignments`, ``qvm-device assign`` and ``qvm-device unassign``
+  **See:** :option:`qvm-device --assignments`, :doc:`qvm-device assign <manpages/qvm-device>` and :doc:`qvm-device unassign <manpages/qvm-device>`
 
 - :doc:`/user/templates/windows/qubes-windows-tools` reintroduction with improved features
   (`#1861 <https://github.com/QubesOS/qubes-issues/issues/1861>`__).
@@ -219,7 +219,7 @@ Performance optimizations
   (`#9362 <https://github.com/QubesOS/qubes-issues/issues/9362>`__).
 
 - Minimal state qubes to make NetVM and USBVM to consume as little RAM as
-  possible.
+  possible, see: :option:`qvm-service minimal-netvm` and :option:`qvm-service minimal-usbvm`
 
 Updating & upgrading
 --------------------
--- a/user/advanced-topics/bind-dirs.rst
+++ b/user/advanced-topics/bind-dirs.rst
@@ -87,7 +87,7 @@ Other Configuration Folders
 
 - ``/etc/qubes-bind-dirs.d`` (intermediate priority, for template wide configuration)
 
-- ``/rw/config/qubes-bind-dirs.d`` (highest priority, for per VM configuration)
+- ``/rw/config/qubes-bind-dirs.d`` (highest priority, for per qube configuration)
 
 
 
@@ -133,11 +133,13 @@ How to remove binds from bind-dirs.sh?
 
 (Editing ``/usr/lib/qubes-bind-dirs.d/40_qubes-whonix.conf`` directly is strongly discouraged, since such changes get lost when that file is changed in the package on upgrades.)
 
+.. _custom-persist:
+
 Custom persist feature
 ----------------------
 
 
-Custom persist is an optional advanced feature allowing the creation of minimal state AppVM. The purpose of such an AppVM is to avoid unwanted data to persist as much as possible by disabling the ability to configure persistence from the VM itself. When enabled, the following happens:
+Custom persist is an optional advanced feature allowing the creation of minimal state qube. The purpose of such an qube is to avoid unwanted data to persist as much as possible by disabling the ability to configure persistence from the qube itself. When enabled, the following happens:
 
 - ``/rw/config/rc.local`` is no longer executed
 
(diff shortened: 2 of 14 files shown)


maiska:
there are other occurences where a change is needed, i will list them here. if you are willing, you can change them, if not i can do this also after merging this PR:

```
- add link to developer/debugging/vm-interface.rst:156:- ``memory/meminfo`` (**xenstore**) - used memory (updated by qubes-meminfo-writer), input information for qmemman;
- remove link to qvm-service:
developer/debugging/vm-interface.rst:288:- ``systemctl start qubes-update-check.timer`` (and similarly stop) - called when enabling/disabling update checking in a given VM (``qubes-update-check`` :doc:`qvm-service </user/advanced-topics/qubes-service>`)

- add to user/security-in-qubes/firewall.rst:90:   The firewall rules described above only take effect if the qubes-firewall service is running in the net qube. If the service is stopped or has crashed in the net qube **changes to the firewall rules will not be implemented**. If the service is stopped and a new qube is started, the firewall will fail closed: no traffic will pass. For qubes that are *already* attached, changes to the firewall configuration will not be processed until the firewall service is restarted.
- add to user/security-in-qubes/firewall.rst:94:  The Whonix-Gateway (e.g. ``sys-whonix``) does not respect the qubes-firewall service. Firewall rules configured via ``qvm-firewall`` or in the Qubes Manager will have no effect. See the `Whonix documentation <https://www.whonix.org/wiki/Qubes>`__ for more details on networking in this configuration.


- add link to user/how-to-guides/how-to-install-software.rst:260:1. ``qubes-updates-proxy`` (and its deprecated name: ``qubes-yum-proxy``) - a service providing a proxy for templates - by default enabled in NetVMs (especially: sys-net)
- here this can link to the service page: for clarity if needed since this is an user page and not advanced or developer
:doc:`qvm-service </user/advanced-topics/qubes-service>`)


- add user/downloading-installing-upgrading/upgrade/qubes-dist-upgrade-releasever-4.3-help.txt:21:                                         - enable minimal-netvm / minimal-usbvm services

- add user/templates/windows/qubes-windows-tools.rst:47: Using the Qubes graphics driver will not provide seamless mode unless the ``qvm-features`` parameter ``gui`` is set to ``1``. To disable the graphics driver, the parameter ``gui`` has to be set to an empty string, while the parameter ``gui-emulated`` has to be set to ``1``.
- add to user/templates/windows/qubes-windows-tools.rst:334:   **Windows 10 and 11:**  If the experimental graphics driver is installed and enabled by setting the ``qvm-features`` parameter ``gui`` to 1, seamless mode can be enabled and will stay enabled even after reboot of the qube.
- add user/templates/windows/qubes-windows-tools.rst:678:In order to avoid Qubes stopping Windows VMs forcefully after QWT was uninstalled, it is advisable to undefine the qrexec control for these VMs. This can be done using the command ``qvm-features <VMNAME> qrexec ""``.
- add user/templates/windows/qubes-windows-tools.rst:51:**Note**: Xen PV disk drivers are not installed by default. This is because they seem to cause problems (BSOD = Blue Screen Of Death). We're working with upstream devs to fix this. **However**, the BSOD seems to only occur after the first boot, and everything works fine after that. **Enable the drivers at your own risk**, of course, but we welcome reports of success/failure in any case (backup your VM first!). With disk PV drivers absent, ``qvm-block`` will not work for the VM, but you can still use standard Qubes inter-VM file copying mechanisms. On the other hand, the Xen PV drivers allow USB device access even without QWT installation if ``qvm-features stubdom-qrexec`` is set as ``1``.

- user/advanced-topics/managing-vm-kernels.rst:280:- ``default-kernelopts-common.txt`` - default kernel options, in addition to those specified with ``kernelopts`` qube property (can be disabled with ``no-default-kernelopts`` feature)

- add developer/services/disposablevm-implementation.rst:138:- The qube has the ``internal`` feature enabled. Qubes GUI applications hide :term:`internal qubes<internal qube>`;

- add user/reference/glossary.rst:146:      A qube which has the ``internal`` feature set. Those qubes are hidden from GUI tools such as user menu and have many GUI features disabled. They are used for the :term:`management qube` and preloading disposable qubes. In most cases, internal qubes should not be manipulated by the user directly.

to Add
- user/templates/windows/qubes-windows-tools.rst:300:   For audio, the parameter ``audio-model`` can be selected as ``ich6`` or ``ich9``; select the value that gives the best audio quality. Audio quality may also be improved by setting the following parameters, but this can depend on the Windows version and on your hardware:
- user/templates/windows/qubes-windows.rst:18:- audio output and input (available even without QWT installation if ``qvm-features audio-model`` is set as ``ich6``)
- user/templates/windows/qubes-windows-migrate.rst:6:For Windows 7, 10, and 11, there is a way to migrate backups created under Qubes R4.1 to R4.2 and R4.3. For this, a backup must be created under R4.1, and this backup has to be restored under Qubes R4.2 or R4.3, respectively, resulting in a VM well integrated into this Qubes version. If ``qvm-features <VMname> audio-model ich6`` is set, Windows will even have audio, although for Windows 10 and 11, it is somewhat scratchy.
- user/templates/windows/qubes-windows-tools.rst:40:- **Audio** - Audio support is available even without QWT installation if ``qvm-features audio-model`` is set as ``ich6``



- developer/services/disposablevm-implementation.rst:66:There are several events that may trigger the creation or deletion of preloaded disposables. Qubes OS attempts to maintain as many preloaded disposables as the user set through the ``preload-dispvm-max`` feature, and any preloaded disposables that are invalid will be deleted or replaced as soon as possible. This is balanced by the need to ensure the system is responsive, thus the refreshing of disposables may take some time.
- developer/services/disposablevm-implementation.rst:72:Maximum number of preloaded disposables is defined by the ``preload-dispvm-max`` feature. This feature can be set on any disposable template (this defines the maximum number of disposables preloaded from this template), and on the ``dom0`` (this defines the maximum number of disposables preloaded from the system-wide default disposable template). In case those values conflict, the ``dom0`` value takes precedence (for example, if ``preload-dispvm-max`` is set to 0 on the ``default-dvm`` disposable template, which is the default disposable template, and to 4 on ``dom0``, Qubes OS will attempt to preload 4 disposables from ``default-dvm``).
- developer/services/disposablevm-implementation.rst:78:  - Changing the ``preload-dispvm-max`` feature;
- developer/services/disposablevm-implementation.rst:79:  - Changing system-wide :py:attr:`~core-admin:qubes.vm.dispvm.DispVM.default_dispvm` if system-wide ``preload-dispvm-max`` is set to a different value than the disposable template setting;

- developer/services/disposablevm-implementation.rst:145:- The ``preload-dispvm-threshold`` feature controls how much free memory must be present on the system before attempting to create a new preloaded disposable. Used to ensure preloaded disposables do not consume all available memory, which would prevent starting other qubes.

- add to developer/debugging/vm-interface.rst:50:- ``/qubes-service/SERVICE_NAME`` - subtree for VM services controlled from dom0 (using the ``qvm-service`` command or Qubes Manager). One of ``1``, ``0``. Note that not every service will be listed here, if an entry is missing, it means “use VM default”. A list of currently supported services can be found in the ``qvm-service`` man page.
- add to user/advanced-topics/salt.rst:383:This enables, disables, or sets to default, services as in ``qvm-service``.
- use program: for user/security-in-qubes/ctap-proxy.rst:63:The above assumes a ``work`` qube in which you would like to enable ctap. Repeat the ``qvm-service`` command for all qubes that should have the client proxy enabled. Alternatively, you can add ``qubes-ctap-proxy`` in :menuselection:`VM settings -> Services` in the Qube Manager of each qube you would like to enable the service. Attempting to start the ``qubes-ctap-proxy`` service in the device-hosting qube (``sys-usb``) will fail.

- add to developer/system/networking.rst:120:Starting with Qubes 4.0, there is opt-in support for IPv6 forwarding. Similar to the IPv4, traffic is routed and NAT is applied at each network gateway. This way we avoid reconfiguring every connected qube whenever uplink connection is changed, and even telling the qube what that uplink is - which may be complex when VPN or other tunneling services are employed. The feature can be enabled on any network-providing qube, and will be propagated down the network tree, so every qube connected to it will also have IPv6 enabled. To enable the ``ipv6`` feature use ``qvm-features`` tool and set the value to ``1``. For example to enable it on ``sys-net``, execute in dom0:
- add to developer/system/networking.rst:120:Starting with Qubes 4.0, there is opt-in support for IPv6 forwarding. Similar to the IPv4, traffic is routed and NAT is applied at each network gateway. This way we avoid reconfiguring every connected qube whenever uplink connection is changed, and even telling the qube what that uplink is - which may be complex when VPN or other tunneling services are employed. The feature can be enabled on any network-providing qube, and will be propagated down the network tree, so every qube connected to it will also have IPv6 enabled. To enable the ``ipv6`` feature use ``qvm-features`` tool and set the value to ``1``. For example to enable it on ``sys-net``, execute in dom0:
- use manpages consistently developer/system/template-manager.rst:86:  - Fields (corresponding to :doc:`qvm-prefs <core-admin-client:manpages/qvm-prefs>` and :doc:`qvm-features <core-admin-client:manpages/qvm-features>` tags with the same names)
    - user/how-to-guides/how-to-copy-and-paste-text.rst:62:You can also change them using :doc:`qvm-features <core-admin-client:manpages/qvm-features>`.
- needs rewrite -  There is no feature more like that: user/how-to-guides/how-to-enter-fullscreen-mode.rst:54:As an alternative to the Xfce4 method, you can enable fullscreen mode for selected qubes by using a :doc:`gui-* <core-admin-client:manpages/qvm-features>` feature called ``gui-allow-fullscreen``.

- add to developer/system/template-manager.rst:150:  - Note that the contents of these files are stored in ``qvm-features`` upon installation. See the section below for details.
- add to developer/system/template-manager.rst:160:The template manager needs to keep metadata of installed templates such as versions and origin. This data can be stored via ``qvm-features`` to keep things consistent when, e.g., ``qvm-remove`` is used. Besides, backups are also more easily handled this way.
- developer/system/template-manager.rst:200:  - Note that the newlines in this field are converted to ``|`` to work better with existing tools like ``qvm-features``.
- add to user/how-to-guides/how-to-use-disposables.rst:192:This can also be changed from the command line from the :term:`GUI domain`, with :program:`qvm-features` targeting ``dom0``:
- add to user/how-to-guides/how-to-use-disposables.rst:213:This can also be changed from the command line from the :term:`GUI domain`, with :program:`qvm-features` targeting ``default-mgmt-dvm``:



```

maiska:
otherwise:
- use consistently thoughout documentation
There are two services (:doc:`core-admin-client:manpages/qvm-service`, :doc:`service framework </user/advanced-topics/qubes-service>`):
vs Use :doc:`qvm-service <manpages/qvm-service>`
not both!
same for qvm-features.

developer/releases/4_3/release-notes.rst
```diff
@@ -33,7 +33,7 @@ Major features and improvements since Qubes 4.2
 - Device “self-identity oriented” assignment (a.k.a New Devices API)
   (`#9325 <https://github.com/QubesOS/qubes-issues/issues/9325>`__).
 
-  **See:** :option:`core-admin-client:qvm-device --assignments`, ``qvm-device assign`` and ``qvm-device unassign``
+  **See:** :option:`qvm-device --assignments`, :doc:`qvm-device assign <manpages/qvm-device>` and :doc:`qvm-device unassign <manpages/qvm-device>`
```
maiska, line 36: i would prefer if there is a correct link to the specific command as in
**See:** :option:`qvm-device --assignments`, :doc:`qvm-device assign <manpages/qvm-device>` and :doc:`qvm-device unassign <manpages/qvm-device>`

https://doc.qubes-os.org/projects/core-admin-client/en/latest/manpages/qvm-device.html#assign
and

user/advanced-topics/disposable-customization.rst
```diff
@@ -104,7 +104,7 @@ For instance, to have Mozilla Thunderbird application open all attachments in a
 
       qvm-appmenus --get-available --i-understand-format-is-unstable <QUBE> | grep -i thunderbird
 
-In the current case, we identified the application ID to be ``net.thunderbird.Thunderbird`` (in your case it may as well be ``thunderbird``). Finally, enable the ``app-dispvm.net.thunderbird.Thunderbird`` service via the :ref:`qube settings <how-to-enable-a-service-in-the-settings>`.
+In the current case, we identified the application ID to be ``net.thunderbird.Thunderbird`` (in your case it may as well be ``thunderbird``). Finally, enable the :option:`app-dispvm.net.thunderbird.Thunderbird <qvm-features app-dispvm.*>` service via the :ref:`qube settings <how-to-enable-a-service-in-the-settings>`.
```
maiska, line 107: - clarify: Here the ref to qvm-features can be misleading as we are talking about a service. Either add a clarifying sentence, or let the ref be for the time being till this doc is audited.

user/how-to-guides/how-to-edit-a-policy.rst
```diff
@@ -82,4 +82,4 @@ See also
 --------
 
 * :doc:`/developer/services/qrexec`
-* `qubes-core-qrexec's documentation <https://dev.qubes-os.org/projects/qubes-core-qrexec/en/latest/>`__
+* :doc:`core-qrexec:index`
```
maiska, line 85: - I would prefer a custom title for qrexec, such as qrexec developer documentation

user/advanced-topics/bind-dirs.rst
```diff
@@ -151,41 +153,36 @@ Custom persist is an optional advanced feature allowing the creation of minimal
...
 
-Then, configure a persistent directory with ``qvm-features``:
+2. Then, configure a persistent directory with :doc:`qvm-features <manpages/qvm-features>` in the :term:`admin qube`:
 
-.. code:: console
-
-      $ qvm-features my-app-vm custom-persist.my_persistent_dir /var/my_persistent_dir
+   .. code:: console
 
+         [user@dom0] $ qvm-features QUBE_NAME custom-persist.FEATURE_NAME PATH
```
maiska, line 166: do not delete the example please

user/advanced-topics/bind-dirs.rst
```diff
@@ -151,41 +153,36 @@ Custom persist is an optional advanced feature allowing the creation of minimal
...
 
 
-Bind dirs are obviously still supported but this must be configured either in the template (``/usr/lib/qubes-bind-dirs.d`` and ``/etc/qubes-bind-dirs.d``) or from dom0 using ``qvm-features``. The bind dirs declaration must be done this way: ``qvm-features <VMNAME> custom-persist.<ARBITRARY NAME> [PRE-CREATION SETTINGS]<PATH>``
-
-To use this feature, first, enable it:
-
-.. code:: console
-
-      $ qvm-service -e my-app-vm custom-persist
+Bind dirs are obviously still supported but this must be configured either in the template (``/usr/lib/qubes-bind-dirs.d`` and ``/etc/qubes-bind-dirs.d``) or from dom0 using :doc:`manpages/qvm-features`. The bind dirs declaration must be done with this kind of feature: :option:`qvm-features custom-persist.{FEATURE_NAME} <qvm-features custom-persist.*>` (where :samp:`FEATURE_NAME` is an arbitrary name). The value should be formatted like this: :samp:`[PRE-CREATION SETTINGS:]<PATH>` (where :samp:`[PRE-CREATION-SETTINGS:]` is optional and :samp:`<PATH>` is an absolute path.
```
maiska, line 156: - do not remove  The bind dirs declaration must be done this way: qvm-features <VMNAME> custom-persist.<ARBITRARY NAME> [PRE-CREATION SETTINGS]<PATH>.
Add the explanations later. Wordings such as ' with this kind of' are fluff and should be omitted.
- add custom title to the manpage

user/advanced-topics/bind-dirs.rst
```diff
@@ -151,41 +153,36 @@ Custom persist is an optional advanced feature allowing the creation of minimal
 
 
 
-Bind dirs are obviously still supported but this must be configured either in the template (``/usr/lib/qubes-bind-dirs.d`` and ``/etc/qubes-bind-dirs.d``) or from dom0 using ``qvm-features``. The bind dirs declaration must be done this way: ``qvm-features <VMNAME> custom-persist.<ARBITRARY NAME> [PRE-CREATION SETTINGS]<PATH>``
-
-To use this feature, first, enable it:
-
-.. code:: console
```
maiska, line 158: - add also Enables minimal state qube with the custom persist feature. When enabled, user bind dirs must be defined using qvm-features custom-persist.* as /rw/config/qubes-bind-dirs.d/ will be ignored. This will also disable /home and /usr/local persistence as well as /rw/config/* automatically executed configs.

maiska (changes requested):

PR #1734, opened 2026-07-16
Make qubes-doc installable with pyproject.toml

### LINKED ISSUE #10737: Make some build dependencies optional in qubes-doc
[How to file a helpful issue](https://www.qubes-os.org/doc/issue-tracking/)

### The problem you're addressing (if any)

When building qubes-doc, the user has to install useless dependencies. 

### The solution you'd like

From: https://github.com/QubesOS/qubes-doc/pull/1642#issuecomment-3983544375

I think we should have 3 different dependencies, based on 3 different scenarios:

1. the user that wants to build a trusted version of the documentation. They only need:
   ```
   sphinx
   sphinxnotes-any
   sphinxnotes-strike
   sphinx-rtd-theme
   ```

2. someone (i.e. unman maintaining the onion site) or something (Read the Docs) that builds a version for the web:
   ```
   sphinx-reredirects
   sphinxext-opengraph
   ```

3. the contributor that might also want dev dependencies:
   ```
   sphinx-autobuild
   sphinx-lint
   ```

The dependencies of 2 and 3 should be optional and independent. The two extensions of option 2 should be optional also in `conf.py` and triggered with an environment variable like the `QUBES_DOC_FULL_BUILD` in https://github.com/QubesOS/qubes-doc/pull/1636

The 3rd option should be fixed by https://github.com/QubesOS/qubes-doc/pull/1642

This also should be documented.

### The value to a user and who that user might be

* Minor: the current full build process takes 21s instead of 22s on my machine.
* Limit the dependencies

### Completion criteria checklist

(This section is for developer use only. Please do not modify it.)

### LINKED ISSUE #10738: Remove duplicated dependencies in qubes-doc
[How to file a helpful issue](https://www.qubes-os.org/doc/issue-tracking/)

### The problem you're addressing (if any)

qubes-doc has two files listing the dependencies, it's hard to maintain.

Related: https://github.com/QubesOS/qubes-doc/pull/1642

### The solution you'd like

Anything that won't duplicate the requirements:

* drop `requirements.txt`
* make `pyproject.toml` use it
* use a `uv.lock` file

That might require:

* Cleaning up the `requirements.txt` dependencies
* Making qubes-doc a proper python package

### The value to a user and who that user might be

If someone wants to bump any dependency version, it will be easier.

### Completion criteria checklist

(This section is for developer use only. Please do not modify it.)

parulin:
* Move extensions as part of the `qubes_doc` package
* Update `conf.py`
* Update `pyproject.toml`:
  * Convert `qubes_doc` to a real package
  * Move some packages to optional dependencies groups
  * Bump sphinx version
* Remove `requirements.txt`
* Update `.readthedocs.yaml` to use the new install method
* Update installation instructions accordingly
* Add basic instructions to build the docs in README

After all this, the PDF would still not build on Read The Docs, so a workaround is to remove (almost) unused `sphinxnotes-*` extensions

Fix QubesOS/qubes-issues#10738
Fix QubesOS/qubes-issues#10737

Thanks to @hippalectryon-0 for some parts. See also: #1642


parulin:
Tested:

* [x] html build, with the pull request build
* [x] html build, locally with pip
* [x] html build, locally with poetry
* [x] html build, locally with uv
* [x] html build, locally with uv without optional dependencies
* [x] [epub and pdf on read the docs](https://app.readthedocs.org/projects/parulin-qubes-doc/builds/?version__slug=installable-pyproject-toml)

unman:
I think there is too much being done here, which could  be spread across separate PR.
Looking at 10737, I dont think I want to introduce another level of complexity to the build. I've though carefully about this and I do not want to do it - the benefit to users is minimal.

parulin:
Okay, what changes do you want to keep? The one allowing us to move to Sphinx >9.1?

Note that this PR doesn't change anything to most users. The only difference is that someone who want to create a perfect copy of the official doc website have to install an extra group of dependencies.

PR #1742, opened 2026-07-29
Clarify the name of the doc and current version

parulin:
* Indicate "core documentation" (it's not core-admin or else)
* Use the release name
* Put an admonition on development version if built on Read The Docs
    
Partial fix of: https://github.com/QubesOS/qubes-issues/issues/10263

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -21,11 +21,11 @@ copyright = f'%Y, {author} and others'
 
 # The major project version, used as the replacement for the |version| default
 # substitution. i.e. '4.3'
-version = 'latest'
+version = 'development'
 
 # The full version, including alpha/beta/rc tags
 # Note: this value is not used in the Qubes OS documentation
-release = version 
+release = version
 
 
 # -- General configuration ---------------------------------------------------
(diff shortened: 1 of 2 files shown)


conf.py
```diff
@@ -13,19 +13,19 @@
 
 # -- Project information -----------------------------------------------------
 
-project = 'Qubes OS'
+project = 'Qubes OS core documentation'
```
maiska: why core documentation? It sounds weird. It is the official Qubes OS documentation.
There are other documentations such as comments even for core components of Qubes OS. 
I find it confusing tbh :)

conf.py
```diff
@@ -95,7 +95,7 @@
 
 html_theme = 'sphinx_rtd_theme'
 
-html_title = f'{project} Documentation'
+html_title = f'Qubes OS core documentation (development version)'
```
maiska: Same here :) The clarification in the brackets is not needed i think

conf.py
parulin: That's the term we use all over the place, in the index, in the style guide... There is maybe only one mention of "official" documentation in the official documentation :)

And the idea is the following: if you link from `core-admin` to `qubes-doc`, the name of the link should be clear enough. `core-admin` might be part of the core documentation, though?

conf.py
parulin: This was meant to help fixing https://github.com/QubesOS/qubes-issues/issues/10263

**Note:** I'm not debating, just answering (yeah, I can!), I will make the changes later.

parulin:
I reverted to the original `project` name and to the default value for `html_title`.

PR #1743, opened 2026-07-29
Clarify the name of the doc and current version for R4.3

parulin:
Same as #1742 but for R4.3

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -21,7 +21,7 @@ copyright = f'%Y, {author} and others'
 
 # The major project version, used as the replacement for the |version| default
 # substitution. i.e. '4.3'
-version = '4.3'
+version = 'R4.3'
 
 # The full version, including alpha/beta/rc tags
 release = '4.3.0'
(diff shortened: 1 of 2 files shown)


conf.py
```diff
@@ -13,15 +13,15 @@
 
 # -- Project information -----------------------------------------------------
 
-project = 'Qubes OS'
+project = 'Qubes OS core documentation'
```
maiska: same here for the core documentation

conf.py
```diff
@@ -21,7 +21,7 @@
 
 # The major project version, used as the replacement for the |version| default
 # substitution. i.e. '4.3'
-version = '4.3'
+version = 'R4.3'
 
 # The full version, including alpha/beta/rc tags
 release = '4.3.0'
```
maiska: version should be 4.3.1

parulin: I thought that the docs are only relative to a major Qubes release? Also, Qubes versions/releases are the opposite of how Sphinx handle versions/releases.

maiska: 4.3.0 is very specific. Then 4.3.x  but it is still somewhat weird. Let's make ti the current version 4.3.1 and it should be updated accordingly.

parulin: So ... something like the following?

```
version = '4.3.1'
release = version
```

PR #1745, opened 2026-08-07
Add a little FAQ to Split GPG-2 (and 1)

parulin:
There are some questions that are [frequently asked](https://forum.qubes-os.org/t/anyone-setup-sequoia-pgp/42574/5) in the context of Split GPG-2. That seems irrelevant for the general FAQ, so I put it at the end of the dedicated pages:

* there are a lot of outdated posts and guides suggesting to install `split-gpg2-dom0` which is not required.
* there is a confusion about Split GPG-2 being the new version of Split GPG-1. It's not.

### DIFF
--- a/user/security-in-qubes/split-gpg-2.rst
+++ b/user/security-in-qubes/split-gpg-2.rst
@@ -1,3 +1,5 @@
+.. _split-gpg-2:
+
 ===========
 Split GPG-2
 ===========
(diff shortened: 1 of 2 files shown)


maiska:
why not add these FAQs to the official FAQ so that they do not get lost, and there is only one spot for FAQ?

parulin:
Well, to me, it would be an infrequently asked question everywhere except on this page. I'm a bit lost about the general FAQ, it is quite in bad shape so I didn't want to worsen the situation (from my POV). But if you think it would be better here, I can move things in the general FAQ.

maiska:
Or you just can remove the title "*frequently asked questions" from both of these files

parulin:
For clarity, I think some kind of title for this section would be better than nothing, don't you think? Why not "Questions about Split-GPGX"?

maiska:
How about  just using an info box. Here the information is important.
Info box, that is highlighted falls more in place that having a super specific FAQ with one question and one sentence as an answer :)

PR #1749, opened 2026-08-26
Correct outdated devices widget descriptions

armarmgc:
Replace references to an icon with a yellow square, which is not used anymore, with the current USB trident icon.

Convert mentions of 'Device Widget' to 'Devices Widget' to use the correct name.


user/how-to-guides/how-to-use-usb-devices.rst
```diff
@@ -33,7 +33,7 @@ With Qubes Device Widget
 
    Qubes Devices Widget tray icon
 
-Click the Device Widget: a list of available devices appears. USB-devices have a USB-icon to their right: |usb icon|
+Click the Devices Widget: a list of available devices appears.
```
maiska: Hey, would you like to mark Device Widget with :guilabel:` Device Widget` in the text where applicable? :)

armarmgc: I am not sure if a guilabel is appropriate, since the Devices Widget only appears as a tray icon, not as a button or a menu item. The Devices Widget text only appears as a tool-tip when the icon is hovered over.

maiska: yh, i had also the dilemma, but it is the next best thing, apart from simple highlighting and a custom icon role ... for the several occurences I think we are set with guilabel for now, since it is part of the GUI, program is the next next best thing imho.

armarmgc: I noticed that the 'Getting Started' page introduces it as 'Qubes Devices widget', with 'Qubes Devices' in bold, but the documentation sometimes includes the 'Qubes' in front, and sometimes does not. This happens for the other widgets as well.

It would be good to use a consistent name for it and determine what markup to use, but I think that should be done in a separate pull request, where we make references to all the widgets, not just the Devices Widget, to be consistent.

maiska: Hm, would you like to fix overall to :guilabel:`Qubes Devices widget` ?

armarmgc: I think that would work, but I think we should do this fix for all the widgets at once in a separate pull request. If we only change the devices widget to use a guilabel here, it would cause an inconsistency between the markup used for the devices widget and the other widgets.

maiska: Let's take this one step at a time :) this PR will be for this widget. If you want to get another PR ready in the background for the rest or should i ? :)

armarmgc: I added a commit to this PR for consistently using the name `Qubes Devices widget` and a guilabel.

I created another pull request (https://github.com/QubesOS/qubes-doc/pull/1752) for using consistent naming and a guilabel for all the widgets.

maiska: cool, thank you! :)

PR #1744, opened 2026-08-01
Exclude GNOME from usability & ux documentation

RandyTheOtter:
As far as I can tell Qubes OS team in not considering GNOME as a desktop environment anymore, see https://github.com/QubesOS/qubes-issues/issues/1806

p.s. I asked around and it seems like we prefer Qt over GTK for UI too

PR #1750, opened 2026-08-26
Update for outdated migration to GNOME

armarmgc:
The 'Usability & UX' page describes an outdated status on the migration to GNOME. The page has been updated to reflect the current status of staying with Xfce.


parulin:
Already done here: https://github.com/QubesOS/qubes-doc/pull/1744

